PR #6, opened 2017-09-27
readme clarification and small correction

willingc:


yuvipanda:
Thanks for tracking this, @willingc!

PR #12, opened 2017-09-28
Incident report for hub 403 incident on Sep 27 2017

yuvipanda:


incident-reports/2017-09-27-hub-403.md
```diff
@@ -0,0 +1,123 @@
...
+
+The PR has been merged and deployed, bringing the outage to an end. 
+
+# Conclusion
+
+This was caused by two factors:
+
+1. Local development environment being slightly different from production
+   environment (single node vs multi node)
+2. Testing in staging not thorough enough
```
choldgraf: I would add one more thing, which is: "changing lots of things at the same time." It sounds like a reason this was hard to debug is that we have a few different moving pieces that we're juggling (e.g., jupyterhub will be upgraded soon and that'll "just fix" some problems) that caused us to miss this. Not much we can do about that short-term as a lot of this code is evolving rapidly, but good to keep in mind to try keeping our # of changed variables at a minimum on each iteration.

incident-reports/2017-09-27-hub-403.md
```diff
@@ -0,0 +1,123 @@
...
+   and make sure they are part of marking a deployment green or red. 
+   [Issue](https://github.com/jupyterhub/mybinder.org-deploy/issues/10)
+2. Write down the responsibilities of the people doing the deployment.
+   Specifically include the lesson of 'never ignore any errors in staging, they
+   will always bite you in production' prominently. 
+   [Issue](https://github.com/jupyterhub/mybinder.org-deploy/issues/11)
+   
+## Local development
+
+1. Consider figuring out an easy way for people to run a development environment
```
choldgraf: This seems like an important point to me...any other possibilities than using another cloud instance?

incident-reports/2017-09-27-hub-403.md
yuvipanda: Added note!

incident-reports/2017-09-27-hub-403.md
yuvipanda: I amended this PR with a link to an issue I opened about this with a possible solution: https://github.com/jupyterhub/binderhub/issues/137

yuvipanda:
Merging this since I'm about to perform some other git surgery. Feel free to continue to leave comments here!

minrk:
Hooray for incident reports!

minrk:
An external test battery to run against an existing deployment would also be useful. We could make sure that runs against staging before deploying to prod.

PR #20, opened 2017-09-29
Upgrade support charts + fix prometheus to actually work

yuvipanda:


yuvipanda:
I've manually deployed these for now, these don't autodeploy yet.

PR #51, opened 2017-10-05
r2d version bump

choldgraf:


choldgraf:
wtf?

yuvipanda:
you had the ordering backwards. your base should be 'beta' and 'head' should be 'staging'

choldgraf:
derp

PR #59, opened 2017-10-05
add github client id, secret to common secret config

minrk:
and bump binderhub to support them (https://github.com/jupyterhub/binderhub/pull/165)


yuvipanda:
Shouldnt we use a different key for staging and prod?

minrk:
Maybe when it's doing something other than applying a rate limit. I think one token for the whole deployment makes sense for now.

yuvipanda:
Cool! We can re examine if we actually start hitting the limits!

minrk:
Sounds good. I'll deploy this to beta once staging is up and I've verified that it's using them correctly.

minrk:
It's working. Yay!

PR #48, opened 2017-10-04
update instructions for deploy

choldgraf:
update instructions so they're a bit more clear and verbose. LMKWYT @yuvipanda

Issue #10, opened 2017-09-28
Write end-to-end tests for mybinder.org


yuvipanda:
These should exercise common pathways that real users do, and should be fast enough to be used as part of each and every deployment automatically.

Things it should test:

1. A repo requiring new image building is ok
2. A repo that launches an image that is already built is also ok
3. Users can connect to the notebook after launching

yuvipanda:
There's one now that tests (2) and (3) under tests/ in this repo. I'm just structuring them as pytest tests, which is great!

minrk:
For 1. to get maximum realistic test, is it possible to delete the image for the test repo prior to requesting the build? Does the test env have sufficient permission to do that?

yuvipanda:
@minrk yeah, technically it can do that - since we already have the service account key for gcr.io login in config/secret/common.yaml. I guess we could just use that and do a delete before building. But I'm unsure how we can make that idempotent tho - if two intsances of this test are running at the same time, they'll trample each other.

Another possibility is we have a github repository that we make a new commit to each time we do this. We push it into a new branch, do the build, verify it works, then remove the branch. This will be much more idempotent test that trying to whitebox our way into deleting the image. What do you think?

minrk:
@yuvipanda a dedicated test repo that we push trivial commits to (e.g. `date > file`) sounds like the best option.

yuvipanda:
@minrk I've created an org (https://github.com/binderhub-ci-repos) for this.

PR #81, opened 2017-10-16
Refactor readme links and minor edits to text

willingc:
- Factor out common or subject to change links so they can easily be updated at the end of the file
- Clarify the steps where information is needed from Travis and include a link to the Travis build
- Make minor markdown style edits

This PR is a follow up to #48. The target audience is someone who does deploys infrequently to mybinder.org.


README.md
```diff
@@ -1,100 +1,118 @@
...
+   to [staging.mybinder.org][].
+4. Go to [staging.mybinder.org][] to look at the changes.
+5. Verify that [staging.mybinder.org][] works as intended. Please take your
+   time to check that the change is working as expected.
+
+**If the changes look correct:**
+
+6. Make a new PR, merging [staging][] into the [beta][] branch.
+7. Get this PR merged, and wait for Travis to make a deployment to [beta][].
+8. CELEBRATE! :tada:
```
yuvipanda: I think step 8 should be 'verify it works correctly on beta', similar to point 5.

willingc: Good suggestion. Will update that.

README.md
```diff
@@ -1,100 +1,120 @@
 # [mybinder.org-deploy][]
 
 This repository contains configuration files and documentation related to the
-[binderhub](https://github.com/jupyterhub/binderhub) deployment open to the
-public at [beta.mybinder.org](https://beta.mybinder.org).
+[binderhub][] deployment open to the public at [beta.mybinder.org][].
```
choldgraf: how is it that I have written so much `.md` and have never discovered this feature of it?

README.md
```diff
@@ -1,100 +1,120 @@
...
+the `staging` branch of the repository. Second, if all looks well, you'll
+deploy to the production branch of the repository (called `beta`).
+
+1. Make the changes on your fork.
+2. Make a PR to the `staging` branch with the changes you want.
+3. Review, accept, and merge this PR. This will make Travis deploy the changes
+   to [staging.mybinder.org][].
+4. Go to [staging.mybinder.org][] to look at the changes.
+5. Verify that [staging.mybinder.org][] works as intended. Please take your
+   time to check that the change is working as expected.
```
choldgraf: What do folks think about (in a diff PR) having a "test" repo that we tell people to build to make sure it works? I feel like being explicit about "you should really make sure `staging` works" is important

yuvipanda: I think that ties into #10. I'm not sure if we can provide explicit instructions when we can automate it instead...

my strategy is to make a dummy commit to http://github.com/yuvipanda/example-requirements and then make sure it builds..

README.md
```diff
@@ -1,100 +1,120 @@
...
+
+6. **Immediately revert the PR that was made to the [staging][] branch.**
+7. Verify that [staging.mybinder.org][] is working as it was before the PR
+   and revert.
+8. Troubleshoot and make changes to your fork. Repeat the process from Step 1.
+
+## Upgrading dependencies for the mybinder.org deployment
+
+Upgrading dependencies used by [mybinder.org][] requires making changes
+to the `config` files of repositories that are used to build the
```
choldgraf: I think this would be "config file of the repository that is used to build the `mybinder.org` service" (singular) since it's always `common.yaml` and it's always in this repo. Does that sound right?

choldgraf:
this is great! some small comments but other than that LGTM

README.md
choldgraf: ya I agree automating would be better

README.md
willingc: 🤷‍♀️

README.md
willingc: Good point. Let's leave as is here and flesh out when addressing #10.

README.md
willingc: Good catch. Will reword.

yuvipanda:
@willingc can you resolve conflicts? I think we can merge after.

minrk:
Conflicts resolved

Issue #43, opened 2017-10-03
Switch billing over to our accounts


choldgraf:
This is an issue to keep track of progress on the billing front. The background is that we need to switch over the billing for `mybinder.org` to our accounts instead of Jeremy's. I've been going back and forth with BIDS admins about figuring this out and we're making progress. I've got an email out to Google Cloud sales so hopefully they can answer some questions for us. Right now we need to figure out:

1. Can we pay for google cloud with a check instead of a credit card? (makes things much easier for billing purposes)
2. What's the expected monthly cost for Binder?
3. Are there any opportunities to get an education discount for our use?

Will update this issue as new information arises.

choldgraf:
This has hit a snag, as Berkeley wasn't immediately keen on letting us get binder reimbursed through our grant...I'm trying to follow up on this with other folks at berkeley who have paid for cloud computing w/ grants. hopefully something will come out of that

PR #82, opened 2017-10-16
Allow POST/PUTs of upto 64mb

### LINKED ISSUE #66: max file size is 1MB
Create a notebook with the cell:

```python
from IPython.display import publish_display_data
kB = 1024
publish_display_data({
    'application/x-fake-size-test': 'x' * 1024 * kB 
})
```

run that cell, and save will fail with HTTP 413 (too big!).

We have [this annotation](https://github.com/jupyterhub/mybinder.org-deploy/blob/614d6f59c1677c5a385552947226e3afa46a8f2e/config/common.yaml#L25), which I assume is meant to address exactly this, but it doesn't appear to have the desired effect.

yuvipanda:
Supports notebooks / other uploads of this size.

Fixes #66

### DIFF
--- a/support/values.yaml
+++ b/support/values.yaml
@@ -6,6 +6,9 @@ nginx-ingress:
       annotations:
         prometheus.io/scrape: "true"
   controller:
+    config:
+      # Allow POSTs of upto 64MB, for large notebook support.
+      proxy-body-size: 64m
     stats:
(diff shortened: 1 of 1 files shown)


yuvipanda:
This is still manually deployed (until #43 gets fixed, at least).

PR #85, opened 2017-10-16
Merge #23, #48, #82, #84 into beta

yuvipanda:
Should be no-ops


yuvipanda:
We were drifting off, so even though these should be no-ops let's do a merge regardless.

PR #86, opened 2017-10-17
staging: non-json output fixes

minrk:
bump repo2docker, binderhub for fixes to non-json outputs


minrk:
This deploy failed (the deploy didn't, but the tests did) due to the current load on Binder.

PR #100, opened 2017-10-26
Rename beta to prod, and make it happen!

yuvipanda:
This switches the URLs from beta.mybinder.org to mybinder.org!

This way, we can do a seamless switchover, start redirecting folks
from beta.mybinder.org to mybinder.org, and when the hub there is
fully drained, we can remove it too!

This also keeps the current beta.mybinder.org infrastructure alive,
in case we need to switch to it.


yuvipanda:
Note that this is actually a noop, the real action happens when we merge this into prod.

yuvipanda:
Note that the mybinder.org site is still not up fully yet (waiting on a DNS change to happen), and there's already a branch here named 'prod' that is deploying to it.

yuvipanda:
Note that I also need to change a lot of the README!

.travis.yml
```diff
@@ -33,6 +33,7 @@ deploy:
     branch:
       - staging
       - beta
```
minrk: having deleted config/beta.yaml, beta branch shouldn't be here anymore

Issue #11, opened 2017-09-28
Write up deployment guidelines & responsibilities


yuvipanda:
Write up a small document that lays out the responsibilities of the various people involved in a deployment. 

See https://phabricator.wikimedia.org/L3 for something a little similar (although not the same)

PR #103, opened 2017-10-31
Update README to match new reality

yuvipanda:
Apologies for the delay in updating docs to match!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ drift away from `staging` too much.
 
 ## Pre-requisites
 
+The following are tools and technologies that mybinder.org uses. You should have
+a working familiarity with them in order to make changes to the mybinder.org deployment.
+
 ### Google Cloud Platform
 
 MyBinder.org currently runs on Google Cloud. There are two Google Cloud projects
--- a/README.md
+++ b/README.md
@@ -58,6 +58,12 @@ to deploy these changes by following the steps above in [Deploying a change][].
 This section explains how to upgrade the [mybinder.org][] deployment after
 making a change in the [BinderHub][] repo.
 
+BinderHub is deployed via a helm chart that is tied to a particular commit on the
+BinderHub repository. BinderHub is a "requirement" for this mybinder.org deployment,
+which is why it is specified in requirements.yaml. Upgrading the version of BinderHub
+that is used in mybinder.org corresponds to updating the BinderHub helm chart version,
+which we step through below.
+
 1. Merge changes to [BinderHub][].
 2. Open the [Travis build for BinderHub](https://travis-ci.org/jupyterhub/binderhub),
    navigate to the page corresponding to the master branch.
@@ -95,6 +101,11 @@ making a change in the [BinderHub][] repo.
 
 This section explains how to upgrade the [mybinder.org][] deployment after
 making a change in the [repo2docker][] repo.
+
+BinderHub uses a docker image with repo2docker in it. When a new commit is merged in
+the repo2docker repository, a new version of this image is pushed. We then configure
+BinderHub to use the newly built image (which is identified by a tag) by editing `values.yaml`.
+The following lines describe how to point mybinder.org to the new repo2docker image
 
 1. Merge changes to [repo2docker][].
 2. Open the [Travis build for repo2docker](https://travis-ci.org/jupyter/repo2docker),
--- a/README.md
+++ b/README.md
@@ -182,8 +182,8 @@ that we use:
 1. `binder-staging` contains all resources for the staging deployment
 2. `binder-prod` contains all resources for the production deployment
 
-We'll hand out credentials to anyone who wants to the staging deployment, so
-please just ask!
+We'll hand out credentials to anyone who wants to play with the staging deployment,
+so please just ask!
 
 [mybinder.org-deploy]: https://github.com/jupyterhub/mybinder.org-deploy
 [mybinder.org]: https://beta.mybinder.org
--- a/README.md
+++ b/README.md
@@ -16,6 +16,46 @@ branch to [mybinder.org][]. In general (except when performing a
 deployment), these branches should always be the same, and `prod` should not
 drift away from `staging` too much.
 
+## Pre-requisites
+
+### Google Cloud Platform
+
+MyBinder.org currently runs on Google Cloud. There are two Google Cloud projects
+that we use:
+
+1. `binder-staging` contains all resources for the staging deployment
+2. `binder-prod` contains all resources for the production deployment
+
+We'll hand out credentials to anyone who wants to play with the staging deployment,
+so please just ask!
+
+While you only need merge access in this repository to deploy changes, ideally
+you should also have access to the two Google Cloud Projects so you can debug
+things when deployments fail.
+
+### Kubernetes Basics
+
+We heavily use Kubernetes for the mybinder.org deployment, and it is important you
+have a working knowledge of how to use Kubernetes. Detailed explanations are out
+of the scope of this repository, but there is a good [list of tutorials](https://kubernetes.io/docs/tutorials/).
+Specifically, going through the [interactive tutorial](https://kubernetes.io/docs/tutorials/kubernetes-basics/)
+to get comfortable using `kubectl` is required.
+
+### Helm Basics
+
+We use [helm](https://helm.sh) to manage our deployments, and it is imporant you
+have a working knowledge of how to use helm. Detailed explanations are out of the
+scope of this repository, but [docs.helm.sh](https://docs.helm.sh) is an excellent
+source of information. At a minimum, you must at least understand:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,56 +105,57 @@ making a change in the [repo2docker][] repo.
    Copy the text in `<YOUR-IMAGE-NAME>`. **Note**: You may need to unfold the
    code in the `Deploying application` line in order to see this text.
 3. In your fork of the [mybinder.org-deploy][] repository, open
-   `config/common.yaml`.
-4. Toward the end of the file you will see `repo2dockerImage`, replace the
+   `mybinder/values.yaml`.
```
choldgraf: above it's `requirements.yaml` but here it's `values.yaml`...are we intentionally splitting this up into two files now?

README.md
```diff
@@ -169,6 +174,17 @@ Related repositories used by the [mybinder.org][] service are:
...
 
+## Google Cloud Projects
+
+MyBinder.org currently runs on Google Cloud. There are two Google Cloud projects
+that we use:
+
+1. `binder-staging` contains all resources for the staging deployment
+2. `binder-prod` contains all resources for the production deployment
+
+We'll hand out credentials to anyone who wants to the staging deployment, so
```
choldgraf: to *access* the staging...

README.md
choldgraf: looks like everywhere else you call it `values.yaml`, I think `requirements.yaml` is wrong?

yuvipanda: Yes, those two files have two very different functions (one specifies dependencies and the other does default values. This is the default helm structure (https://github.com/kubernetes/helm/blob/master/docs/charts.md#the-chart-file-structure)

README.md
yuvipanda: good catch!

README.md
choldgraf: ah ok - hmm, lemme look over the language again then. Why do we treat r2d and binderhub differently in terms of how they are specified here? Maybe it'll be easier to discuss in person.

yuvipanda: Our earlier practice of putting the version in values.yaml was a hack that had its problems. It required that we write custom code to pass it to helm, and also meant we could not auto-deploy changes to non-binderhub charts. Using requirements.yaml properly lets us get rid of our hack & auto-deploy changes to *all* the charts we wanna use.

yuvipanda: Because binderhub is a chart we are installing - helm knows how to get a chart, download it, install it, etc. The repo2docker version is not actually a version at all, but just the name of the image we are asking binderhub to use - that is a config option we pass to binderhub.

If you look at https://github.com/kubernetes/helm/blob/master/docs/charts.md#the-chart-file-structure it says:

```
  requirements.yaml   # OPTIONAL: A YAML file listing dependencies for the chart
  values.yaml         # The default configuration values for this chart
```

README.md
```diff
@@ -32,9 +32,9 @@ deploy to the production branch of the repository (called `beta`).
 
 **If the changes look correct:**
 
-6. Make a new PR, merging [staging][] into the [beta][] branch.
-7. Get this PR merged, and wait for Travis to make a deployment to [beta][].
-8. Verify that [beta.mybinder.org][] works as intended. Please take your
+6. Make a new PR, merging [staging][] into the [prod][] branch.
+7. Get this PR merged, and wait for Travis to make a deployment to [prod][].
+8. Verify that [mybinder.org][] works as intended. Please take your
    time to check that the change is working as expected.
```
choldgraf: How about we specifically say "try building a binder image (make sure the build process happens, this may require a commit to be made to the repository you try to build)"

README.md
```diff
@@ -72,20 +72,24 @@ making a change in the [BinderHub][] repo.
    <img src="docs/static/travis-screenshot.png" width="500" />
```
choldgraf: above, under the short "BinderHub" description section, let's add something like

"BinderHub is deployed via a helm chart that is tied to the latest commit on the BinderHub repository. Upgrading the version of BinderHub that is used in `mybinder.org` corresponds to updating the BinderHub helm chart version, which we step through below."

README.md
```diff
@@ -101,56 +105,57 @@ making a change in the [repo2docker][] repo.
    Copy the text in `<YOUR-IMAGE-NAME>`. **Note**: You may need to unfold the
```
choldgraf: Under the short description under the title, we could put:

"In contrast to BinderHub (which is deployed with a helm chart), repo2docker is utilized as a Docker Image that contains a particular version of repo2docker in it. When a new commit is added to repo2docker, a new version of this image is pushed. The following lines describe how to point `mybinder.org` to the new repo2docker image."

README.md
choldgraf: a few suggestions to clear this up above...WDYT?

README.md
```diff
@@ -16,6 +16,46 @@ branch to [mybinder.org][]. In general (except when performing a
 deployment), these branches should always be the same, and `prod` should not
 drift away from `staging` too much.
 
+## Pre-requisites
+
```
choldgraf: "The following are tools and technologies that `mybinder.org` uses. You should have a working familiarity with them in order to make changes to the `mybinder.org` deployment."

yuvipanda: Done

README.md
```diff
@@ -16,6 +16,46 @@ branch to [mybinder.org][]. In general (except when performing a
...
+you should also have access to the two Google Cloud Projects so you can debug
+things when deployments fail.
+
+### Kubernetes Basics
+
+We heavily use Kubernetes for the mybinder.org deployment, and it is important you
+have a working knowledge of how to use Kubernetes. Detailed explanations are out
+of the scope of this repository, but there is a good [list of tutorials](https://kubernetes.io/docs/tutorials/).
+Specifically, going through the [interactive tutorial](https://kubernetes.io/docs/tutorials/kubernetes-basics/)
+to get comfortable using `kubectl` is required.
```
choldgraf: "you can also find guidelines on using kubernetes in the <link> zero to jupyterhub tutorial </link>"

yuvipanda: Nope, I don't think that's an appropriate place for us to link people to. Any advice we have there
will be highly specific, and I think that'll be super confusing.

README.md
choldgraf: Also, "for a general list of terminology with some basic definitions, see the `<link>zero to jupyterhub glossary </link>`

choldgraf:
nice additions - a couple comments in pre-reqs

README.md
yuvipanda: Don't think we should do that either! I think 'crossing the wires' as such will only confuse things, since anything in z2jh is going to be highly specific.

README.md
choldgraf: really? I was imagining just linking to the debugging page:

https://zero-to-jupyterhub.readthedocs.io/en/latest/debug.html

which has a lot of useful info that would be relevant to this deployment as well, no?

yuvipanda:
@choldgraf thanks! I pushed a new commit based on one comment & responded to the other two.

README.md
yuvipanda: The problem is how do you as the user figure out what is relevant and what is not? For example the example about proxy.secretToken is quite irrelevant. 

Also, I think this PR is already big enough, and should focus on making sure the info in the README is accurate (which it currently isn't). We're already getting tangential with the pre-requisites IMO and that should all wait for a different PR (one you could start!).

README.md
choldgraf: Ah sorry I didn't mean the glossary page, I meant the "tools" page:

https://zero-to-jupyterhub-with-kubernetes.readthedocs.io/en/latest/tools.html

seems that all of those are equally relevant for debugging/maintaining/etc the Binder deployment. But if you disagree I won't block on this

README.md
choldgraf: that's fair. I guess in my mind for someone to be deploying / debugging mybinder.org, they've already got some familiarity with the team and won't be learning in the dark. Either way, good point re: too large of a PR, I'm fine w/ the code as it currently is. Think we should wait for another input?

README.md
yuvipanda: Ah, I had totally forgotten that page exists. IMO, that belongs in a different page that actually explains how this deployment is structured, different places to learn about it, and different ways to debug common problems. Some of this discussion should also be part of #11 rather than here.

README.md
yuvipanda: If you're happy, I think it's good to merge.

choldgraf:
it sounds like while there may be some improvements to be made to the README, but @yuvipanda makes a good point that those should wait for another PR. Let's get this one merged so that it is technically correct, and we can iteratively improve later if we want.

choldgraf:
Gonna merge this one since right now the readme is incorrect and this fixes that...if other folks have thoughts, please bring 'em up here or open an issue etc!

PR #123, opened 2017-11-02
Add an nginx based HTTP redirector

yuvipanda:
Generic enough that we can use this later for other redirects too.

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -13,3 +13,10 @@ binderhub:
       host: hub.mybinder.org
 
   googleAnalyticsCode: "UA-101904940-1"
+
+redirector:
+  redirects:
+    - type: host
+      host:
+        from: beta.mybinder.org
+        to: staging.mybinder.org
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -19,4 +19,4 @@ redirector:
     - type: host
       host:
         from: beta.mybinder.org
-        to: staging.mybinder.org
+        to: mybinder.org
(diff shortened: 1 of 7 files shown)


config/prod.yaml
```diff
@@ -13,3 +13,10 @@ binderhub:
       host: hub.mybinder.org
 
   googleAnalyticsCode: "UA-101904940-1"
+
+redirector:
+  redirects:
+    - type: host
+      host:
+        from: beta.mybinder.org
```
choldgraf: shouldn't this just point to `mybinder.org`?

yuvipanda: good catch

yuvipanda: Fixed

mybinder/templates/redirector/ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: redirector
+  labels:
+    app: redirector
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    kubernetes.io/tls-acme: "true"
+spec:
+  rules:
+    {{ range $redirect := .Values.redirector.redirects }}
```
willingc: If I run `helm lint --debug --home './config/' mybinder/` on the PR, I get the following error for this line:
```
==> Linting mybinder/
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/: render error in "mybinder/templates/redirector/ingress.yaml": template: mybinder/templates/redirector/ingress.yaml:12:33: executing "mybinder/templates/redirector/ingress.yaml" at <.Values.redirector.r...>: can't evaluate field redirects in type interface {}
```

yuvipanda: Did you pass in -f config/prod.yaml too? redirects are defined only in the prod / staging specific files...

I'll add a shim to values.yaml so this continues to work, however!

PR #127, opened 2017-11-07
get git-crypt binary

minrk:
and checksum bin before installing

avoids recompilation of git-crypt on every install


yuvipanda:
Nice! Thanks, @minrk!

@Carreau you might be interested in this.

Issue #146, opened 2017-11-14
firewall/limit outbound traffic


minrk:
We should be limiting outbound (and possibly inbound to a lesser extent) traffic in the user pods to limit what mischief they can get up to.

yuvipanda:
I agree! We can do some limiting on the VPC level (disallow all outbound connections on port 25 for example) and some at the Kubernets NetworkPolicy level (pods aren't allowed to talk to each other).

choldgraf:
+1, better to get out ahead of these kinds of potential problems rather
than playing catchup when there's a fire to put out
--

minrk:
The network policy I would like to start with:

1. *whitelist* ports 80, 443 (all other ports blocked until someone has a legitimate need for other ports)
2. throttle outbound bandwidth (not sure what is a reasonable value). Ideally, I would like to throttle this *excluding* the notebook port, so we aren't throttling the user experience, only sending data elsewhere. I'm not sure if that's feasible.

Possible stretch goals:

1. apply a network quota per container, so that if/when it exceeds the limit it gets killed
2. track usage (for a short period, only for abuse detection) of client ip addresses, so that we can ban ips if need be

I'd like the tracking of pod creations by ip in general, so we can see if an ip address/range is responsible for traffic spikes.

I'll try to do some reading to see what our options are here. [istio](https://istio.io/docs/tasks/policy-enforcement/rate-limiting.html) appears to be a candidate.

yuvipanda:
More info at https://github.com/jupyterhub/mybinder.org-deploy/issues/36.

Also https://kubernetes.io/docs/concepts/services-networking/network-policies/ should be useful!

choldgraf:
I think another thing we can do is track non-user-specific network data etc (via prometheus?). It'll be pretty easy to run analytics on that to see if there are outlier users in general, and to decide on limits etc that hit most usecases just fine.

choldgraf:
I think another thing we can do is track non-user-specific network data etc (via prometheus?). It'll be pretty easy to run analytics on that to see if there are outlier users in general, and to decide on limits etc that hit most usecases just fine.

minrk:
[...]

PR #151, opened 2017-11-17
add firewall-rules script

minrk:
script to setup firewall rules for a kubernetes cluster, via gcloud firewall-rules API

- allow internal communication
- block all outgoing traffic
- whitelist http,https

I've applied this to staging and it appears to do what I intend (I can fetch things, but cannot ssh, etc.).

The firewall applies to our whole cluster, which means this applies to binderhub and builders as well as user containers. This is positive, because it means builds are similarly restricted, and not an issue because all of our existing requests to the outside world are via HTTP[S] already.

gets the baseline of #146

### DIFF
--- a/scripts/firewall-rules
+++ b/scripts/firewall-rules
@@ -1,0 +1,141 @@
+#!/usr/bin/env python3
+"""
+Create firewall rules for a Kubernetes cluster
+
+Assumes that you have first activated the correct project for gcloud API, e.g.
+
+    gcloud config set project binder-staging
+
+Currently creates three egress rules:
+
+- allow internal kube communication
+- whitelist http/s to the world
+- block all other egress traffic communication
+"""
+
+import json
+import os
+import pipes
+from subprocess import check_output
+import sys
+
+# CIDR for the whole world
+WORLD_CIDR = '0.0.0.0/0'
+
+
+def gcloud(*args):
+    """Run a gcloud command and return its parsed JSON output"""
+    cmd = ['gcloud', '--format=json']
+    cmd.extend(args)
+    print(' '.join(map(pipes.quote, cmd)))
+    return json.loads(check_output(cmd).decode('utf8', 'replace'))
+
+# Usage: gcloud compute firewall-rules create NAME (--action=ACTION | --allow=PROTOCOL[:PORT[-PORT]],[PROTOCOL[:PORT[-PORT]],...]) [optional flags]
+#   optional flags may be  --action | --allow | --description |
+#                          --destination-ranges | --direction | --help |
+#                          --network | --priority | --rules | --source-ranges |
+#                          --source-tags | --target-tags
+
+
+def add_rule(name, rule, existing_rules, replace=False):
+    """Add a single firewall rule"""
+    if name in existing_rules:
+        # TODO: check if the rule is correct
+        print(f"{name} already exists.", end=' ')
+        if replace:
+            print("deleting...")
+            gcloud('compute', 'firewall-rules', 'delete', '-q', name)
+        else:
+            print("skipping...")
+            return
+    cmd = ['compute', 'firewall-rules', 'create', name]
+    for key, value in rule.items():
+        cmd.append(f'--{key}={value}')
+    gcloud(*cmd)
+
+
+def main(cluster, replace=False):
+    cluster_info = gcloud('container', 'clusters', 'describe', cluster)
+    prefix = f"binder-{cluster}-fw-"
+    cluster_cidr = cluster_info['clusterIpv4Cidr']
+    rules = {
+        rule['name']: rule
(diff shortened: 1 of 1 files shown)


scripts/firewall-rules
```diff
@@ -0,0 +1,141 @@
...
+import sys
+
+# CIDR for the whole world
+WORLD_CIDR = '0.0.0.0/0'
+
+
+def gcloud(*args):
+    """Run a gcloud command and return its parsed JSON output"""
+    cmd = ['gcloud', '--format=json']
+    cmd.extend(args)
```
willingc: @minrk do you wish to have this be generic? It seems that its a potential security issue if args contains a malicious command.

willingc: OK. I see how you are using this below and why you have it generic.

scripts/firewall-rules
```diff
@@ -0,0 +1,141 @@
...
+    # get nodes for our cluster
+    cluster_nodes = [
+        node for node in nodes
+        if any(
+            node['name'].startswith(prefix)
+            for prefix in instance_group_prefixes)
+    ]
+    assert cluster_nodes, f"No cluster nodes for {instance_group_prefixes}"
+
+    # construct a set of the node netowork tags
```
willingc: Typo in network

betatim:
Did you change something on staging despite this PR still being open? The activity log on console.cloud.google lists something about firewall rules.

Mostly just trying to understand why the build for #152 is timing out :)

yuvipanda:
This is great, but I think we're sortof re-inventing https://cloud.google.com/deployment-manager/docs/ here. Long term, we should move closer to the https://github.com/berkeley-dsep-infra/data8xhub model, where we have CI that's managing our *entire* infrastructure (clusters, firewall rules, etc) with something like https://github.com/berkeley-dsep-infra/data8xhub/blob/master/templates/gdm.yaml for the gcloud aspects. 

That would allow us to have fully reproducible infrastructure without having to write our own code! 

I am happy to not block this work on using GDM though! Am not sure what the deployment model for this as is would be though. 

Thank you for all the great work, @minrk!

minrk:
I’ve disabled the rules on staging as they blocked something helm was doing during the deploy, which I don’t have time to investigate right now. I don’t understand what helm was trying to do that was being blocked, so if anyone can figure out what port forwarding it’s doing that requires egress that would be great. My first hunch is the service net.

 I’m AOK using an existing tool, I just didn’t know about gdm. I actually created these rules by hand in the console when I was testing, and wrote this script mainly as a record/reproducibility step. I’m not committed to the approach.

I think we do need to automate and/or script the creation of new clusters, which is one of the missing pieces.

betatim:
>  so if anyone can figure out what port forwarding it’s doing that requires egress that would be great

I tried to find the old travis log but because we kept hitting "restart" the latest one doesn't contain the error anymore. My best idea for getting to see the error is to turn on the FW rules again and try to deploy to staging :-/

minrk:
I saved the failed log [here](https://gist.github.com/fc190c4c3708f50ea506588e8ed94e42) before retriggering the build after disabling the rules, which succeeded. It's some port forwarding that helm is doing. I'll try again later, probably Monday. My guess is the whitelist for internal communication was insufficient.

minrk:
I've updated the 'allow-kube' rule here to allow anything on 10.x.y.z because I couldn't figure out what subnets were relevant. Since this is only a temporary solution while we wait for egress network policy support, I think this is okay.

I've reapplied this to the staging cluster for testing and successfully did a deploy to staging, but if anyone runs into an issue like last time, feel free to delete all of the rules on the [egress tab here](https://console.cloud.google.com/networking/firewalls/list?project=binder-staging&tab=EGRESS) (They all start with `binder-staging-fw-...`).

minrk:
The latest deploy to staging succeeded and tests passed, so I think this is okay. I'd like to apply these rules to prod as well, but I won't do that without an okay from the team, and will probably wait until Monday since it's a holiday weekend in the US.

minrk:
Anyone object to merging and applying this one to prod now? If it turns out to cause a problem, egresses can be deleted [from the console](https://console.cloud.google.com/networking/firewalls/list?project=binder-prod&tab=EGRESS).

choldgraf:
IMO if we wanna deploy something that might cause tiny fires we should do it now, or plan to wait a couple of weeks for the eLife blog post bump to go down. Either is fine w/ me!

minrk:
I'd like to do it before we get too much visibility, so my vote is now.

yuvipanda:
I'll be flying rest of the day but feel free to push this through, @minrk

minrk:
Thanks! I'll give it a go. And if anyone sees this as breaking anything, feel free to just turn the rules off on the console and we'll come back to the drawing board.

minrk:
applied to prod-a

PR #152, opened 2017-11-17
Deploy bhub 275, 285, 197 and 283, update deploy docs

betatim:
Only thing that should change the deployment is jupyterhub/binderhub#275 which adds urlPath handling to the legacy handler.

The travis output seems to have changed, so I updated the docs as well. I think I did it right: https://travis-ci.org/jupyterhub/binderhub/builds/303366405#L671


yuvipanda:
@betatim can you deploy this to prod too?

choldgraf:
I believe this resulted in an SSH error : https://travis-ci.org/jupyterhub/mybinder.org-deploy/builds/303621152 we haven't been able to figure out what's going wrong w/ that

PR #156, opened 2017-11-17
Add a test that does a full build + launch

### LINKED ISSUE #10: Write end-to-end tests for mybinder.org

yuvipanda:
Fixes #10


minrk:
awesome!

PR #159, opened 2017-11-24
minor fixes in test_build

minrk:
- ensure branch cleanup happens (if `yield` raises due to test failure, cleanup was being skipped)
- echo build messages to stdout, so they will be in 'captured stdout' report on test failure
- use f-strings because we're on Python 3.6! 🎉 

I made two small commits directly to staging to get the build working, since it was preventing successful deploy:

- 056181f403f959b12b3d023bf0b28518da6ec289 ensures the deploy key is used in the initial clone, because *a* key is required to clone with ssh. Local testing doesn't run into this because most of us will have a default ssh key that works.
- 657214fa390219d44577e1e3b438029776c91230 fixes permissions on the deploy private key, which git-crypt was unpacking as world-readable and ssh refused to use private keys with open permissions

### DIFF
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -27,24 +27,22 @@ def push_dummy_gh_branch(repo, branch, keyfile):
         with open(branchfile, 'w') as f:
             f.write(branch)
         subprocess.check_call(['git', 'add', branchfile], cwd=gitdir)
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks @minrk!

PR #161, opened 2017-11-24
use tc-init to limit egress bandwidth in user containers

minrk:
The current limit is 1mbit (128kB/s). Should it be something different? Lower? Higher?

As it is now, there aren't very many valid use cases for sending data out from a binder, but once there is the option to load GitHub credentials and push, then 128k might start feeling pretty slow. I'd rather wait and then turn it up than find out we have a problem and turn it down, though.

This is the upload bandwidth for connections initiated from the container.

Specifically **not** affected:

1. downloads from the container initiated from outside (e.g. browser downloads/UI)
2. downloads to the container initiated from the container (e.g. download of data files into the container)

Only running binders are affected, not builds which are still a place where arbitrary user code executes. But builds don't yet expose initContainer config at the moment, and our concurrent build limit and the possibility of a build timeout (do we have one?) also makes this a less valuable target.

This is the main part of #146

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -37,6 +37,20 @@ binderhub:
       memory:
         guarantee: 1G
         limit: 4G
+      initContainers:
+      - name: tc-init
+        image: minrk/tc-init:0.0.3
+        env:
+          - name: EGRESS_BANDWIDTH
+            value: 1mbit
+        securityContext:
+          # capabilities.add seems to be disabled
+          # by the `runAsUser: 1000` in the pod-level securityContext
+          # unless we explicitly run as root
+          runAsUser: 0
+          capabilities:
+            add:
+            - NET_ADMIN
 
   repo2dockerImage: jupyter/repo2docker:093184b
 


choldgraf:
I'm also on the side of "be conservative in how much we allow unless enough people complain and we find a _really_ compelling case for why to raise the limit, rather than the other way around". Allowing for more egress traffic is IMO a good reason to deploy your own binderhub rather than using `mybinder.org` :-)

minrk:
I agree. Do you think we should turn it down even more, e.g. to 64 or 32 kBps? Or is 128k low enough?

betatim:
I reckon 128k is low enough and better than nothing (current state). So let's merge this and if we feel we need to lower/increase the limits it is an easy thing to do.

betatim:
TFW you only had ISDN for internet is back

betatim:
`curl -s https://raw.githubusercontent.com/sivel/speedtest-cli/master/speedtest.py | python -` thinks that you can still upload at about 100Mbit/s :-/

minrk:
init container config doesn't appear to be applying. I'll have to look at that later.

betatim:
https://github.com/kubernetes/kubernetes/issues/47264#issuecomment-339137300 might be relevant

yuvipanda:
egress also counts data being sent from the website to the user, so I think this would make it seem like the user is browsing mybinder.org over a 128k connection, which should be pretty bad. 

Let's set it at 100Mbit, and work on either a daemonset or sidecar containers that progressively limit it after some amount of usage?

minrk:
I think init-container support was added to the hub chart after the latest version used in the binder chart, which is why this isn't taking effect. That means we need to get the binderhub chart working with the latest hub chart.

@yuvipanda if you're right about this also affecting UI traffic, I agree that 1Mbit is too slow. But I'd like to be sure and will start testing. I'm far from 100% confident in my tests, so I'll double check (networking is magic!). It's possible that I measured download speed when the throttling wasn't working.

minrk:
Follow-up: @yuvipanda is 100% right! #166 causes this to have the behavior I expected it to have, though.

PR #162, opened 2017-11-27
grep-grep-awk magic for culling old pods

betatim:
Some more wisdom for removing stale pods.

via @minrk He also points out that we could make a notebook that uses the k8s API to do this in a nicer fashion.

### DIFF
--- a/WISDOM.md
+++ b/WISDOM.md
@@ -2,3 +2,14 @@
 
 * When you are in an outage, focus only on fixing the outage - do not try to do anything else.
 * Prefer minor annoyances happening infrequently but at regular intervals, rather than major annoyances happening rarely but at unpredictable intervals.
+
+
+# Cull all pods older than four hours
+
+```
+kubectl get pod | grep '^jupyter-' | grep '\(\([12][0-9]\|[4-9]\)h\|d\)$'
+```
+or to get all those that hav eexisted for more than 24hours
+```
+kubectl delete pod $(kubectl get pod | grep '^jupyter-' | grep 'd$' | awk '{print $1}')
+```
--- a/WISDOM.md
+++ b/WISDOM.md
@@ -2,14 +2,3 @@
 
 * When you are in an outage, focus only on fixing the outage - do not try to do anything else.
 * Prefer minor annoyances happening infrequently but at regular intervals, rather than major annoyances happening rarely but at unpredictable intervals.
-
-
-# Cull all pods older than four hours
-
-```
-kubectl get pod | grep '^jupyter-' | grep '\(\([12][0-9]\|[4-9]\)h\|d\)$'
-```
-or to get all those that hav eexisted for more than 24hours
-```
-kubectl delete pod $(kubectl get pod | grep '^jupyter-' | grep 'd$' | awk '{print $1}')
-```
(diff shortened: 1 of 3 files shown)


WISDOM.md
```diff
@@ -1,3 +1,14 @@
 # A place to put the collective wisdom of the Binder project.
 
 * When you are in an outage, focus only on fixing the outage - do not try to do anything else.
+
+# Cull all pods older than four hours
+
+```
+kubectl get pod | grep '^jupyter-' | grep '\(\([12][0-9]\|[4-9]\)h\|d\)$'
```
choldgraf: add a `--namespace=prod` in the `get pod` section!

WISDOM.md
```diff
@@ -1,3 +1,14 @@
...
+```
+or to get all those that hav eexisted for more than 24hours
```
choldgraf: `existed` (typeo)

WISDOM.md
```diff
@@ -1,3 +1,14 @@
...
+```
+kubectl delete pod $(kubectl get pod | grep '^jupyter-' | grep 'd$' | awk '{print $1}')
```
choldgraf: namespace here too

WISDOM.md
choldgraf: also there's no culling here, only a listing of pods!

choldgraf: I think we want

```
kubectl delete pod --namespace=prod $(kubectl get pod --namespace=prod | grep '^jupyter-' | grep '\(\([12][0-9]\|[4-9]\)h\|d\)$' | awk '{print $1}')
```

WISDOM.md
```diff
@@ -1,3 +1,14 @@
 # A place to put the collective wisdom of the Binder project.
 
 * When you are in an outage, focus only on fixing the outage - do not try to do anything else.
+
+# Cull all pods older than four hours
```
yuvipanda: This is an extremely destructive action, since it kicks out current users off mybinder.org that you should do only as either a:

1. Last resort
2. When  pods are in pending and the autoscaler is not working
3. When you know you are being hit by a specific hub or spawner bug

choldgraf: Maybe we should modify it to instead be pods stuck in containercreating that are older than X hours. Would we ever want to delete pods running normally?

Also we should definitely add in the yuvi wisdom about when to resort to this and what other avenues to try first... I'll track done the conversation in gitter when I'm at a computer

yuvipanda: Yeah, if stuff is stuck in ContainerCreating your core problem is elsewhere and needs debugging. Mostly fixing the core problem takes care of it, and just deleting the pods only fixes the symptom and it usually comes back instantly - and so there's not much cases where it's useful.

This is useful only for Pending pods.

choldgraf:
What do y'all think about the state of this PR? I feel like it now has a more natural home in the "site reliability guide"...@betatim would you be willing to port it over there? @yuvipanda also seemed to have reservations about the "culling old user pods" section - yuvi do you think we shouldn't add this information at all?

betatim:
This PR changed direction: I added a new section to the SRE guide that is the beginning of a collection of copy&paste ready commands you might use to do things or as part of a longer command.

choldgraf:
@yuvipanda I'm gonna merge this since the changes you requested are no longer applicable to the current state of the PR (which provides commands to *list* old pods, but doesn't say anything about deleting them)

choldgraf:
thanks a bunch @betatim :-)  we should iterate on the commands cheatsheet as we spot-check downtime etc!

PR #165, opened 2017-11-27
Revert "use tc-init to limit egress bandwidth in user containers"

yuvipanda:
Reverts jupyterhub/mybinder.org-deploy#161

IIRC, this will make it seem to users that mybinder.org is being accessed over a 
128k connection, which is too slow. I'd recommend a 100Mbit limit.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -37,20 +37,6 @@ binderhub:
       memory:
         guarantee: 1G
         limit: 4G
-      initContainers:
-      - name: tc-init
-        image: minrk/tc-init:0.0.3
-        env:
-          - name: EGRESS_BANDWIDTH
-            value: 1mbit
-        securityContext:
-          # capabilities.add seems to be disabled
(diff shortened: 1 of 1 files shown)


minrk:
> IIRC, this will make it seem to users that mybinder.org is being accessed over a
128k connection, which is too slow. I'd recommend a 100Mbit limit.

From the description of #161, I don't think that's true. The *only* traffic that should be affected is connections initiated from the container. User downloads from the UI should specifically not be affected.

I'd like to close this PR until it's verified that the user download is actually affected. I think 100Mbit is *way* to much to allow users to send data out.

yuvipanda:
@minrk closed. I agree that if it doesn't affect browser connections 100Mbit might be too much. I'm a bit confused about how it affects only those cases, but it's possible my knowledge of how `tc` works is either outdated or wrong.

minrk:
@yuvipanda I could be wrong, too! I'll do more testing, but that was the conclusion I came to after initial tests.

PR #166, opened 2017-11-27
whitelist local network in tc-init

minrk:
follow-up to #161

@yuvipanda pointed out that egress bandwidth was throttling UI, which would be too harsh.

This PR whitelists 10.x.x.x traffic, effectively preventing the egress bandwidth from applying to traffic coming from or going to browsers through the proxy, or local communication with the Hub, etc.

This is still having no effect on binder until the latest updates to the jupyterhub chart propagate through the binderhub chart.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -39,8 +39,10 @@ binderhub:
         limit: 4G
       initContainers:
       - name: tc-init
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks, @minrk!

PR #169, opened 2017-11-28
bump charts, update ingresses

minrk:
latest jupyterhub chart changes how letsencrypt ingresses are handled. JupyterHub's ingresses have changed:

1. jupyterhub lego support has moved to an internal egress with its own controller
2. the 'regular' ingress has removed automatic kube-lego support, so lego config needs to be applied manually here

- [x] needs to wait for binderhub chart to get https://github.com/jupyterhub/binderhub/pull/305 for the ingress updates

Landing this PR gets two important changes from the jupyterhub chart:

1. updates to KubeSpawner, which should fix pods orphaned by spawn timeouts (this is most of the really long-running pods, I think)
2. support for initContainers, so the tc-init bits enabling bandwidth throttling will start taking effect

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -25,14 +25,15 @@ binderhub:
     proxy:
       service:
         type: ClusterIP
+      https:
+        enabled: true
+        letsencrypt:
+          contactEmail: yuvipanda@gmail.com
     ingress:
       enabled: true
       annotations:
         ingress.kubernetes.io/proxy-body-size: 64m
         kubernetes.io/ingress.class: nginx
-      https:
-        enabled: true
-        type: "kube-lego"
     singleuser:
       memory:
         guarantee: 1G
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -25,15 +25,12 @@ binderhub:
     proxy:
       service:
         type: ClusterIP
-      https:
-        enabled: true
-        letsencrypt:
-          contactEmail: yuvipanda@gmail.com
     ingress:
       enabled: true
       annotations:
         ingress.kubernetes.io/proxy-body-size: 64m
         kubernetes.io/ingress.class: nginx
+        kubernetes.io/tls-acme: "true"
     singleuser:
       memory:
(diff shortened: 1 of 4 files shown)


minrk:
btw, once this gets the binderhub update, it will include the jupyterhub chart version with init-container support and thus the tc-init bandwidth throttling will start taking effect.

minrk:
This should work now, once jupyterhub/binderhub#305 is merged and requirements.txt here is updated to get the resulting binderhub chart. Others should feel free to make that change here if that PR is merged during the night, Europe time.

mybinder/values.yaml
```diff
@@ -30,9 +30,7 @@ binderhub:
       annotations:
         ingress.kubernetes.io/proxy-body-size: 64m
         kubernetes.io/ingress.class: nginx
-      https:
```
yuvipanda: Does this set up the tls secret for the binderhub ingress properly? I think the kube-lego stuff sets it up in the ingress template, so we should probably get rid of all that magic and have the same structure here we have for jupyterhub's external ingress?

minrk: Yeah, this works for both. The tls config for binderhub is at the top of this file, not affected here. The binderhub ingress has explicit support for kube-lego, so you just have to say `kind: kube-lego` and it's off. JupyterHub's ingress *used to* have this support and behave the same, but has removed it in favor of lego support in the internal ingress and requiring explicit configuration of the external ingress.

If anything, I might prefer to change the jupyterhub one back to behaving like the binderhub one does now, with explicit kube-lego support. One reason is that the jupyterhub one now requires setting a lot more explicit configuration, including setting the host twice, and requiring the tls secret name be set explicitly rather than being able to use chart templates.

But I don't think that needs to happen right away.

minrk:
I think this ought to be good to go, now. I think it would be nice to land this in production before tomorrow because it should reduce the orphaned pod problem that we keep having.

yuvipanda:
Go for it, @minrk

minrk:
🚀

PR #170, opened 2017-11-29
Add pod health check notebook

minrk:
Notebook that inspects the state of jupyterhub and running pods and creates a simple report, including identification of probable orphaned pods (i.e. user pods with no corresponding user registered with JupyterHub).

Example report from running earlier today:

screen shot 2017-11-29 at 14 59 24

Includes functions for deleting the orphaned pods if so desired.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+*~
+__pycache__
+*.py[co]
+.cache
+
+mybinder/charts
+mybinder/requirements.lock
+
+.ipynb_checkpoints
--- a/scripts/pod-health-check.ipynb
+++ b/scripts/pod-health-check.ipynb
@@ -1,0 +1,477 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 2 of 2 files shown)


choldgraf:
this is so cool! could we include a `binder/` folder w/ requirements needed
to run the notebook? That way we could run this from binder!

Also, I had put together a binder-stats repo to do analytics on binder.
Perhaps I should get rid of that and put it here? Or we should put this
notebook into `binder-stats` and make it part of the JupyterHub org?

https://github.com/choldgraf/binder-stats

yuvipanda:
This is awesome!

I don't believe you can run this from inside binder though, since it needs authentication to both kubernetes API

minrk:
I'm fine either way, I don't think I have a strong preference for whether it goes here or a dedicated repo, but as @yuvipanda pointed out, this must be run in a context with access to the deployment (i.e. on your own machine).

choldgraf:
ya I agree it'd be better in this repo too, didn't think about the credentials thing. Thanks @minrk !

PR #171, opened 2017-11-29
Add Site Reliabillity Guide

willingc:
http://test-mybinder-deploy.readthedocs.io/en/sre-guide/


minrk:
Very nice!

choldgraf:
wow this is great, thanks a lot @willingc :-D

two quick things:

1. It seems that this is duplicating some content we had in various markdown files before (e.g. I think in `contributing.md`. Should we remove the old content so that it's compiled in one place here?
2. I'll take a closer look through all the text when you remove the WIP!

willingc:
I've updated the guide to a cleaner style similar to BinderHub. Links have been checked and updated.

I think what's left to do is decide which content should remain in the README and what could be moved into the Site Reliability Guide. My recommendation would be to do that in a separate PR.

betatim:
After your link fixes did you also update the rendered version at: http://test-mybinder-deploy.readthedocs.io/en/sre-guide/ ? I just browsed it and the incident_template link there doesn't work.

willingc:
Good catch @betatim. You are better than `make linkcheck` which passed cleanly. Ugh recommonmark and markdown. I made some changes. This should be good to go.

minrk:
Thanks @willingc!

PR #173, opened 2017-11-29
Bump repo2docker version

yuvipanda:
- Change image prefix so all current images are invalidated. This
  gets in the crucial fix https://github.com/jupyter/repo2docker/pull/150
- Change image prefix to have repo2docker commit hash on it.

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -4,7 +4,7 @@ binderhub:
       - mybinder.org
 
   registry:
-    prefix: gcr.io/binder-prod/prod-v4-1-
+    prefix: gcr.io/binder-prod/r2d-3cd4519
 
   hub:
     url: https://hub.mybinder.org
--- a/config/staging.yaml
+++ b/config/staging.yaml
@@ -4,7 +4,7 @@ binderhub:
       - staging.mybinder.org
 
   registry:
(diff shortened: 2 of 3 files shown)


config/prod.yaml
```diff
@@ -4,7 +4,7 @@ binderhub:
       - mybinder.org
 
   registry:
-    prefix: gcr.io/binder-prod/prod-v4-1-
+    prefix: gcr.io/binder-prod/r2d-3cd4519
```
minrk: Is this a precedent where we should update the prefix every time we bump repo2docker in order to invalidate all builds?

yuvipanda: I don't think we should. We should only do so when we have a breaking change we want everyone to get, such as this (or the artful deploy). However, there's nothing wrong with bumping it up every time, and definitely leads to more consistent results. Am happy with either...

yuvipanda: haha, I love how I started with "I don't think we should" and ended with "Am happy with either" lol. I meant "I don't think we *do*", in that we don't have precedent.

minrk: OK. I think if it's less frequent, maybe tying it to version numbers in repo2docker is the way to go in the long run, but this works for now, at least.

PR #174, opened 2017-11-29
Block access to GCE Metadata service

yuvipanda:


minrk:
👍 assuming you verify from testing that this does what you mean it to!

yuvipanda:
Validated that this works!

PR #180, opened 2017-11-30
band-aid: sleep to give the deployment time to wake up

minrk:
we *should* be using a readiness probe so that `deploy.py` doesn't return before the deployment is fully ready.

This sleep should let us reduce the number of times we need to re-run travis in the meantime.

Remove this when we get a proper readiness probe!


betatim:
Feeling pragmatic so going to deploy this. Let's see if it improves things.

PR #183, opened 2017-11-30
deploy repo2docker fixes

minrk:
versions bumped:

- jupyterlab 0.29.2
- ipywidgets 6.0.1

venv uses frozen env spec for reproduciblity

image cache is invalidated due to jupyterlab version pinning problem in previous version

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -4,7 +4,7 @@ binderhub:
       - mybinder.org
 
   registry:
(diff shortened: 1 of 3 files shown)


minrk:
relevant upstream PRs: https://github.com/jupyter/repo2docker/pull/156 and https://github.com/jupyter/repo2docker/pull/158

PR #188, opened 2017-11-30
firewall: allow git protocol port

minrk:
unblock the port used by `git clone git://...`


yuvipanda:
@minrk +1, have you run this already?

minrk:
Just ran it

PR #192, opened 2017-11-30
Bump up nginx-ingress replicas to 5

yuvipanda:
1 wasn't good enough, was running into errors!


willingc:
**Incident:** Increased 400 errors and mybinder.org not accessible by some.

**Resolution:** Increasing the nginx-ingress replicas from 1 to 5 to better support 8 nodes.

PR #193, opened 2017-11-30
Merge #190, #192, #188 to prod

yuvipanda:


yuvipanda:
These are all noops, the replica one because I hand-fixed it when it went bad.

PR #199, opened 2017-12-01
Incident report for 2017-11-30 proxy OOM

betatim:
This is the writeup of the incident on 2017-11-30.

Some questions:
* I attached names to what I and Chris were doing/saying. Are we Ok with that? Past reports do not do this. Is this because it was only one person responding? I think in this case it is important to see the two different people because I think it shows were we can improve.
* @yuvipanda frequently uses `!log something something` on gitter, how does that work? Would be good as a way to reconstruct history later
* How do we discuss how things went and what to do different next time? This is important for the "Action items", in particular the "process" section.

[hackmd](https://hackmd.io/CYVgDAZgRhAcBMBaKsLEQFgJzCVkAzAGyKwDGFApvCBAIzADswQA?edit)

@choldgraf let me know what you think. I'd like to appear.in to chat about how things went, and a virtual high five. We noticed something was wrong and managed to (somehow) get things back running again. I don't have a lot of kubernetes experience so it makes me feel more comfortable that we do not have to resort to the "redeploy whole thing" hammer to get back online when there isn't anyone to ask for advice. So from my POV this is worth a high five!

### DIFF
--- a/docs/source/incident-reports/2017-11-30-oom-proxy.md
+++ b/docs/source/incident-reports/2017-11-30-oom-proxy.md
@@ -1,15 +1,15 @@
 # Incident report 2017-11-30 4:23PM PST
 
-# Summary
+## Summary
 Tim and Chris noticed at 0:59 CET (need to convert to PST) that mybinder.org showed an error that creating temporary users failed when launching a new server. This was noticed because Tim happened to look at grafana. One node running a proxy had run out of memory which broke the proxy. Probably a build container had used up all the memory on the node as builds had no memory limits enforced.
 
 
-# Timeline
+## Timeline
 
-## 2017-11-30 16:22 PST
+### 2017-11-30 16:22 PST
 Node runs out of memory, service starts degrading.
 
-## 16:59 PST
+### 16:59 PST
 Investigation starts after noticing grafana looks weird. Error message when launching a new server:
 ```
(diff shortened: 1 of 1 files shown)


betatim:
One more thing: I enjoyed playing with staging to confirm/reject the hypothesis that `kubectl replace -f...` was what lead to user pods being killed. Maybe we should encourage more experimenting with kubectl, restarting, breaking the setup on purpose and watching it fix itself on staging. You don't really want to type any command for the first time when it is the real thing. At least I am nervous that I might make things even worse.

docs/source/incident-reports/2017-11-30-oom-proxy.md
```diff
@@ -0,0 +1,96 @@
+# Incident report 2017-11-30 4:23PM PST
+
+# Summary
```
willingc: `##`

willingc: Add an additional `#` throughout or Sphinx will be grumpy with too many `#` top levels. ex. `## -> ###`

betatim: Done

willingc:
I'm going to go ahead and merge this since additions/changes can be added in a new PR.

PR #203, opened 2017-12-02
Add playground templates

betatim:
Set up a copy of nteract playground on playground.mybinder.org


mybinder/templates/pen/deployment.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: pen
+spec:
+  selector:
+    matchLabels:
+      app: pen
+      component: nginx
```
yuvipanda: change this too?

mybinder/templates/pen/deployment.yaml
```diff
@@ -0,0 +1,22 @@
...
+  replicas: 1
```
yuvipanda: This should be templatized so we can change this in config.yaml

mybinder/templates/pen/deployment.yaml
```diff
@@ -0,0 +1,22 @@
...
+  template:
+    metadata:
+      labels:
+        app: pen
+        component: nginx
+    spec:
+      containers:
+      - name: pen
+        image: python:3.6-slim
```
yuvipanda: This should also be parameterized.

mybinder/templates/pen/ingress.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: pen
+  labels:
+    app: pen
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    kubernetes.io/tls-acme: "true"
+spec:
+  rules:
+    - host: pen.mybinder.org
```
yuvipanda: Should be parameterized

mybinder/templates/pen/service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: pen
+  labels:
+    app: pen
+    component: nginx
+spec:
+  ports:
+  - port: 80
```
yuvipanda: This should match the port we're running in the deployment pods

betatim: this is not the port that is exposed to the outside world?

yuvipanda: bah, you're right. I think what this needs is a 'targetPort' since I assume we won't be running the playground on port 80

mybinder/templates/pen/deployment.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: pen
+spec:
+  selector:
+    matchLabels:
+      app: pen
+      component: playground
```
betatim: good labels?

yuvipanda: https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/labels.md has more info on standard labels.

mybinder/templates/pen/ingress.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: pen
+  labels:
+    app: playground
+    component: frontend
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    kubernetes.io/tls-acme: "true"
```
betatim: delete these lines too?

config/staging.yaml
```diff
@@ -53,3 +53,9 @@ redirector:
       host:
         from: docs.staging.mybinder.org
         to: mybinder.readthedocs.io
+
+playground:
+  image: yuvipanda/play.nteract.io:v0.1
```
yuvipanda: Usually images are separated out as something like:

```
image:
   repository: yuvipanda/play.interact.io
   tag: v0.1

File: config/staging.yaml [hunk 9]
```diff
@@ -53,3 +53,9 @@ redirector:
...
+  host: playground.staging.mybinder.org
```
yuvipanda: This should probably be:
```
ingress:
   hosts:
     - <host>
```

config/staging.yaml
yuvipanda: Also all the config here except for hostnames should be in values.yaml, since we try to limit the config differences between staging and prod.

config/staging.yaml
```diff
@@ -53,3 +53,9 @@ redirector:
...
+  replicas: 1
+  port: 3000
```
yuvipanda: This shouldn't actually be configurable I think. If it is, it should be under 'image'

mybinder/templates/playground/ingress.yaml
```diff
@@ -0,0 +1,23 @@
...
+  rules:
+    - host: {{ .Values.playground.host }}
+      http:
+        paths:
+          - path: /
+            backend:
+              serviceName: playground
+              servicePort: 8000
+  tls:
+    - secretName: kubelego-tls-pen
```
yuvipanda: playground rather than pen?

mybinder/templates/pen/ingress.yaml
yuvipanda: You need these for automatic HTTPS to work.

config/staging.yaml
betatim: how do you then combine them together in `deployment.yaml`? or do we there also have `image` and `tag`? Tried to copy repo2docker image specification but that has the tag in the image name :-/

yuvipanda:
The ingress annotations also need to be restored for nginx ingress to pick it up and automatic TLS to work.

config/staging.yaml
yuvipanda: Mostly like https://github.com/jupyterhub/binderhub/blob/master/helm-chart/binderhub/templates/deployment.yaml#L47. repo2DockerImage is an exception because it is passed straight through to binderhub.

yuvipanda: I usually look at the designs in https://github.com/kubernetes/charts/tree/master/stable for conventions to follow.

mybinder/templates/playground/ingress.yaml
```diff
@@ -0,0 +1,25 @@
...
+  labels:
+    app: playground
+    component: frontend
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    kubernetes.io/tls-acme: "true"
+spec:
+  rules:
+    {{ range $host := .Values.playground.ingress.hosts }}
+    - host: {{ $host.host }}
```
yuvipanda: This should be just $host I think?

mybinder/templates/playground/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: pen
+  labels:
+    app: playground
+    component: frontend
+spec:
+  ports:
+  - port: 80
+    protocol: TCP
+    targetPort: 8000
```
yuvipanda: This should be 3000 right? That's the port that the container image is listenig on.

mybinder/templates/playground/ingress.yaml
```diff
@@ -0,0 +1,25 @@
...
+      http:
+        paths:
+          - path: /
+            backend:
+              serviceName: playground
+              servicePort: 8000
```
yuvipanda: This should be 80, and match the port in the service, not targetPort

yuvipanda:
I cleaned this up a little, thanks for setting most of this up, @betatim!

betatim:
Thanks for the help!

mybinder/values.yaml
```diff
@@ -85,6 +85,12 @@ binderhub:
 
   repo2dockerImage: jupyter/repo2docker:fd74043
 
+playground:
+  image:
+    name: yuvipanda/play.nteract.io
```
choldgraf: should this become a jupyter image or an nteract image?

choldgraf: also would the idea be that we'd create a different image for various kernels people might wanna run? or is the plan to stick w/ python?

yuvipanda: It'll become an nteract image at some point.

It's just a UI, it's creating notebooks on mybinder using an API. We'd just need some uI elemenets there to allow people to select what repo to 'play' with.

PR #221, opened 2017-12-10
bhub 335 (dynamic favicon on fail)

Carreau:
https://github.com/jupyterhub/binderhub/pull/335


Carreau:
failed... reverting.

choldgraf:
@Carreau how did it fail? I found that in the latest push it would *always* fail the first time, then if I hit "rebuild" it would succeed

Carreau:
"Exit status 1" I did not pay particular attention to the why, I did not see any obvious stacktraces or error message beyond exit status of 1.

PR #228, opened 2017-12-13
Revert "Revert "staging: enable docker-in-docker""

minrk:
Retry docker-in-docker deploy

The dind directories on the nodes didn't look right. Deleting and recreating the pods fixed it, but it makes me nervous that this was necessary, especially since the two nodes behaved differently.

One node had no `/var/run/dind/docker.sock`, while the other had `/var/run/dind/docker.sock`, but it was a *directory*, which is weird and wrong (it should have been a socket).

After deploying this, assuming it succeeds, I am going to increase the cluster size to see if a new node gets a properly configured dind without intervention.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -16,6 +16,9 @@ binderhub:
   registry:
     enabled: true
 
+  dind:
+    enabled: true
+
   jupyterhub:
     cull:
       timeout: 600


minrk:
I have a hunch: that a build tried to mount the docker.sock before dind was running. Since build mounts `/var/run/dind/docker.sock` as a volume, if the socket doesn't exist a directory will be created in its place. I don't entirely know how this could happen, since we deployed the dind daemonset yesterday and didn't turn it on until today. I also checked prod, since the dind daemonset is running there as well, and the `dind` directories all look correct. So something funky seems to have happened on staging, specifically.

minrk:
I added a node to the staging cluster and it's got docker.sock mounted correctly.

PR #238, opened 2017-12-20
Add bhub #368 #367

betatim:
Also ban a misbehaving client.


choldgraf:
do we have a process to remind ourselves to *unban* this repo eventually?

willingc:
The process as I see it would be to wait for a request from the repo owner to unban when the issue appears resolved on their side.

PR #246, opened 2018-01-03
Redirect playground to play.nteract.io

yuvipanda:
The nteract.io team is happy to maintain that, so we should support
them with a redirect rather than running our own instance.

I checked with @rgbkrk and he was happy with the redirect!


config/prod.yaml
```diff
@@ -55,8 +55,7 @@ redirector:
...
         to: mybinder.readthedocs.io
-
-playground:
-  ingress:
-    hosts:
-      - playground.mybinder.org
+    - type: host
+      host:
+        from: playground.mybinder.org
+        to: play.nteract.io
```
rgbkrk: 🎉

config/staging.yaml
```diff
@@ -53,8 +53,7 @@ redirector:
...
         to: mybinder.readthedocs.io
-
-playground:
-  ingress:
-    hosts:
-      - playground.staging.mybinder.org
+    - type: host
+      host:
+        from: playground.staging.mybinder.org
+        to: play.nteract.io
```
rgbkrk: we could probably host a staging site or allow a query param that uses staging too

xref: https://github.com/nteract/nteract/issues/2125

rgbkrk:
Great!

Sorry about the lack of quick turnaround on some of the features that would have made this easier, we've mostly been focusing on functionality and maintainability of the frontend first.

config/staging.yaml
yuvipanda: That would be great!

choldgraf:
I'm +1 on this, I generally trust the nteract team to make things awesome and I am all for cross-promotion in the jupyter/OS world :-)

choldgraf:
that said, perhaps we should get some +1s from other binder team members too

yuvipanda:
Thanks everyone :)

PR #247, opened 2018-01-03
Setup a static host for serving badges from

yuvipanda:
This stops sending any tracking cookie with the binder badge.

Partial fix for https://github.com/jupyterhub/binderhub/issues/379


choldgraf:
This is great @yuvipanda !

betatim:
This is smart, I am still thinking "ah ok we need a to spin up a server/pod just for this one image" but really we can just setup a new ingress!

PR #261, opened 2018-01-04
add incident report 2018-01-04

willingc:

Issue #187, opened 2017-11-30
tests after deploy can fail due to ingress reloading


minrk:
If the ingress configuration has not fully loaded before launching the test, there is a chance that the ingress configuration will reload during the build test and drop the connection, resulting in a build failure. If we reliably waited until the ingress configuration was loaded, this wouldn't happen (#181).

yuvipanda:
It's possible that nginx-ingress is just not something we can fix for this. Part of their design is they don't route to the stable service IP, but watch for individual pod IPs & reload config when any pod churns. While this is a performance boost (you are eliminating one hop on iptables), it also means there's gonna be tons of reloads...

So our options are:

1. Make our clients (JS and tests) more resilient to dropped connections (this we should do anyway)
2. Try moving to a different ingress controller

minrk:
I agree we should make clients resilient to dropped connections. It's really the server right now that's not resilient, because a connection cannot be resumed - once a request for launch is initiated, a reconnect (which js eventstream will do automatically, we learned) will be treated as a *new* request for launch. We handle this well in terms of avoiding duplicate builds, but not avoiding multiple launches.

For the tests we have here, we could punt with a longer sleep to wait for the ingress to reload.

willingc:
Looks like sleep is now set to 10. Probably makes sense in the short term to increase the sleep pending a solution to multiple launches. Do we have a sense of how much longer the sleep should be to catch most cases: 20, 30?

minrk:
Seems like it takes a while. We could try 30 or 60.

willingc:
Pick your favorite or split the difference at 45 ;-)

minrk:
If we have the logs of a failing test on Travis, we might be able to see an example for how long it took for the failing command to be reached.

A travis_retry on the tests would also work, because once this has been interrupted once, it won't happen again until the next deploy.

willingc:
Looking at the logs of a few failing tests, they seem to be all over the place in terms of time: ~30s, ~30s, ~600s, ~300s

yuvipanda:
I *think* https://github.com/jupyterhub/mybinder.org-deploy/pull/258 should also help.

yuvipanda:
Going to close this as fixed now since we haven't had an issue with this for a while!

PR #258, opened 2018-01-04
Attempt to wait for deployments to turn over before finishing deploy

yuvipanda:
--wait does not work well enough for our use cases.

PR #262, opened 2018-01-05
band-aids for false-reporting tests

minrk:
Right now, we have an issue where deployments can fail (Travis is red) misleadingly because of nginx dropping connections.

This is tricky right now, because all deployments are 'finished' when this occurs, because nginx appears to reload configuration at runtime, not as part of any initialization that can be awaited (#187).

The band-aids are twofold:

1. Wait longer after finishing deployment before running tests.
   This is because we don't seem to have a way to directly wait for nginx configuration to load.
2. Re-run tests with travis_retry if they fail.
   This accounts for potentially flaky tests, currently failing.
   With the current failure mode, trying tests again immediately after failure will always work.
   This is what we have been doing manually by retriggering travis, but is now automatic and more efficient,
   since it only re-runs the tests and not the whole deploy.
   Doing the retry automatically removes the decision from humans (see #261).

These are not ideal, and should be removed if/when we can find a way to reliably wait for the nginx configuration to have reloaded prior to launching the tests.

### DIFF
--- a/travis/travis-script.bash
+++ b/travis/travis-script.bash
@@ -43,9 +43,13 @@ python3 ./deploy.py deploy ${TARGET}
 # FIXME: add readiness probe so that deploy doesn't finish before
 # hub is available.
 # For now, sleep to give deploy a chance to warm up.
-sleep 10
+# The current reason for this sleep is that the nginx configuration reload drops connections,
+# which occurs *after* deployment is fully ready.
+# There does not appear to be a way to query whether nginx has finished updating.
+sleep 30
 
 # Run some tests to make sure we really did pass!
-py.test --binder-url=${BINDER_URL} --hub-url=${HUB_URL}
+# Currently retrying to account for nginx dropping connections.
+travis_retry py.test --binder-url=${BINDER_URL} --hub-url=${HUB_URL}
 
 echo "Done!"


choldgraf:
I believe that @yuvipanda has tackled a part of this now...is that true?

yuvipanda:
@minrk I merged #258 which should do some of these, but unsure if it's enough. Am happy for these to go in too.

yuvipanda:
Long term I think we need to switch to a different ingress provider than nginx. I'll open an issue.

minrk:
Great! I though #258 wasn't enough because the nginx reload happened at runtime and wasn't awaitable, but maybe we just weren't waiting on the right things at that was all the fix we needed. Let's hold this one until if/when a deploy fails, which will indicate whether 258 solved everything. If several deploys (specifically ones that change the binderhub and/or repo2docker images) succeed in a row, then #258 ought to have been enough, and this band-aid will be unnecessary (yay!).

But if we see a familiar red X from Travis, let's pull this in so we can focus on other things for a bit.

yuvipanda:
https://github.com/jupyterhub/mybinder.org-deploy/pull/273 adds some additional fixes (discovered when we deployed some dind changes) for better deployment experience

yuvipanda:
https://github.com/jupyterhub/mybinder.org-deploy/pull/275 should help too

yuvipanda:
I'm going to close this now, since we seem to have fixed the failures for now! This bandaid is still useful if we start running into flakiness again, so good to keep in mind.

Thanks for all the work, @minrk!

minrk:
Hooray for not needing band-aids!

PR #275, opened 2018-01-09
Bump helm timeout to 600s

yuvipanda:
Currently it is at 300s, which can fail sometimes when large
images need to be pulled.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -17,6 +17,7 @@ def deploy(release):
         release,
         'mybinder',
         '--wait',
+        '--timeout', '600',
         '-f', os.path.join('config', release + '.yaml'),
         '-f', os.path.join('secrets', 'config', release + '.yaml')
     ]


choldgraf:
I kinda feel like we should throw some kind of warning if it looks like the image is really big...it seems like something that often bites people when they could easily make the image smaller if they knew that was the problem

yuvipanda:
This is for the images we deploy - binderhub, Prometheus, nginx, grafana,
etc. Not for user images.

On Jan 8, 2018 5:37 PM, "Chris Holdgraf" <notifications@github.com> wrote:

I kinda feel like we should throw some kind of warning if it looks like the
image is really big...it seems like something that often bites people when
they could easily make the image smaller if they knew that was the problem

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/jupyterhub/mybinder.org-deploy/pull/275#issuecomment-356152405>,
or mute the thread

choldgraf:
sorry that was an off-topic point, I just saw the PR title and the thought popped into my head :-P

PR #286, opened 2018-01-18
Increase Ingress replicas count

betatim:
Increasing the replica count on prod and reducing on staging. Staging is mostly to check that setting individual values for prod and staging works.

I remember a while back we had to increase the number of ingress replicas because "of reasons". There doesn't seem to be anything obviously wrong with the cluster but people can't connect to their spawned pods. This seems like a cheap thing to do.

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -45,6 +45,7 @@ nginx-ingress:
   controller:
     service:
       loadBalancerIP: 35.202.202.188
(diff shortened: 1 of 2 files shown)


betatim:
Replica count on staging has gone down to four. Built a random repository (epifanio/geosciences-250279) from scratch on staging. Build worked, and after a while also connected to the notebook tree view.

PR #290, opened 2018-01-19
updating incident template

choldgraf:
I thought it might be useful if we had a section for adding links to relevant PRs/Issues in the incident template. @willingc what do you think?


choldgraf:
done, LMK!

PR #294, opened 2018-01-23
Modify incident report template a little bit

yuvipanda:
Take inspiration from the Google SRE guide at
https://landing.google.com/sre/book/chapters/postmortem.html

Changes are:

1. TZ should be in whatever is most convenient, not always PST
2. Add a 'Lessons learnt' section
3. Remove 'Supporting Issues/PRs' section, instead every action item
   should have an issue linked to it, and other issues should be linked
   inline as required. Converting Action Items to Issues makes them much
   more visible and forces us to deal with them, when often they can get
   lost without that.
4. Added a bunch of descriptive text to the subheadings

### DIFF
--- a/docs/source/incident-reports/template-incident-report.md
+++ b/docs/source/incident-reports/template-incident-report.md
@@ -1,16 +1,18 @@
 # Template for reports
 
-## {{ incident date: yyyy-mm-dd }}, {{ incident name }}
+# {{ incident date: yyyy-mm-dd }}, {{ incident name }}
 
 ## Summary
 
+Quick summary of what user impact was, how long it was, and how we fixed it.
+
 ## Timeline
 
-All times in PST
+All times in {{ most convenient timezone}}
 
 ### {{ yyyy-mm-dd hh:mm }}
 
-Start of incident.
+Start of incident. First symptoms, possibly how they were identified.
 
 ### {{ hh:mm }}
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
As the folks who did the last few incident reports, let me know if this is an improvement, @choldgraf @willingc @betatim! The only thing I feel strongly about is the GitHub issue per Action item thing.

I also think incident reports are something that is ripe for automation. We used to have a bot that would record all lines starting with '!log' on IRC, so you can then just do that as you are debugging and copy paste a timeline into your report after.

yuvipanda:
https://github.com/jupyterhub/mybinder.org-deploy/pull/297 is an example. I can go through the other incident reports and file issues for those missing them.

minrk:
[this script](https://gist.github.com/b79363d0468b00ee7edb7c089874f998) collects messages starting with `!log` from the gitter API. Example output:

```markdown
Gitter log events since 2018-01-12 09:24 (UTC):
- 2018-01-18 16:29 (UTC) @yuvipanda: kubectl --namespace=prod get pod -l heritage=jupyterhub | grep -P '[6-9]h$' | awk '{ print $1; }' | xargs -L1 kubectl --namespace=prod delete pod
- 2018-01-18 16:48 (UTC) @yuvipanda: delete hub pod with  kubectl --namespace=prod delete pod hub-7649b9bf8-d769g
- 2018-01-18 17:05 (UTC) @yuvipanda: kubectl cordon gke-prod-a-ssd-pool-972e4826-2fhq
- 2018-01-20 11:42 (UTC) @betatim: just stopped the node -0j70, it had no pods left that were named jupyter-*
- 2018-01-20 12:09 (UTC) @betatim: the node came right back after i deleted it so i think with autoscaler enabled deleting a node is not the way to reduce the size of the cluster. What is the right way of doing it? Talk to @yuvipanda at some point
```

choldgraf:
I think this is a great addition! No real changes to the text itself. I wonder if it'd be possible to link minrk's script at the top of the template for reference, that's quite useful!

willingc:
Nice idea @choldgraf. Instead of the top of the page, let's add a link to the script just below the timeline heading but before the first timeline entry.

PR #300, opened 2018-01-24
bumping binderhub + packing users and reducing memory ceiling

choldgraf:
This gets a new version of BinderHub which lets us change how users are distributed across nodes. It then asks kubernetes to pack them onto fewer nodes and lowers the memory limit. Still needs the BinderHub version bump which I'll add once tests pass. cc @yuvipanda

### DIFF
--- a/mybinder/requirements.yaml
+++ b/mybinder/requirements.yaml
@@ -12,5 +12,5 @@ dependencies:
    version: 0.1.12
    repository: https://kubernetes-charts.storage.googleapis.com
  - name: binderhub
-   version: 0.1.0-fbd816c
+   version: 0.1.0-3f81760
    repository: https://jupyterhub.github.io/helm-chart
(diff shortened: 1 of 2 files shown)


yuvipanda:
I think the limit and guarantee should match per the proposal in #299

choldgraf:
doesn't this match that? You suggested either 1G or 2G. I did a quick look of the current pod usage and saw a number of pods that were > 1G so decided to make 2G the limit:

image

yuvipanda:
@choldgraf the gaurantee and limit should be the same, so I think they should both be 2G. Currently limit is 2G but guarantee is 1G.

choldgraf:
ohhh I see - I missed that point in the issue, fixed!

yuvipanda:
I noticed that this wasn't actually packing our pods, and some investigation yielded https://github.com/jupyterhub/mybinder.org-deploy/pull/304 which should fix that.

PR #303, opened 2018-01-25
grafana: install histogram panel

minrk:
could help in visualizing histogram data

which we are using for timing build/launch/etc.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -163,6 +163,7 @@ grafana:
       annotations:
         kubernetes.io/ingress.class: nginx
         kubernetes.io/tls-acme: "true"
+    installPlugins: mtanda-histogram-panel
   serverConfigFile:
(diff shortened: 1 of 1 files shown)


betatim:
Merge?

minrk:
I would like that, yes :)

choldgraf:
@minrk what's a pattern we can use with this to test it out on https://grafana.staging.mybinder.org/?orgId=1 before merging into prod?

betatim:
https://grafana.mybinder.org/render/dashboard-solo/db/main-dashboard?refresh=1m&orgId=1&from=1516870983117&to=1516914183117&panelId=4&width=1000&height=500&tz=UTC%2B01%3A00

Any idea what happened there? This is from grafana.mybinder.org so really shouldn't be connected to this PR, but this was the only PR merged this afternoon?

PR #308, opened 2018-01-31
Docs for https and meta charts

choldgraf:
This adds documentation about the mybinder.org helm meta chart, as well as HTTPS setup for the site.

### DIFF
--- a/docs/source/production_environment.md
+++ b/docs/source/production_environment.md
@@ -71,6 +71,77 @@ Related repositories used by the [mybinder.org][] service are:
    repository is converted into a docker image to be run for the user,
    go to [repo2docker][].
 
+## The Deployment Helm Meta Chart
+
+BinderHub is deployed using a Kubernetes Helm Chart, which is a specification
+for instructing Kubernetes how to deploy particular applications. Sometimes,
+applications depend on others in order to function properly, similar to how
+a package might depend on other packages (e.g., Pandas depends on Numpy).
+These dependencies are specified with a Helm "Meta Chart".
+
+For example, let's say that you'd like to begin using Prometheus in your
+Kubernetes deployment. Since Prometheus has a helm chart for deploying it
+on Kubernetes, we can add it as a dependency in a Helm Meta Chart. We'd
+create a file called `requirements.yaml` and put the following in it:
+
+```yaml
+dependencies:
+- name: prometheus
+  version: 4.6.16
+  repository: https://kubernetes-charts.storage.googleapis.com
+```
+
+This also allows us to pin a *version* of Prometheus, which improves
+reliability of the site.
+
+```note::
+It is still possible to deploy each of these applications on their own *without*
+a Meta Helm Chart, this is simply a way of clustering dependencies together
+and simplifying the deployment structure.
+```
+
+Another benefit of Meta Charts is that you can use a single configuration
+file (`config.yaml`) with multiple Helm Charts. For example, look at the
+[BinderHub Helm Chart](https://github.com/jupyterhub/mybinder.org-deploy/blob/staging/mybinder/values.yaml). Note that there are multiple
+top-level sections (e.g., for [jupyterhub](https://github.com/jupyterhub/mybinder.org-deploy/blob/5aa6dde60c9b5f3012686f9ba2b23b176c19b224/mybinder/values.yaml#L53) and for [prometheus](https://github.com/jupyterhub/mybinder.org-deploy/blob/5aa6dde60c9b5f3012686f9ba2b23b176c19b224/mybinder/values.yaml#L204)) and that each section
+has a corresponding entry in the Helm Meta Chart. In this way, we can provide
+configuration for each dependency of BinderHub without needing a separate
+file for each, and we can deploy them all at the same time.
+
+For more information, we recommend investigating the structure of the
+[Binder Helm Meta Chart](https://github.com/jupyterhub/mybinder.org-deploy/blob/staging/mybinder/requirements.yaml). In addition, the Kubernetes
+organization keeps a [curated list of Helm Charts](https://github.com/kubernetes/charts) that you can specify in
+your Meta Chart in order to deploy different types of applications.
+
+## HTTPS configuration for `mybinder.org`
+
+Using HTTPS requires having a signed certificate. BinderHub uses [kube-lego](https://github.com/jetstack/kube-lego),
+a tool that obtains and deploys a free *Let's Encrypt* certificate automatically.
+This section describes how to use `kube-lego` to configure and deploy HTTPS support.
+
+`kube-lego` provides 90 day SSL certificates for `mybinder.org` through
+the [letsencrypt](https://letsencrypt.org/) service. As the 90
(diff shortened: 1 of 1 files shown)


choldgraf:
@yuvipanda this is the remainder from the de-bus-factoring knowledge dump we had a few weeks back, I realized we never made a PR for this bit

docs/source/production_environment.md
```diff
@@ -71,6 +71,80 @@ Related repositories used by the [mybinder.org][] service are:
...
+BinderHub is deployed using a Kubernetes Helm Chart, which is a specification
+for instructing Kubernetes how to deploy particular applications. Sometimes,
+applications depend on others in order to function properly, similar to how
+a package might depend on other packages (e.g., Pandas depends on Numpy).
+These dependencies are specified with a Helm "Meta Chart".
+
+For example, let's say that you'd like to begin using Prometheus in your
+Kubernetes deployment. Since Prometheus has a helm chart for deploying it
+on Kubernetes, we can add it as a dependency in a Helm Meta Chart. We'd
+create a file called `requirements.yaml and put the following in it:
```
betatim: missing ` after the .yaml

docs/source/production_environment.md
```diff
@@ -71,6 +71,80 @@ Related repositories used by the [mybinder.org][] service are:
...
+- name: prometheus
+  version: 4.6.16
+  repository: https://kubernetes-charts.storage.googleapis.com
+```
+
+This also allows us to pin a *version* of Prometheus, which improves
+reliability of the site.
+
+```note::
+It is still possible to deploy these applications on their own *without*
```
betatim: these -> each of these

docs/source/production_environment.md
```diff
@@ -71,6 +71,80 @@ Related repositories used by the [mybinder.org][] service are:
...
+a Meta Helm Chart, this is simply a way to clustering dependencies together
+and simplifying the deployment structure.
+```
+
+Another benefit of Meta Charts is that you can use a single configuration
+file (`config.yaml`) with multiple Helm Charts. For example, look at the
+[BinderHub Helm Chart](https://github.com/jupyterhub/mybinder.org-deploy/blob/staging/mybinder/values.yaml). Note that there are multiple
+top-level sections (e.g., for [jupyterhub](https://github.com/jupyterhub/mybinder.org-deploy/blob/staging/mybinder/values.yaml#L53) and for [prometheus](https://github.com/jupyterhub/mybinder.org-deploy/blob/staging/mybinder/values.yaml#L203)) and that each section
```
betatim: make the links refer to a specific commit? IMHO that is better than a generic link to a particular line which will end up pointing to a blank line or some totally unrelated bit of code.

(press `y` after following one of the URLs above to get a URL with a hash in it)

betatim:
Some small nits to pick, otherwise 👍

docs/source/production_environment.md
choldgraf: > (press y after following one of the URLs above to get a URL with a hash in it)

 *back away in awe and terror...what kind of magical github wizardry is this*

choldgraf:
@betatim comments addressed - that `y` trick is awesome

docs/source/production_environment.md
betatim: This is how you know you spend too much time staring at github.com ;)

betatim:
Leaving this for a bit to see if @yuvipanda wants to comment, otherwise going to merge it soon.

docs/source/production_environment.md
```diff
@@ -71,6 +71,80 @@ Related repositories used by the [mybinder.org][] service are:
    repository is converted into a docker image to be run for the user,
    go to [repo2docker][].
 
+## The Helm Meta Chart
```
willingc: I'm not sure that I like the term Helm "Meta" Chart. I'm googling and note finding this as a standard K8 term (or my google-fu is off). "Meta" is a pretty overloaded term without a concrete meaning.

What I'm wondering is if this would be better as a section as "Dependencies on Other Charts", "How to use other dependency charts in a Helm Chart", Subcharts and a Parent Chart.

docs/source/production_environment.md
```diff
@@ -71,6 +71,80 @@ Related repositories used by the [mybinder.org][] service are:
...
+file for each, and we can deploy them all at the same time.
+
+For more information, we recommend investigating the structure of the
+[Binder Helm Meta Chart](https://github.com/jupyterhub/mybinder.org-deploy/blob/staging/mybinder/requirements.yaml). In addition, the Kubernetes
+organization keeps a [curated list of Helm Charts](https://github.com/kubernetes/charts) that you can specify in
+your Meta Chart in order to deploy different types of applications.
+
+## HTTPS configuration for `mybinder.org`
+
+Using HTTPS requires having a signed certificate. In the past this required
```
willingc: ~~In the past ... well).~~ BinderHub uses kube-lego, a tool that obtains and deploys a free Let's Encrypt certificate automatically for you. This section describes how to use kube-lego to configure and deploy HTTPS support.

docs/source/production_environment.md
choldgraf: Yeah - I agree with you. How about we rename the section to "Dependencies on other charts" and within there we can mention that this is done with something called a "Meta" chart in k8s world?

betatim: What does the helm world use for this? I don't know the answer but probably worth trying to mirror their jargon. I like "How to use other charts as a dependency in a Helm Chart"

choldgraf: I think the helm community uses "meta chart" https://www.google.com/search?ei=BxN6Wtq9GIKC0wKulZGAAg&q=helm+%22meta+chart%22&oq=helm+%22meta+chart%22&gs_l=psy-ab.3...14718.15661.0.15880.2.2.0.0.0.0.84.162.2.2.0....0...1c.1.64.psy-ab..0.1.78...35i39k1.0.P52wtw8YrIA

choldgraf: but it doesn't look like this is anything "official", just what tends to pop up

yuvipanda: Meta chart is also the name (in the k8s slack / in-person-conversations) I've heard for this specific thing - where there's not much content in the chart itself, and it is primarily used only for dependency management. Debian uses 'meta-packages' for the same concept (https://packages.debian.org/stable/metapackages/ for example).

The distinction is between binderhub (which uses the helm-chart as a dependency) vs mybinder.org (which uses dependencies only for deployment).

We can also try using other terms if we want to, since as you say 'meta' is overloaded. 'Deployment chart' perhaps?

willingc: I would be cool with 'Deployment chart' or 'Meta chart for deployment' or 'Deployment Meta chart' ;-)

yuvipanda: I like Deployment Meta Chart :)

willingc: @choldgraf You choose from ^^

choldgraf:
ok, changes made - LMKWYT!

choldgraf:
(I think this one is just waiting for somebody to hit the green button :-) )

PR #319, opened 2018-02-07
Temporarily block berkeley's nbinteract repos

yuvipanda:
They seem to be launching on page load, overwhelming our
servers


minrk:
Deploying https://github.com/jupyterhub/binderhub/pull/370 should reduce the impact of events like this in the future

PR #326, opened 2018-02-07
Unban berkeley nbinteract repos

yuvipanda:
They now have a 'Run Widgets' button that needs pressing


minrk:
Let's keep an eye on it!

PR #340, opened 2018-02-08
instructions for deleting pods matching name

choldgraf:
for the SRE guide, this is a useful / common command!

### DIFF
--- a/docs/source/command_snippets.md
+++ b/docs/source/command_snippets.md
@@ -44,3 +44,17 @@ pods. The `--no-headers` asks kubectl to not print column titles as a header.
 The `awk` command selects the 7th column in the output (which is the node name).
 The sort / uniq / sort combination helps print the number of pods per each node in
 sorted order.
+
+## Delete all pods that match a given name
+
+Sometimes you want to delete all the pods for a given repository. The easiest
+way to do this is to name-match the part of the pod name that corresponds to
+the repo (since there will be a bunch of random characters as well).
+
+Here's a python script that will match pods with a given name and delete them
+(you need to uncomment the line to actually do the deleting). The script can
+be run by changing your directory to the root of this repository, then running
+
+```
+python scripts/delete_pods_matching_name.py <your-query> --delete
+```
--- a/scripts/delete_pods_matching_name.py
+++ b/scripts/delete_pods_matching_name.py
@@ -1,0 +1,21 @@
+from subprocess import check_output, call
+import argparse
+
+parser = argparse.ArgumentParser(description="Quickly delete pods with a name that includes `match`.")
+parser.add_argument("match", help="A string to match to pod names.")
+parser.add_argument("--delete", dest='do_delete', action='store_true', help="Whether to run the `kubectl delete` command.")
+args = parser.parse_args()
+
+match = args.match
+do_delete = args.do_delete
+# The -l matches on labels, and all user pods have the label component=singleuser-server
+pods = check_output('kubectl --namespace=prod get pod -o name -l component=singleuser-server'.split())
+
+for line in pods.split('\n'):
+    if match in line:
+        name = line.split()[0]
+        command = 'kubectl --namespace=prod delete {}'.format(name)
+        if do_delete is True:
+            call(command.split())  # When you know it'll work!
+        else:
+            print(command)  # For making sure your command is correct
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ mybinder/charts
 mybinder/requirements.lock
 
 .ipynb_checkpoints
+
+docs/_build
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -39,6 +39,15 @@ source_parsers = {
     '.md': 'recommonmark.parser.CommonMarkParser',
 }
 
+# Use autostructify so we can get some rST in the markdown
+from recommonmark.transform import AutoStructify
+def setup(app):
+    app.add_config_value('recommonmark_config', {
+      'enable_eval_rst': True
+    }, True)
+    app.add_transform(AutoStructify)
+
+
 # The suffix(es) of source filenames.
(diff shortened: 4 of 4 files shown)


docs/source/command_snippets.md
```diff
@@ -44,3 +44,23 @@ pods. The `--no-headers` asks kubectl to not print column titles as a header.
...
+
+## Delete all pods that match a given name
+
+Sometimes you want to delete all the pods for a given repository. The easiest
+way to do this is to name-match the part of the pod name that corresponds to
+the repo (since there will be a bunch of random characters as well). Here's
+a python script that will match pods with a given name:
+
+```python
+from subprocess import check_output, call
```
yuvipanda: IMO you should make this into a script that can be run like:

./scripts/delete-repos <something>

and put that in repo.

choldgraf:
@yuvipanda how about now? uses a python script and `literalincludes` it

docs/source/command_snippets.md
```diff
@@ -44,3 +44,18 @@ pods. The `--no-headers` asks kubectl to not print column titles as a header.
...
+Sometimes you want to delete all the pods for a given repository. The easiest
+way to do this is to name-match the part of the pod name that corresponds to
+the repo (since there will be a bunch of random characters as well).
+
+Here's a python script that will match pods with a given name and delete them
+(you need to uncomment the line to actually do the deleting). The script can also
+be found in `/scripts/delete_pods_matching_name.py`.
+
+```eval_rst
+.. literalinclude:: ../../scripts/delete_pods_matching_name.py
```
yuvipanda: I think this snippet should just be:

python scripts/delete_pods_matching_name.py <name

instead. I think people are much more likely to copy paste it rather than see the source, copy paste it into a file, then execute it. You can always see the source from the file directly.

scripts/delete_pods_matching_name.py
```diff
@@ -0,0 +1,10 @@
+from subprocess import check_output, call
+
+match = 'jupyter-repothats-destroyingbinder'
```
yuvipanda: This should be replaced with:

import sys
match = sys.argv[1]

This allows running the script without modifying it.

scripts/delete_pods_matching_name.py
```diff
@@ -0,0 +1,10 @@
...
+pods = check_output('kubectl get pod --namespace=prod'.split())
+for line in pods.split('\n'):
+    if match in line:
+        name = line.split()[0]
+        command = 'kubectl delete pod --namespace=prod {}'.format(name)
+        print(command)  # For making sure your command is correct
+        # call(command.split())  # When you know it'll work!
```
yuvipanda: Uncomment this? :)

choldgraf: hmm - what if we add a kwarg like "--delete"? The reason I had it uncommented was to urge caution on behalf of the user...think that's overkill?

yuvipanda: You can do that too, at which point you might want argparse.

choldgraf:
how about the latest push?

scripts/delete_pods_matching_name.py
```diff
@@ -0,0 +1,19 @@
...
+parser.add_argument("--delete", dest='do_delete', action='store_true', help="Whether to run the `kubectl delete` command.")
+args = parser.parse_args()
+
+match = str(args.match)
+do_delete = bool(args.do_delete)
+pods = check_output('kubectl get pod --namespace=prod'.split())
+for line in pods.split('\n'):
+    if match in line:
+        name = line.split()[0]
+        command = 'kubectl delete pod --namespace=prod {}'.format(name)
```
yuvipanda: Can you put --namespace=prod just after kubectl?

scripts/delete_pods_matching_name.py
```diff
@@ -0,0 +1,19 @@
...
+parser = argparse.ArgumentParser(description="Quickly delete pods with a name that includes `match`.")
+parser.add_argument("match", help="A string to match to pod names.")
+parser.add_argument("--delete", dest='do_delete', action='store_true', help="Whether to run the `kubectl delete` command.")
+args = parser.parse_args()
+
+match = str(args.match)
+do_delete = bool(args.do_delete)
+pods = check_output('kubectl get pod --namespace=prod'.split())
+for line in pods.split('\n'):
+    if match in line:
```
yuvipanda: This means if a user accidentally says 'match' is 'Running' or 'jupyter' or 'dind' or 'nginx' it'll delete a lot of unintentional pods. I'd recommend changing the command in line 11 instead to:

kubectl --namespace=prod get pod -o name -l component=singleuser-server

The -l matches on labels, and all user pods have the label component=singleuser-server.

scripts/delete_pods_matching_name.py
yuvipanda: And remove the pod, in conjunction with my other comment?

scripts/delete_pods_matching_name.py
```diff
@@ -0,0 +1,19 @@
+from subprocess import check_output, call
+import argparse
+
+parser = argparse.ArgumentParser(description="Quickly delete pods with a name that includes `match`.")
+parser.add_argument("match", help="A string to match to pod names.")
+parser.add_argument("--delete", dest='do_delete', action='store_true', help="Whether to run the `kubectl delete` command.")
+args = parser.parse_args()
+
+match = str(args.match)
```
yuvipanda: These are already string and bool (because of the store_true), you don't need to cast them.

You can also just directly use them in the two places they are used instead of assigning temporary variables.

yuvipanda: (I don't feel strongly about this one, just find it confusing, but that might be just me!)

choldgraf:
updated - LMK!

yuvipanda:
Thanks, @choldgraf!

PR #341, opened 2018-02-08
Add script to kill old user pods

yuvipanda:


docs/source/command_snippets.md
```diff
@@ -58,3 +58,21 @@ be run by changing your directory to the root of this repository, then running
...
+removes nodes from the kubernetes cluster when they have
+been 'empty' for more than 10 minutes However, we
+have issues where some pods get 'stuck' and never actually
+die, sometimes forever. This causes nodes to not be
+killed automatically.
+
+You can use the 'kill-old-user-pods.py' script to kill them. You can run it with:
+
+```bash
+./scripts/kill-old-user-pods.py <number-of-hours> --delete
```
betatim: rename it to `old-user-pods.py` to encourage people to use it for listing as well as deleting? If I hadn't jsut read the doc string in the script to learn that without --delete it just prints I would have scrolled right past this script and cooked some kind of `awk + grep` one liner up to get a list.

betatim:
LGTM

docs/source/command_snippets.md
yuvipanda: Done

PR #349, opened 2018-02-08
updating yuvi's old pod script

choldgraf:
@yuvipanda some small improvements to the old pods script


yuvipanda:
Thanks @choldgraf

PR #353, opened 2018-02-08
Bump Prometheus + configure it better

yuvipanda:
- No longer need to maintain our own scrape config, since
  cadvisor is now set up properly by chart
- Switch to using large SSD disk (than spinning disk) for
  prometheus data, gives orders of magnitude better performance.
  (Only on production)
- Retain data for 60d instead of default 15d, since we have
  the size and performance to do that now (Only on production)
- Bump chart version bringing us Prometheus 2.1, with big performance
  and other enhancements.

Ref #350


choldgraf:
image

👌 👌 👌 👌 👌 👌

PR #355, opened 2018-02-09
Lower the CHP log verbosity to ERROR from DEBUG

betatim:
Reduce the amount of logs generated by the CHP. Related to #309.

Is "error" too little logging? Thoughts?


minrk:
Given that we are logging a bunch at the nginx level, I think this is appropriate. CHP's logs aren't that interesting, as everything it proxies also logs requests.

PR #360, opened 2018-02-12
adding incident report 2018/02/12

choldgraf:
This is the hub launch failure incident. There may still need to be some information added as we investigate the root cause of this, but there are already some takeaways in terms of team process etc.

(feel free to edit/push if you like)

cc @yuvipanda @willingc

### DIFF
--- a/docs/source/incident-reports/2018-02-12-launch-fail.md
+++ b/docs/source/incident-reports/2018-02-12-launch-fail.md
@@ -1,0 +1,67 @@
+# 2018-02-12, Hub Launch Fail
+
+## Summary
+
+Binder was successfully building user pods, but was then failing to direct
+users to the built pods. It was fixed by deleting the `binder` and `hub` pods.
+
+## Timeline
+
+All times in PST
+
+### 2018-02-12 14:03
+
+We realized that there's a high usage on the mybinder deployment. Tried
+building a repository and it would get to the "launching" step then never
+proceed further. Eventually it'd return a "your image took too long to launch" error.
+
+### 2018-02-12 14:06
+
+From the [grafana board](https://grafana.mybinder.org), we realized that in the
+"Launch Times Summary" plot we showed *all* pods as failing to launch.
+
+### 14:08
+
+We delete the `binder` and `hub` pods in the `prod` deployment.
+
+### 14:09
+
+Two people confirm that their pods now build and launch fine, Grafana also
+shows successful "Launch Times Summary" data.
+
+## Lessons learnt
+
+### What went well
+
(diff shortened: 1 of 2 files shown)


choldgraf:
I'm not sure, and I suspect that @yuvipanda did this just out of a hunch than anything else :-)  this is actually what inspired me to open #361 , since I think often-times things like "delete the hub pod" are safe bets we can try that *might* fix the problem.

For this PR specifically, I think we need to do a bit more investigating to figure out why exactly this is what fixed the problem.

yuvipanda:
@willingc, @choldgraf  trying to re-trace my thought process...

1. Ok, so the issue could be 1. In the Kubernetes Cluster, 2. In JupyterHub (launching processes + proxy), 3. In BinderHub (talking to the JupyterHub API)
2. I launched a new pod (via mybinder.org), and did `kubectl --namespace=prod get pod`. This showed the pod was launching. I tried in a few seconds, and pod was ready. This made me feel that the problem was probably not the cluster.
3. But In the UI, it was still at 'launching', even after the pod was ready. This made me think the problem was either in the hub (it didn't detect that pod has launched) or in binderhub (it wasn't redirecting users after hub has launched)
4. I looked at the binderhub and jupyterhub logs, and nothing really seemed 'obvious'ly wrong. I was looking for lines 'shaped' like stack traces, to notice errors. This is usually surprisingly effective, but not this time!
5. I took a guess and just deleted the hub and binderhub pods, mostly because the service has been out for almost an hour, I was fairly certain the problem was in one of these two, stackdriver would still preserve logs if I wanted to dig deeper, and that the Deployment object would recreate the Pods after I had deleted them. 
6. My fallback was, if the deletion had not worked, I would look through the stackdriver logs and try to figure out if the bug was in jupyterhub or binderhub. But thankfully, the deletion made it work - this makes me think it is a race condition or some other bug in binderhub / jupyterhub. We can still dive into it by inspecting the logs in stackdriver, if we want. Thankfully the deletion worked and we were back up.

I hope this is helpful! I could push into this PR weaving this narrative into it too if you'd like.

yuvipanda:
Gonna merge now, can edit later!

Thanks for putting it together, @choldgraf!

@willingc I hope my comment here answered your question!

willingc:
Yep @yuvipanda, answered my question. I was doing some research last night on logging and monitoring. Found out termination messages can be used to provide a bit more info when `kubectl describe` is run on a pod. Helps with quick triage prior to consulting detailed logs.

betatim:
Thanks for your thought process Yuvi!

PR #367, opened 2018-02-13
Automatically deploy to prod if tests succeed on staging

yuvipanda:
The original reasoning for having staging and prod deployments
be both manual was we didn't have automated tests we could rely on.
So a human had to manually test staging before deploying to prod.
It was a stop-gap until we had better automation.

Now we have automated tests, and so the stop gap should go away.

This removes an entire step from our deployment workflow,
which should make it easier for a larger group of users to do.
It'll also take less time than it does now.

Someone will still need to watch for deployment failures, but since
the deployment will stop if staging deploy fails, that should be ok.

Ultimately, we need to make deployments easy enough and painless
enough that number of people doing deploys is higher and more
diverse than now, and I think this is a good step in that direction.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,13 +36,11 @@ deploy:
   script: ./travis/travis-script.bash
   on:
     branch:
-      - staging
-      - prod
+      - master
 
 branches:
   only:
-    - staging
-    - prod
+    - master
 
 notifications:
   webhooks:
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 # mybinder.org-deploy
 
-Staging: [![Staging Build Status](https://travis-ci.org/jupyterhub/mybinder.org-deploy.svg?branch=staging)](https://travis-ci.org/jupyterhub/mybinder.org-deploy) |
-Production: [![Production Build Status](https://travis-ci.org/jupyterhub/mybinder.org-deploy.svg?branch=prod)](https://travis-ci.org/jupyterhub/mybinder.org-deploy/branches)
+Deployment: [![Staging Build Status](https://travis-ci.org/jupyterhub/mybinder.org-deploy.svg?branch=master)](https://travis-ci.org/jupyterhub/mybinder.org-deploy)
 
 Deployment, configuration, and Site Reliability documentation files for the
 public [mybinder.org][] service.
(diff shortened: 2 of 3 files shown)


yuvipanda:
After this gets merged, I'll also archive the 'staging' and 'production' branches to tags, and we'll just be using the familiar 'master' branch.

yuvipanda:
After reading https://increment.com/development/center-stage-best-practices-for-staging-environments/ from @willingc, I think it'll be useful to have a way to deploy something to just staging (not to just prod!). This I think we can do by looking for a [staging-only] title in the PR. I'll implement that as a separate PR though.

yuvipanda:
I've also updated README to document the simpler workflow.

choldgraf:
this looks great to me - I agree we should have the ability to just deploy to staging if need be, but that's a separate issue from this PR

yuvipanda:
Thanks, @choldgraf! 

I'm gonnna let this one soak for at least 24h to give others a chance to chime in!

minrk:
I think this is great! I like the idea of separately deploying to staging as an explicit step, e.g. for manually verifying something that may not be covered by the tests, or conservatively rolling out a particularly big change, etc.

minrk:
I think there is one small thing this makes more complicated: if Travis gives us the red ❌ what state are we in? e.g. if the tests fail on staging, then staging has been updated to the non-passing state, but prod hasn't been touched. If the tests fail instead on prod (shouldn't happen after passing on staging with reliable tests, but we know that's not 100% of the time), then prod, too is an updated state that hasn't passed the tests. So we've got five states, as far as I can tell:

1. Travis barfed, e.g. failed to install basic things as happens sometimes. Nothing has been deployed anywhere, we can safely relaunch the tests.
2. deploy to staging fails (e.g. unsatisfiable chart bug)
3. tests on staging fail (staging fully deployed, but something's not right)
4. deploy to prod fails (unlikely)
5. tests on prod fail (prod fully deployed, but something's not right. Since the same tests passed on staging, this is *likely* a spurious failure)
6. everything passed! Travis is ✅

Knowing how to identify which state we are in and what to do about it (when in doubt revert, which is covered in the updated troubleshooting steps 👍 ) would be a good thing to make as clear as we can.

willingc:
Nice! This is :heart:

One case along similar lines as Min's comment: A staging only change (A) has been merged but not yet deployed to prod. A new PR (B) is merged into master. 

If tests pass, does A+B get deployed to prod?
If tests fail, is it due to A or B?

yuvipanda:
Followup after this:

- [ ] Explain clearly the various stages of deployment (we don't actually have that documented anywhere)
- [ ] List those stages clearly in the travis deployment script, so it is easy to identify when something failed
- [ ] Have a way to push to staging only.

I'm going to merge this now, and file issues for these later!

Thanks for the feedback, everyone!

yuvipanda:
Also, @willingc:

>If tests pass, does A+B get deployed to prod?

Yep, A+B get deployed to prod together

> If tests fail, is it due to A or B?

Hard to tell, so we should have guidelines about when to use staging only deploys, and have some sort of 'lock' so if one person has a staging-only deploy they need to do a prod deploy soon (in some fashion) before others can continue. Good catch!

yuvipanda:
I've done the stage explanation document + annotated .travis.yml with it, and opened #403 for staging only deploys!

PR #370, opened 2018-02-14
Describe use of git-crypt for secrets in docs

minrk:
and give an example of using ssh-vault to distribute the shared secret for git-crypt

Would anyone like to give the secret-sharing procedure a try?


choldgraf:
This is awesome btw @minrk , thanks for adding. I'll take a pass through when I have a moment to set up SMTP and will try to make a PR with clarifications etc.

PR #379, opened 2018-02-15
Make green more prominent + add newlines

yuvipanda:


choldgraf:
I want to start an indie hipster band and call it BOLDREVERSEGREEN

choldgraf:
This is all coming together. Kubectl to major Tom will be the breakout hit

PR #391, opened 2018-02-15
Move prod deploy back to deploy than script

yuvipanda:
Turns out travis will merrily continue with other steps if
a step fails!!!! Since we do not want prod to deploy if previous
steps have failed, we will put that in deploy separately

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,6 @@ install:
   - pip install pyyaml pytest pytest-timeout requests
 
 script:
-  - true
-
-before_deploy:
 - mkdir -p ${HOME}/bin
 - |
   # Install gcloud SDK!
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -44,27 +41,24 @@ before_deploy:
...
-- |
-  # Verify prod deployment
-  py.test --binder-url=https://mybinder.org --hub-url=https://hub.mybinder.org
 
 env:
   global:
     - CLOUDSDK_CORE_DISABLE_PROMPTS=1
     - PATH=${HOME}/google-cloud-sdk/bin:${HOME}/bin:${PATH}
 
+# HACK: This is in deploy and not script simply because travis will *not*
```
willingc: Wow. Good catch.

PR #405, opened 2018-02-16
Create annotation in Grafana whenever we do a deployment

yuvipanda:
Helps us record all deployments in a stable place,
and see it in the graphs too!


minrk:
Watching the JupyterLab post go by, and I want to give this PR a great big hug. 🥇

PR #412, opened 2018-02-16
Double the escaping, for double the power

yuvipanda:
Intricate dance of escape sequences and escaping escape
sequences when escape sequences need escaping


choldgraf:
lol this is ridiculous

PR #413, opened 2018-02-16
unbanning nbinteract-image on staging only

choldgraf:


config/staging.yaml
```diff
@@ -18,6 +18,12 @@ binderhub:
           hosts:
             - hub.staging.mybinder.org
 
+  extraConfig:
+    bans: |
```
yuvipanda: you might have to remove the whole segment from values.yaml too. We'll find out!

choldgraf:
it seemed to work!

choldgraf:
(that's from https://www.nbinteract.com/recipes/recipes_nbinteract_graphing.html#)

Issue #416, opened 2018-02-16
Document all the components of mybinder.org that we have installed and use


yuvipanda:
mybinder.org uses several components that are not directly related to BinderHub / JupyterHub. These are usually dependent helm charts. We should document what these are, how they are used, why they are used, etc.

Current list:

- [x] Prometheus
- [x] Grafana
- [ ] nginx-ingress (#488)
- [ ] kube-lego (#488)

yuvipanda:
Hackmd for Grafana: https://hackmd.io/GYIwrAzBDGIKYFoCGSAcBOBAWYATEC6AjAOzDa4kAM6JAbCnNOkA#

Hackmd for Prometheus: https://hackmd.io/BwYwzAjApgZgTAVgLRQEYDZhICxl01AEwgHYkBDATkt3IRnJPTCA#

yuvipanda:
I'll make PRs in a few hours. In the meantime, feel free to make edits!

yuvipanda:
nginx and kube-lego docs in #488

PR #421, opened 2018-02-17
Add metrics + dashboard component docs

yuvipanda:
Ref #416


yuvipanda:
@choldgraf I can't run make html on this locally, can you test before merging!

yuvipanda:
This fails locally with

```
Running Sphinx v1.7.0
loading pickled environment... not yet created
loading intersphinx inventory from https://docs.python.org/objects.inv...
intersphinx inventory has moved: https://docs.python.org/objects.inv -> https://docs.python.org/2/objects.inv
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 20 source files that are out of date
updating environment: 20 added, 0 changed, 0 removed
reading sources... [ 10%] components/dashboards                                                                                                                                                                                                               
Exception occurred:
  File "/home/yuvipanda/code/mybinder.org-deploy/lib/python3.6/site-packages/recommonmark/states.py", line 134, in run_role
    content=content)
TypeError: 'NoneType' object is not callable
The full traceback has been saved in /tmp/sphinx-err-9yluxuu5.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
Makefile:20: recipe for target 'html' failed
make: *** [html] Error 2
```

betatim:
Unsuccessfully tried to work out what causes the error :-/ Deleting all of the first part (up to L25 or so) seems to fix it, but I couldn't narrow it down further. Nothing looks broken.

choldgraf:
ok I think I figured this out. looks like recommonmark has a bug in it so that it can't translate markdown links to rst links if they link to local markdown files (https://github.com/sphinx-doc/sphinx/issues/3800). So I changed those links to be `html` and also fixed the image path so it's in `_static` while I'm at it (it wasn't showing up on my computer)

betatim:
> changed those links to be `html`

How do we avoid getting warnings for those? In #424 one of the minor reasons to move to a .rst was that there were warnings about the "missing" `html` files.

I don't have strong feelings over rst vs md, but I do hate with a passion things that produce "known to be harmless warnings" because the lazy part of me uses that as an excuse to ignore all warnings. So somehow silencing those would be great :)

choldgraf:
I'm not really sure - in general markdown support within sphinx is kinda buggy / unpredictable, and from what I can tell the recommonmark repos aren't super well-maintained :-/

yuvipanda:
Thanks for finding and fixing it, @choldgraf!

PR #424, opened 2018-02-18
Move incident reporting index page to rst

betatim:
This removes some warnings generated during doc build and makes it
easier to keep the two TOCs of incident reports in sync


choldgraf:
🎉

PR #437, opened 2018-02-20
merging the delete pods script

choldgraf:
I took a stab at improving the pod deletion helper scripts. It does the following things:

1. You can now give pod age in fractions of an hour
2. It removes the "delete pods by name" script and renames the `delete old pods" script
3. There is now *one* script that has flags for either matching pod name or pod age. That way you can do both at the same time (e.g. "delete all jupyterlab pods over 30 minutes old")

cc @yuvipanda as maybe you have thoughts on this!


yuvipanda:
Thanks, @choldgraf! Can you update the docs in the SRE guide too?

PR #440, opened 2018-02-21
Merging delete-pod sections and adding a getting started page

choldgraf:
This cleans up the `list pods of a given age / name" sections, and also adds a section for getting started. We can probably improve the latter page but it's a start!


choldgraf:
cc @yuvipanda who suggested I update the docs to match the new delete pods script :-)

PR #441, opened 2018-02-21
redirect /user/... requests for stopped pods back to home

minrk:
adds an nginx container and route in the CHP table at /user so that instead of the Hub getting requests that ultimately 403 when a pod stops, users are redirected back to the home page.

In the future, we can use this to serve a better, more informative error page, rather than a blind redirect

Will deploy this to staging only first, then to prod


minrk:
Once deployed, a request to `hub.staging.mybinder.org/user/xxx` will be redirected to `staging.mybinder.org`.

Testing deployment to staging now.

PR #444, opened 2018-02-21
avoid redirecting API requests to binderhub

minrk:
check for /user/.../api requests and non-gets, and return a 404

we only want to send humans all the way back to the landing page


minrk:
#442 dropped the 3XX response codes on the Hub a bunch, but increased 4XX on binderhub (specifically 405 on OPTIONS requests) caused by redirecting API requests back to binderhub.

https://github.com/jupyterhub/binderhub/pull/462 eliminates the 405 on the OPTIONS request, but we shouldn't be forwarding most of those requests anyway

PR #445, opened 2018-02-21
pod stats notebook in `scripts`

choldgraf:
another helper notebook - this one just lists the current pods and dumps it into a DataFrame. Also has a simple viz to look at the distribution of pod age for a given repo. There are probably a lot of other cool things you could do with the notebook, so this is just a start :-)

at some point it could potentially be merged with the `pod-health` notebook too (if this is too redundant, feel free that we make this merge now)


minrk:
Nice!

PR #446, opened 2018-02-22
Return 404s on non-existent pages instead of redirecting

yuvipanda:
External tools (such as Juno) were using the fact that a
dead server will return 404 to check when they needed to
get a new server. This restores that behavior for now so
we do not break those.


minrk:
FWIW, it was *not* a 404 before, it was a 403 on /hub/login after redirects.

Issue #448, opened 2018-02-22
Check if binder is still running 


navoshta:
_This has been previously discussed in jupyterhub/binderhub#287._

[Juno app](https://itunes.apple.com/app/juno-jupyter-notebook-client/id1315744137) is caching any URL it receives from mybinder.org to avoid requesting extra instanced and to reduce load. As discussed previously, the decision was to simply try accessing the URL and checking if it returns 404 (i.e. it was destroyed) or 200 (i.e. still running). 

> As to seeing if a given user instance is still running, you could also make an API request to the notebook URL you got back. It should 404 if it is dead.

There are plans to introduce a redirect to mybinder.org for destroyed binders, therefore any binder URL will always return 200. Unfortunately, this would trick Juno into thinking every URL is in fact OK and it will never request a new URL for a repo.

While said redirect was in place, I managed to check few things, and it seems I could try using `…/api/status` instead, which wasn’t redirected and returned 404 for timed out binders. Is it something we should use in next Juno release? When are you planning to bring this redirect back?

yuvipanda:
This has been temporarily 'fixed' in #446 and #447, since other projects relying on the API might also rely on this 404 response.

yuvipanda:
@navoshta can you verify if your app is back to working?

navoshta:
@yuvipanda Yes, it's working now. Thanks! Are there plans to bring the redirect back?

minrk:
@navoshta yes, I think you should be using `/api/status`, as discussed in  jupyterhub/binderhub#287. That wouldn't have seen this issue. The redirect-to-200 was for 'likely human' requests (GETs to non-API urls). I think we can avoid restoring the 200 redirect, but I would recommend updating the URL to check anyway.

I think ultimately the dead-pod URL should serve a human-readable page, and the status should probably be 404. It was a 200 as a temporary fix because I didn't want to spend the time to get the patch serving a human-readable error page while we were dealing with an outage, so I redirected to a known page.

For humans, I think the current situation is worse than it was earlier (jupyterhub 403) and during this time (working but inexplicable binder landing page), but a good error page is the right thing to do in the long run, and I'll work on that. I think we can live with nginx 404 for a short period.

minrk:
#461 introduces a more permanent 404 page, so I think you can safely assume that it won't go back to 200. Prior to yesterday, I believe the result was a 403 on `/hub/login` after redirects.

navoshta:
[...]

PR #461, opened 2018-02-22
serve an HTML error page for stopped pods

### LINKED ISSUE #443: better handling of stopped pods
this is related to https://github.com/jupyterhub/binderhub/issues/457 and https://github.com/jupyterhub/binderhub/issues/458

#441 adds a special handler for the `/users/*` route in CHP. Rather than try to design a nice error page while we are dealing with load problems, I went for the simpler solution of redirecting to the home Binder URL.

We should serve an informative error page, instead. Perhaps the simplest version of this is to keep it just a redirect here, but link to `binder/mypodissgone.html`.

Similarly related to this, we should consider if redirect logic like this is appropriate for the binder chart itself, as opposed to here in our specific mybinder.org deployment.

minrk:
instead of redirecting to home or serving generic nginx 404

This should be a more permanent solution to #448

Related to https://github.com/jupyterhub/binderhub/issues/457 and https://github.com/jupyterhub/binderhub/issues/458

This fixes #443 *except* for the fact that we should think about promoting this behavior to the binderhub chart itself.

### DIFF
--- a/mybinder/templates/proxy-patches/configmap.yaml
+++ b/mybinder/templates/proxy-patches/configmap.yaml
@@ -9,12 +9,50 @@ data:
       listen 80;
 
       # Return 404 for anything that comes here
-      # This preserves the same behavior as before, just without overloading the hub
+      # serve an informative error about stopped pods
+      error_page 404 /404.html;
+      location /404.html {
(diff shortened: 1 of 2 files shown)


betatim:
Nice!

PR #469, opened 2018-02-22
incident report for 2018-02-20

### LINKED ISSUE #430: Incident report for jupyterlab launch
@minrk has started an incident report doc here: https://hackmd.io/BwYwrAhgpgDBDsBaAzCALAM0WswsRBmUQCMQBOIiAE0hOpiA

cc/ @jupyterhub/binder-team

minrk:
it was a doozy! We all learned a lot. Issues created across several repos for action items.

Ultimately, this boiled down to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/461 which hadn't propagated to the mybinder.org deployment. The result was that jupyterhub itself had greatly increased load during high traffic and couldn't handle the number of users. I believe that most of the rest had to do with how well various components deal with the Hub not keeping up (including the hub itself) and how we can/should keep Hub load from getting too great. I think we have most of these in place, but the bug brought the threshold of what the Hub can actually handle well below our thresholds for preventing too much load.

Feel free to make edits to language here or in follow-up PRs.

closes #430

### DIFF
--- a/docs/source/incident-reports/2018-02-20-jupyterlab-announcement.md
+++ b/docs/source/incident-reports/2018-02-20-jupyterlab-announcement.md
@@ -1,0 +1,392 @@
+# 2018-02-20, JupyterLab Announcement swamps Binder
+
+## Summary
+
+The JupyterLab [announcement post](https://blog.jupyter.org/jupyterlab-is-ready-for-users-5a6f039b8906) drove a great deal of traffic to mybinder.org.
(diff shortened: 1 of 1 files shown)


choldgraf:
This was a great experience IMO - we handled a pretty nasty outage and have a lot of lessons-learned / next steps in the process. Good job all!

I think we can spot-check this in follow-up PRs if we're missing something or need to follow up

PR #471, opened 2018-02-22
Increase per repo limit to 300

betatim:
This follows #460. While we promised to run at the limit of 200 for a while we are now going to 300 straight away. Thinking is that we have been running at >250 live binders for the last 6hours, >300 for over 5h. CPU usage of jhub is still low.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -95,7 +95,7 @@ binderhub:
             - NET_ADMIN
       schedulerStrategy: pack
   repo2dockerImage: jupyter/repo2docker:f069a7b
-  perRepoQuota: 200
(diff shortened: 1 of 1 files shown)


choldgraf:
I'm a little bit hesitant to do this just yet, as we've had some instability in Binder today that was immediately fixed by deleting a bunch of stale jupyterhub pods. I feel like bumping up the limit would have exacerbated this problem :-/

betatim:
Agreed. I think we should wait and see what happens when we get close to ~500 running pods again. That was roughly when we started seeing the launch success rate dropping off. Let's cross that hurdle and if all is stable reconsider this.

betatim:
Closing because super old.

PR #473, opened 2018-02-22
Simplify 404 page text

### LINKED ISSUE #462: Review stopped-binder error page
We are now serving [this html](https://github.com/jupyterhub/mybinder.org-deploy/blob/055f218f7835b3f408521f9cf238612cade5ac3e/mybinder/templates/proxy-patches/configmap.yaml#L29) when a user tries to visit a stopped pod. It would be useful to get some review passes on the text there to best communicate to users what may have happened and what they can/should do next.

Cases include:

- their own binder stopped
  - due to an error
  - culled by idle culler
  - culled for max age
- they were given a link from someone copy/pasting from their browser address bar

You can test the current version of the page by visiting https://hub.mybinder.org/user/doesntexist

cc @willingc @choldgraf

willingc:
Provides clarity to the user on troubleshooting when they receive a 404 page as suggested by @minrk.

This is a first pass. Additional PRs could improve the styling.

Closes #462

### DIFF
--- a/mybinder/templates/proxy-patches/configmap.yaml
+++ b/mybinder/templates/proxy-patches/configmap.yaml
@@ -29,29 +29,37 @@ data:
   404.html: >
     <html>
     <head>
-    <title>Your Binder has stopped</title>
+    <title>Binder not found</title>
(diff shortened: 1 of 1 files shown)


choldgraf:
I love it!

PR #482, opened 2018-02-23
nginx incident from today

choldgraf:
The NGINX pods froze up and we had to delete them!

cc @yuvipanda


yuvipanda:
Thanks, @choldgraf

PR #488, opened 2018-02-23
Docs for: ingress, kube-lego, GKE, GCR & cloud product philosophy

### LINKED ISSUE #416: Document all the components of mybinder.org that we have installed and use

### LINKED ISSUE #417: Document the cloud services we use in mybinder.org
In running mybinder.org, we use many cloud services. We should document what these are, why we use them, and how we use them!

Current list is:

- [ ] Google Kubernetes Engine
- [ ] Google Container Registry
- [ ] StackDriver logging
- [ ] Google Network Load Balancer

This will also let us be explicit and clear about the cloud-specific things we are doing. None of these services are unique to Google, and we should continue to stay that way!

yuvipanda:
Lots of documentation on various components we use:

1. nginx-ingress for getting HTTP traffic into our cluster
2. kube-lego for Let's Encrypt ceritficates
3. Google Kubernetes Engine (GKE) for hosted Kubernetes
4. Google Container Registry (GCR) for hosted Container Registry

Also included is some guidance on when to use hosted vendor's cloud
products.

closes #416 closes #417

### DIFF
--- a/docs/source/command_snippets.md
+++ b/docs/source/command_snippets.md
@@ -105,10 +105,10 @@ the new minimum is higher than the current size and there is no need to increase
 the size of the cluster the autoscaler will not scale up the cluster even though
 it is below the minimum size.
 
-## Manually decrease cluster size
+## Manually scaling down the cluster size
 
-The autoscaler has issues scaling nodes *down*, and so scaling down needs to be
-manually done. The problems are caused by problem is caused by:
+The autoscaler has issues scaling nodes *down*, so scaling down needs to be
+manually done. The problems are caused by:
 
 1. The cluster autoscaler will never remove nodes that have user pods running.
 2. We can not tell the Kubernetes Scheduler to 'pack' user pods efficiently -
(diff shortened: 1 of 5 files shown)


choldgraf:
This is really really nice! Lots of valuable information in there. Within a few months there will be enough info on this site to teach a class :-D

Made several edits throughout for clarity etc. I also added a new page "terminology" where we can add k8s jargon etc.

@yuvipanda want to take a quick scan to make sure it's 👍 to you and then we can merge?

yuvipanda:
@choldgraf yay! I looked at your changes and they all seem good. Had one minor nit, so pushed a change with that. This is good to go from my perspective now.

Thank you for making the clarifying edits!

choldgraf:
cool - I found a little typo and pushed that as well...let's give this a day or so to see if anybody else has thoughts, else we can always iterate on this more in subsequent PRs

PR #495, opened 2018-02-27
tweaks to health check notebook

minrk:
- add idle users check
- make hub_api_request reusable
- fix username lookup now that kubespawner uses annotations instead of labels

### DIFF
--- a/scripts/pod-health-check.ipynb
+++ b/scripts/pod-health-check.ipynb
@@ -91,29 +91,36 @@
    "outputs": [],
    "source": [
     "import base64\n",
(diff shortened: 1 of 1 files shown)


choldgraf:
Really nice! Just a note that we have a small amount of code redundancy for deleting old pods etc (see https://github.com/jupyterhub/mybinder.org-deploy/blob/master/scripts/delete-pods.py). That said if this is ready to go on your end, I think we should merge and iterate later

betatim:
I'll merge it now and we can discuss how to share code between our various notebooks.

PR #502, opened 2018-03-06
cull idle kernels via repo2docker appendix

minrk:
uses notebook server's culling mechanisms to shutdown kernels and servers that are idle. This should solve one very specific case where the notebook server has more information than the proxy's activity tracking: a left-open notebook tab with an active websocket connection.

The notebook culler has `.cull_connected` which allows the culler to ignore the active connections when considering a notebook idle. This means that leaving a notebook open but not running anything for ~ten minutes it will still get culled, which is not happening right now due to websocket keepalive traffic.

Since this is a harsher shutdown (opening a notebook and reading can result in culling), we could give it a longer timeout than the network-activity culler we are also using.

This is WIP because only the appendix files are added, not actually enabled, which must wait for https://github.com/jupyterhub/binderhub/pull/459

### DIFF
--- a/appendix/extra_notebook_config.py
+++ b/appendix/extra_notebook_config.py
@@ -1,0 +1,3 @@
+import os
+c.NotebookApp.extra_template_paths.append('/etc/jupyter/binder_templates')
+c.NotebookApp.jinja_template_vars.update({'binder_url': os.environ.get('BINDER_URL', 'https://mybinder.org')})
--- a/appendix/extra_notebook_config.py
+++ b/appendix/extra_notebook_config.py
@@ -5,11 +5,19 @@ c.NotebookApp.jinja_template_vars.update({
 })
 
 # shutdown notebook after no activity for 10 minutes
-c.NotebookApp.shutdown_no_activity_timeout = 10 * 60
-# shutdown idle kernels after 10 minutes
-c.MappingKernelManager.cull_idle_timeout = 10 * 60
-# check kernel idleness every minute
-c.MappingKernelManager.cull_interval = 60
+if os.getenv('CULL_TIMEOUT'):
+    # shutdown the server after no activity
+    c.NotebookApp.shutdown_no_activity_timeout = int(os.getenv('CULL_TIMEOUT'))
+
+if os.getenv('CULL_KERNEL_TIMEOUT'):
+    # shutdown kernels after no activity
+    c.MappingKernelManager.cull_idle_timeout = int(os.getenv('CULL_KERNEL_TIMEOUT'))
+
+if os.getenv('CULL_INTERVAL'):
+    # check for idle kernels this often
+    c.MappingKernelManager.cull_interval = int(os.getenv('CULL_INTERVAL'))
+
 # a kernel with open connections but no activity still counts as idle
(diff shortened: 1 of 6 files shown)


willingc:
@mirnk Just merged the PR in repo2docker which pins Tornado.

minrk:
Thanks. This is almost done.

minrk:
I had to set the restartPolicy to Never via extraConfig here, so that culling internally would actually stop the pod. I've submitted the same change to be the default behavior in KubeSpawner here: https://github.com/jupyterhub/kubespawner/pull/143

appendix/extra_notebook_config.py
```diff
@@ -0,0 +1,23 @@
+import os
+c.NotebookApp.extra_template_paths.append('/etc/jupyter/binder_templates')
+c.NotebookApp.jinja_template_vars.update({
+    'binder_url': os.environ.get('BINDER_URL', 'https://mybinder.org'),
+})
+
+# shutdown notebook after no activity for 10 minutes
```
willingc: Replace 10 min with cull timeout period

appendix/templates/login.html
```diff
@@ -0,0 +1,25 @@
+{% extends "templates/login.html" %}
+{% block site %}
+
+<div id="ipython-main-app" class="container">
+  <h1>You can't access this Binder</h1>
```
willingc: Binder inaccessible.

appendix/templates/login.html
```diff
@@ -0,0 +1,25 @@
...
+  <h2>
+    You can get a new Binder for this repo by clicking <a href="{{binder_url}}">here</a>.
+  </h2>
+  <p>
+    The shareable URL for this repo is: <tt>{{binder_url}}</tt>
+  </p>
+
+  <h4>Is this a Binder that you created?</h4>
+  <p>
+    If so, your authentication cookie used has been deleted or expired.
```
willingc: Your Binder authentication was reset due to inactivity.

appendix/templates/login.html
```diff
@@ -0,0 +1,25 @@
...
+    You can get a new Binder for this repo by clicking <a href="{{binder_url}}">here</a>.
```
willingc: Launch a new Binder ...

appendix/extra_notebook_config.py
betatim: drop the 10minute bit, at least I think none of this actually sets it to ten minutes

minrk: oops, out-of-date comment. Used to be hardcoded, now comes from env

appendix/templates/login.html
minrk: We don't know that 'inactivity' is the cause, only that it has been deleted or expired somehow. It could have been manually cleared, overridden by a misconfigured other Binder, etc.

betatim:
LGTM

willingc:
Changes look great. Thanks @minrk

minrk:
This will need jupyterhub/kubespawner#145 before it can work

minrk:
It works! Deploying to staging.

PR #503, opened 2018-03-06
[wip] add pod-culler replicaset

minrk:
essentially runs our delete script every hour to enforce a max age of six hours on user pods

cull image built with chartpress, publication not automated yet

- [ ] test it out
- [ ] enable chartpress on travis to build / publish images
- [ ] maybe use [CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/)

### DIFF
--- a/chartpress.yaml
+++ b/chartpress.yaml
@@ -1,0 +1,6 @@
+charts:
+  - name: mybinder
+    imagePrefix: minrk/binder-
+    images:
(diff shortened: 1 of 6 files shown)


minrk:
After thinking about it, I now think this belongs on the jupyterhub helm chart as culler.maxAge. I'll close here when I open a PR over there.

PR #513, opened 2018-03-14
kubetools module

choldgraf:
I put together a tiny python module that I use to make it easier to work with kubernetes in python. It's definitely some degree of replication over the k8s Python API, but it basically just lets you do simple `kubectl` system calls from a python session. Right now it's only two functions:

1. `get_all` will run `kubectl get <kind> -o wide` and return the results in a DataFrame...this makes it easy if you wanna do things like "delete all pods that match condition XXX".
2. `delete` will run `kubectl delete <kind> <name>` w/ an optional `force=True` flag.

Not sure if this is too specific to be included in mybinder.org, but I've found these useful for quickly doing non-trivial things with `kubectl` so thought I'd make a PR in case it's useful for others...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ mybinder/requirements.lock
 
 docs/_build
 travis/crypt-key
+
+# Chris sandbox file name
+sandbox.ipynb
--- a/scripts/kubetools/__init__.py
+++ b/scripts/kubetools/__init__.py
@@ -1,0 +1,2 @@
+"""A collection of functions to facilitate working with k8s in Python."""
+from .kubetools import get_all, delete
--- a/scripts/kubetools/kubetools.py
+++ b/scripts/kubetools/kubetools.py
@@ -1,0 +1,53 @@
+from subprocess import check_output, check_call
+import pandas as pd
+
+AGE_IN_MINS = {'s': 1/60., 'm': 1, 'h': 60, 'd': 60*24}
+
+def _age_str_to_minutes(item):
+    kind = item[-1]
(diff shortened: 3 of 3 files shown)


yuvipanda:
Maybe a separate repo to hold personal but useful-to-others scripts like this? While it might be nbd, each merge here does trigger a deploy and I think we can be more free with merges in a separate repo. It also makes things look less 'official' and more personal. mybinder.org-maint maybe?

yuvipanda:
(am also happy for these to remain here if you / others don't think the overhead is worth it!)

choldgraf:
Hmm, I think this is a good point, maybe it'd be easier to spin off the
scripts into something like a `mybinder-scripts` repo?

choldgraf:
I dumped most of this code here: https://github.com/jupyterhub/kubetools

I'm not sure if it'd be better to move _all_ the scripts there, or if this should be just for the little python module. What are your thoughts?

willingc:
@choldgraf Can we perhaps rename the repo kubetools to something that better reflects that the tools in the repo are ones used for JupyterHub and Binder developments? I don't want to accidentally send the message that we are offering and "supporting" kubernetes tools for the general k8s community.

choldgraf:
good point! will change the repo name
--

willingc:
Perhaps the original suggestion mybinder-scripts or mybinder-tools

choldgraf:
ok, renamed the repo and it lives here now:

https://github.com/jupyterhub/mybinder-tools

gonna close this one

PR #517, opened 2018-03-16
fix proxy error target

minrk:
error pages are served on /hub/error/:code not /:code


betatim:
Is there a reason not to merge this? Double checking as most PRs in this repo are self merged.

PR #520, opened 2018-03-21
Increase memory limit for CHP nginx container

betatim:
This container seems to regularly bump into the memory limit and gets
killed. Increasing the limit is more of a bandaid to help with the
symptom, than a long term fix.

From grafana it seems we get occasional spikes which are above the limit, not sustained long term increase/leaks.

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -44,10 +44,10 @@ binderhub:
       nginx:
         resources:
           requests:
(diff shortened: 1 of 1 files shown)


minrk:
Possibly relevant report of nginx memory growth: https://github.com/kubernetes/contrib/issues/2215

PR #523, opened 2018-03-22
run tests with travis_retry

minrk:
The first run of the tests may fail while things are warming up on a fresh deploy. travis_retry gives the tests a couple of attempts to succeed. This isn't the best, but we've had tests fail when everything is fine due to running too early too often, and I think this will save everyone time.

In particular, one source of recent failures has the proxy-helper which turns what would be Hub 403 responses into Binder-specific 404 pages. It takes a finite time to update the proxy with these redirects. This manifests as a test failure where the expected status is 404 but actual status is 403.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,13 +63,13 @@ before_deploy:
   python ./deploy.py staging staging
 - |
   # Stage 3, Step 2: Verify staging works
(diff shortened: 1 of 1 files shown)


betatim:
`travis_retry` will try three times before giving up. Can't see a real downside as our deploy takes a few minutes already so having to wait for a few extra network requests seems like small potatoes.

PR #525, opened 2018-03-26
incident report for brief outage this morning

minrk:
"no space left on device" errors caused launches to fail for 20 minutes

No stress as the cluster recovered and I was investigating while

Result highlights:

- cordoned two nodes that showed many "no space left" messages. Other two nodes have zero of these.
- added stackdriver monitoring alerts for failing launches

### DIFF
--- a/docs/source/incident-reports/2018-03-26-no-space-left.md
+++ b/docs/source/incident-reports/2018-03-26-no-space-left.md
@@ -1,0 +1,60 @@
+# 2018-03-26, "no space left on device"
+
+## Summary
+
+A node became unhealthy, correlated with a flood of "no space left on device"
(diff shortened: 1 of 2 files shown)


choldgraf:
nice - thanks for the report! two quick thoughts:

1. I'm back in the states, so gonna try and spend some time this week getting SMTP working with grafana
2. Is there any reason *not* to cordon/drain nodes that are ~1 week old or so? (minus the downtime we'd hit for production pods)?

minrk:
> Is there any reason not to cordon/drain nodes that are ~1 week old or so?

I don't think so. We seem to be having issues a bit more quickly than that. More like 3-5 days.

PR #526, opened 2018-03-27
run cull every 11 minutes

minrk:
instead of every 10

reduces coincidence of cull + proxy.check_routes which is every 5 minutes by default, so cull and check_routes were usually happening at the same time, causing brief periods of poor performance every ten minutes.

related to https://github.com/jupyterhub/jupyterhub/pull/1753 and https://github.com/jupyterhub/jupyterhub/issues/1754

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -53,6 +53,9 @@ binderhub:
 
   jupyterhub:
     cull:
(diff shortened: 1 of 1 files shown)


choldgraf:
ooo clever catch

PR #531, opened 2018-04-01
adding recycling guide

choldgraf:
This does two things:

1. Adds a guide to recycling nodes. **Please double check to make sure this is correct.**
2. Adds a short section on what to expect if specific pods are deleted. **Please help me fill in the list below and I'll make the changes to the PR!**

```
Below is a list of each production pod, and the expected outcome that comes with
deleting each one.

* `hub-` All active user sessions will be removed.
* `binder-` The `mybinder.org` website will temporarily go down. Active user sessions will not be affected.
* `proxy-` Attempts to redirect a user to their pre-built pod will fail.
* `proxy-patches-` XXX
* `redirector-` XXX
* `jupyter-` - Deleting a user pod will shut down their session. The user will
  encounter errors when they attempt to submit code to the kernel.
```

### DIFF
--- a/docs/source/command_snippets.md
+++ b/docs/source/command_snippets.md
@@ -56,12 +56,12 @@ kubectl --namespace=prod delete pod <POD-NAME> --grace-period=0 --force
 Below is a list of each production pod, and the expected outcome that comes with
 deleting each one.
 
-* `hub-` XXX
-* `binder-` XXX
-* `proxy-` XXX
-* `proxy-patches-` XXX
-* `redirector-` XXX
-* `jupyter-` - Deleting a user pod will shut down their session. The user will
+* `hub-` active user sessions will not be affected. New and pending launches will fail until the new Hub comes back.
+* `binder-` the `mybinder.org` website will temporarily go down. Active user sessions will not be affected.
+* `proxy-` all current users will lose connections (kernel connection lost) until the proxy returns and the Hub restores the routes. Server state is unaffected. Most browser sessions should recover by restoring connections. All pending launches will fail due to lost connections.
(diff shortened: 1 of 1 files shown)


choldgraf:
I made an attempt at listing the ramifications of deleting each pod - could somebody double check that the one's I've listed are correct, and fill in the `XXX` where I wasn't sure what to put.

Once this is merged, I'm going to start sending a recurring email to the team to remind folks to recycle old nodes :-)

choldgraf:
cc @yuvipanda @minrk and @betatim as I think they have the best intuition for what the effects will be when particular pods are deleted

minrk:
Some corrections:

* `hub-` Active user sessions will not be affected. New and pending launches will fail until the new Hub comes back.
* `proxy-` All current users will lose connections (kernel connection lost) until the proxy returns and the Hub restores the routes. Server state is unaffected. Most browser sessions should recover by restoring connections. All pending launches will fail due to lost connections.
* `proxy-patches-` brief, minor degradation of error messages when browsers attempt to connect to a not-running server. This results in increased load on the Hub, handling requests from browsers whose idle servers have been culled.
* `redirector-` redirect sites (beta.mybinder.org) will 404 instead of sending to mybinder.org.

betatim:
Instead of an email (that will quickly get added to filters...) can we setup a travis cron job to cordon off the oldest node and drain it? It seemed like the auto scaler was doing alright in collecting a node once it was empty.

choldgraf:
ok text updated, I think this PR is ready to go from my end, as long as the content looks correct

@betatim is that a volunteer to figure out how to set up such a cron job? :-)

PR #535, opened 2018-04-03
implement CPU limits

minrk:
limit to 2 CPUs on prod, 1 on staging

still no CPU guarantees. Oversubscribing CPUs should usually be fine.

There are several Binders currently using lots of CPU. I'm going to shut them down once this lands.


willingc:
Nice. Thanks @minrk

PR #544, opened 2018-04-06
Revert "Add a dummy cron job"

minrk:
Reverts jupyterhub/mybinder.org-deploy#534

@betatim I think adding jobs means all jobs are opt-in so having just this dummy cron job *replaced* the default deployment jobs with the dummy cron. As a result, it prevented deploying #543. To add a cron job, we seem to need to build our whole deploy flow with the jobs structure. That should also make it clearer that the cron job only runs the cron stage, not a full deploy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,6 @@ install:
 script:
  - true
 
-jobs:
-  include:
-    - if: type = cron
-      stage: cron
-      script: echo "this is cron"
-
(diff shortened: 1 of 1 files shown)


minrk:
I think it'll look something like:

```
jobs:
  - include:
    stage: cron
    if: cron
  - include:
    stage: deploy
    if: NOT cron
```

which will clearly state that we run the cron job only on cron and run the deploy only on not cron

PR #558, opened 2018-04-17
Lift ban on nbinteract

betatim:
Removing the ban to see what will happen.


betatim:
No worries. If you need help covering lots of different browsers to test changes ping us.

Issue #559, opened 2018-04-17
Cannot bump JupyterHub chart due to networkpolicy error


choldgraf:
This issue is trying to distill some trouble that @betatim, @willingc, and I ran into when trying to upgrade JupyterHub on the Binder deployment.

When bumping the JupyterHub chart, we ran into this error on deploy:

https://travis-ci.org/jupyterhub/mybinder.org-deploy/builds/367366526#L600

```
Error: UPGRADE FAILED: Deployment.apps "proxy" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"hub.jupyter.org/network-access-singleuser":"true", "name":"proxy", "release":"staging", "component":"proxy", "heritage":"Tiller", "hub.jupyter.org/network-access-hub":"true"}: `selector` does not match template `labels`
```

It seems that there's a bug in the chart somewhere, but we weren't able to spot it. We should figure this out.

Gitter chat about this begins ~ here: https://gitter.im/jupyterhub/binder?at=5ad60b78102fac7b7b202ddb

choldgraf:
cc @yuvipanda @minrk @manics who have done some networkpolicy stuff lately

betatim:
I think we have this problem: https://github.com/kubernetes/helm/issues/2494#issuecomment-311691146 namely we don't have a `.spec.selector.matchLabels` in `jupyterhub/templates/proxy/deployment.yaml`

betatim:
I think the reason this fails on "upgrade" is that helm auto generates `.spec.selector.matchLabels` for us and is then confused when the now stored on the cluster bit doesn't match any more. Maybe the long term solution is to also specify this in the template? (after using what Min suggested in [gitter](https://gitter.im/jupyterhub/binder?at=5ad61cce5f188ccc1574db76) to get ourselves unstuck)

minrk:
On a test deployment, I was able to do the upgrade if I deleted the proxy proxy deployment prior to upgrade:

    kubectl delete deployment proxy

From the Issue @betatim linked, we should note that changing labels will not appear to be allowed going forward, so we appear to have to delete and redeploy every time we want to change labels.

betatim:
[...]

PR #560, opened 2018-04-18
bump binderhub again

minrk:
This has a temporary fix for #559 by deleting the proxy deployment prior to `helm upgrade`. This will result in a bit of downtime, but appears to be the only workaround for https://github.com/kubernetes/helm/issues/2494

In general, https://github.com/kubernetes/helm/issues/2494 seems to suggest that we need to delete objects (or whole helm deployments) any time we change labels.


betatim:
LGTM.

PR #565, opened 2018-04-18
moving grafana deploy msg to end of travis

choldgraf:
Right now, a message is posted to grafana just *before* we try deploying an upgrade. This means that we get a lot of double-grafana messages because travis needs to be restarted. This PR moves the grafana message to the _end_ of a deploy so that it serves as a "confirmation" message, and should only happen once

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,16 +49,6 @@ before_deploy:
   helm repo update
   (cd mybinder && helm dep up)
 - |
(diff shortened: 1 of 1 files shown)


minrk:
Maybe it should be after we deploy to prod, but before we test it? If the tests fail, it might still be nice to know that a deploy happened.

choldgraf:
good point, updated

willingc:
Nice! Thanks @choldgraf.

PR #574, opened 2018-04-26
bhub bump for image cleaner in BinderHub

choldgraf:


choldgraf:
@minrk does this look correct re: threshold?

minrk:
I think so. Worth a shot. I'm not 100% sure .6 is valid as 0.6, but I imagine so.

choldgraf:
went ahead and made it 0.6...shall we merge?

PR #579, opened 2018-04-27
Revert "upgrade helm to 2.8.2"

choldgraf:
Reverts jupyterhub/mybinder.org-deploy#578

Kubernetes was unhappy with this and it failed on the upgrade. Here's the relevant travis line:

https://travis-ci.org/jupyterhub/mybinder.org-deploy/jobs/372149241#L582

```
Activated service account credentials for: [travis-deployer@binder-staging.iam.gserviceaccount.com]
Fetching cluster endpoint and auth data.
kubeconfig entry generated for staging.
Error: error installing: deployments.extensions is forbidden: User "travis-deployer@binder-staging.iam.gserviceaccount.com" cannot create deployments.extensions in the namespace "kube-system": Required "container.deployments.create" permission.
```

cc @minrk

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,7 @@ before_deploy:
   chmod +x ${HOME}/bin/kubectl
 - |
   # Stage 1: Install helm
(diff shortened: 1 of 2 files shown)


minrk:
Thanks for the revert! This means we need to change the travis permissions if we want to run a current version of helm.

PR #580, opened 2018-04-30
Retry "upgrade helm to 2.8.2"

minrk:
Reissue #578. It failed (see #579)

I added the `container.deployments.create` permission to the travis-deployers role on gcloud. Let's see if this is enough to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,8 @@ before_deploy:
   chmod +x ${HOME}/bin/kubectl
 - |
   # Stage 1: Install helm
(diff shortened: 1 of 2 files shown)


minrk:
Additionally needed permissions before it would succeed:

- container.deployments.update
- container.services.get

And ran into https://github.com/kubernetes/helm/issues/3785 which seems to be fixed by clearing some bad state in old deployments that are incorrectly marked as still deployed.

If deployments fail with something like:

    Error: UPGRADE FAILED: no ConfigMap with the name "something" found

let's revert the helm version to 2.7.2

PR #591, opened 2018-05-08
bump binderhub

minrk:
updates image-cleaner, fixes docker socket mounting in races between builds and dind

c0a1dae...7a22080

https://github.com/jupyterhub/binderhub/compare/c0a1dae...7a22080

### DIFF
--- a/mybinder/requirements.yaml
+++ b/mybinder/requirements.yaml
@@ -13,5 +13,5 @@ dependencies:
    version: 0.1.12
    repository: https://kubernetes-charts.storage.googleapis.com
  - name: binderhub
-   version: 0.1.0-c0a1dae
+   version: 0.1.0-7a22080
    repository: https://jupyterhub.github.io/helm-chart


minrk:
This eventually succeeded, but had to be helped along:

1. the dind updates produced the race where `/var/run/dind/docker.sock` was created as a directory. This is resolved by SSHing to the node and deleting the directory and restarting the pods. This *should* have been fixed by https://github.com/jupyterhub/binderhub/pull/539, and it may have been, in fact, but the race could still happen during the deploy of the fix. We may never see this one again.
2. intermittent network issues with gcloud (communication failures in helm upgrade) required restarting multiple times

betatim:
We had a big queue of [build pods waiting to start](https://grafana.mybinder.org/d/fZWsQmnmz/pod-activity?refresh=1m&orgId=1&panelId=4&fullscreen&from=1526663848548&to=1526674648548). Picked a random build pod that was waiting, described it and saw the error was that `/var/run/dind/docker/sock` was created as a directory. Fixed by `sudo rmdir /var/run/dind/docker.sock/` on node -cvh2 and then `kubectl delete pod prod-dind-rp7z8`.

So this race still exists.

choldgraf:
@minrk @betatim do you think we should create a new issue for this specific race condition?

minrk:
The issue is jupyterhub/binderhub#349

PR #592, opened 2018-05-08
[staging] Try again updating grafana/prometheus

minrk:
Deploying only to staging


minrk:
Success this time, with a bit of fiddling needed to delete old configmaps and restore some grafana dashboards. Will deploy to prod shortly.

PR #597, opened 2018-05-11
script to list new commits

choldgraf:
I made a little helper script for myself to print the diff between the current "live" repo2docker and binderhub images, and the master branch of their respective repositories. This just adds that script in there.

Simply run `python scripts/list_new_commits.py`

and you'll see output like

```
---------------------

BinderHub: https://github.com/jupyterhub/binderhub/compare/a5d1851...d02a031
repo2docker: https://github.com/jupyter/repo2docker/compare/7a50772...7a50772

---------------------
```

### DIFF
--- a/scripts/list_new_commits.py
+++ b/scripts/list_new_commits.py
@@ -1,0 +1,41 @@
+from yaml import load
+import requests
+
+print('Fetching the SHA for live BinderHub and repo2docker...')
+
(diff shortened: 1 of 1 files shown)


betatim:
Now we just need a script that modifies mybinder.org-deploy when it isn't up2date. `python scripts/live_at_head.py` ;)

PR #606, opened 2018-05-17
bhub 470642d...ee79f00

choldgraf:
This is a <BinderHub / repo2docker> version bump. See the link

https://github.com/jupyterhub/binderhub/compare/470642d...ee79f00


willingc:
@choldgraf @betatim Did you mean to merge this without the revert for logging?

betatim:
I think this is fine. The revert changed us back to an older hash, this PR moves us to one newer than the one that the revert reverted. Something like A -(original PR)-> B -(revert)-> A -(this PR)-> C

PR #612, opened 2018-05-21
steps to use the python commit script

choldgraf:
This updates the deploy steps to use the `list_new_commits.py` script. I hope this will reduce the number of steps needed to make a deploy, and also standardize the PR names that we've got. Lemme know what folks think!

### DIFF
--- a/docs/source/deployment/how.md
+++ b/docs/source/deployment/how.md
@@ -25,8 +25,8 @@ which we step through below.
 1. Merge changes to BinderHub.
 2. Open the [branches page for the BinderHub travis account](https://travis-ci.org/jupyterhub/binderhub/branches).
 3. Wait for the build for your PR merge to pass (it will say "#NNN passed").
-   If it does, then:
-
+   If it does, then continue to step 4. If it doesn't, take a look at the error message, and debug as needed
+   until they pass.
 4. Run the `list_new_commits.py` script in the `scripts/` of the
    `mybinder.org-deploy` repository. It will output something like the following:
 
@@ -34,6 +34,7 @@ which we step through below.
 
        BinderHub: https://github.com/jupyterhub/binderhub/compare/<OLD-HASH>...<NEW-HASH>
        repo2docker: https://github.com/jupyter/repo2docker/compare/<OLD-HASH>...<NEW-HASH>
+       JupyterHub: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/<OLD-HASH>...<NEW-HASH>
 
        ---------------------
 
@@ -75,7 +76,9 @@ The following lines describe how to point mybinder.org to the new repo2docker im
 1. Merge changes to repo2docker.
 2. Open the [branches page for repo2docker](https://travis-ci.org/jupyter/repo2docker/branches).
    And click on the number for the latest build on "Master".
-3. Wait for the build to pass (it will say "#NNN passed"). If it does, then:
+3. Wait for the build for your PR merge to pass (it will say "#NNN passed").
+   If it does, then continue to step 4. If it doesn't, take a look at the error message, and debug as needed
+   until they pass.
 4. Run the `list_new_commits.py` script in the `scripts/` of the
    `mybinder.org-deploy` repository. It will output something like the following:
 
@@ -83,7 +86,8 @@ The following lines describe how to point mybinder.org to the new repo2docker im
 
        BinderHub: https://github.com/jupyterhub/binderhub/compare/<OLD-HASH>...<NEW-HASH>
        repo2docker: https://github.com/jupyter/repo2docker/compare/<OLD-HASH>...<NEW-HASH>
-
+       JupyterHub: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/<OLD-HASH>...<NEW-HASH>
+       
        ---------------------
 
    Since you are updating repo2docker, copy the text in `<NEW-HASH>` for the
--- a/docs/source/deployment/how.md
+++ b/docs/source/deployment/how.md
@@ -24,35 +24,41 @@ which we step through below.
 
 1. Merge changes to BinderHub.
 2. Open the [branches page for the BinderHub travis account](https://travis-ci.org/jupyterhub/binderhub/branches).
-3. Click on the number for the latest build on "Master". It will say "#NNN passed".
-4. You'll see a list of builds that have been run on this branch. Click on the `TEST=helm` job.
-5. If the build succeeds (is green), a new helm chart for BinderHub will automatically
-   be published and listed on GitHub. Go to the [BinderHub Helm Chart](https://jupyterhub.github.io/helm-chart/#development-releases-binderhub)
-   page and grab the hash for the latest published version (at the top).
+3. Wait for the build for your PR merge to pass (it will say "#NNN passed").
+   If it does, then:
 
-       binderhub-<version-number>-<hash-name>
+4. Run the `list_new_commits.py` script in the `scripts/` of the
+   `mybinder.org-deploy` repository. It will output something like the following:
 
-   You want to copy `<hash-name>`. It will look something like
-   example, it is `f87ac35`.
+       ---------------------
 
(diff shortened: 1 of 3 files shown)


docs/source/deployment/how.md
```diff
@@ -24,35 +24,41 @@ which we step through below.
 
 1. Merge changes to BinderHub.
 2. Open the [branches page for the BinderHub travis account](https://travis-ci.org/jupyterhub/binderhub/branches).
-3. Click on the number for the latest build on "Master". It will say "#NNN passed".
-4. You'll see a list of builds that have been run on this branch. Click on the `TEST=helm` job.
-5. If the build succeeds (is green), a new helm chart for BinderHub will automatically
-   be published and listed on GitHub. Go to the [BinderHub Helm Chart](https://jupyterhub.github.io/helm-chart/#development-releases-binderhub)
-   page and grab the hash for the latest published version (at the top).
+3. Wait for the build for your PR merge to pass (it will say "#NNN passed").
+   If it does, then:
```
willingc: If it does, continue to Step 4? If it doesn't ???

docs/source/deployment/how.md
```diff
@@ -24,35 +24,41 @@ which we step through below.
...
-       binderhub-<version-number>-<hash-name>
+4. Run the `list_new_commits.py` script in the `scripts/` of the
+   `mybinder.org-deploy` repository. It will output something like the following:
 
-   You want to copy `<hash-name>`. It will look something like
-   example, it is `f87ac35`.
+       ---------------------
 
-6. In your fork of the [mybinder.org-deploy](https://github.com/jupyterhub/mybinder.org-deploy)
+       BinderHub: https://github.com/jupyterhub/binderhub/compare/<OLD-HASH>...<NEW-HASH>
```
willingc: Should we add JupyterHub as well? If it hasn't changed, no big deal. If it has, the information gets captured aas well.

choldgraf: yeah we definitely could to the `.py` script. Do you know where in BinderHub is the value for which jupyterhub we depend on?

willingc: https://github.com/jupyterhub/binderhub/blob/master/helm-chart/binderhub/requirements.yaml Perhaps just print JupyterHub as an informational message. It's cool to leave it off too.

docs/source/deployment/how.md
jzf2101: Meta question here- is this doc addressed to people who have merge rights already? Because the if it doesn't depends on the audience AFAIK

jzf2101: Meta question- who is the audience here? If the person is a maintainer vs contributor the answer will be difference if the build doesn't pass

docs/source/deployment/how.md
```diff
@@ -24,35 +24,41 @@ which we step through below.
...
+       repo2docker: https://github.com/jupyter/repo2docker/compare/<OLD-HASH>...<NEW-HASH>
+
+       ---------------------
+
+   Since you are updating BinderHub, copy the text in `<NEW-HASH>` for the
+   line that refers to BinderHub. This is the name of the new BinderHub image.
+   We'll now update the config to refer to this image.
```
jzf2101: Does we refer to the maintainers here?

docs/source/deployment/how.md
```diff
@@ -24,35 +24,41 @@ which we step through below.
...
+
+5. In your fork of the [mybinder.org-deploy](https://github.com/jupyterhub/mybinder.org-deploy)
    repository, open `mybinder/requirements.yaml`.
-7. Toward the end of the file, you will see lines similar to:
+6. Toward the end of the file, you will see lines similar to:
 
       - name: binderhub
         version: 0.1.0-9692255
```
jzf2101: Shouldn't this be `<OLD-HASH>` then?

docs/source/deployment/how.md
```diff
@@ -24,35 +24,41 @@ which we step through below.
...
-   Replace the existing hash that comes just after the `-` under 'version' with new hash
-   from step 3. In this example, replace `9692255`  with the hash `f87ac35`that you've
-   copied in step 3. The edited lines will be:
+   Where `9692255` is the same value printed in `<OLD-HASH>` above.
+
+   Replace this hash with the text in `<NEW-HASH>` above. For example, in the
+   above case, we'll replace it with the new hash.
 
       - name: binderhub
         version: 0.1.0-fbf6e5a
```
jzf2101: Again this would likely be `<NEW-HASH>` by the same logic?

docs/source/deployment/how.md
```diff
@@ -68,27 +74,35 @@ The following lines describe how to point mybinder.org to the new repo2docker im
...
-
-       Pushed new repo2docker image: <YOUR-IMAGE-NAME>
-
-   Copy the text in `<YOUR-IMAGE-NAME>`. **Note**: You may need to unfold the
-   code in the `Deploying application` line in order to see this text.
-6. In your fork of the mybinder.org-deploy repository, open
-   `mybinder/values.yaml`.
-7. Somewhere in the file you will see `repo2dockerImage`, it will look like
+   And click on the number for the latest build on "Master".
+3. Wait for the build to pass (it will say "#NNN passed"). If it does, then:
```
jzf2101: By your same logic @willingc we should have something if this fails?

docs/source/deployment/how.md
```diff
@@ -68,27 +74,35 @@ The following lines describe how to point mybinder.org to the new repo2docker im
...
+   Where `65d5411` is the same value in `<OLD-HASH>` above.
+
+7. Replace the *old* hash that is there with what you copied in step 4.
    For example, the edited file will look similar to:
 
-       repo2dockerImage: jupyter/repo2docker:<YOUR-NEW-HASH>
+       repo2dockerImage: jupyter/repo2docker:<NEW-HASH>
 
 8. Merge this change to `mybinder/values.yaml` into the mybinder.org-deploy
    repository following the steps in the [Deploying a change](#deploying-a-change) section
```
jzf2101: More broadly perhaps we can consolidate a lot of these steps either here or in "Deploying a Change" since the steps are rather similar.

docs/source/deployment/how.md
willingc: These docs render to the mybinder Site Reliability Guide. We probably can have a blanket statement that this document is for maintainers of mybinder.org

docs/source/deployment/how.md
willingc: +1

choldgraf:
Hey all - picking up my old PRs now that I'm married and don't have to plan 2 weddings anymore :-)

I updated this PR to incorporate the JupyterHub commits as well, so we can compare all three. I also incorporated most of the feedback that @willingc and @jzf2101 gave (thanks!).

Can we leave the larger organizational comments (e.g. refactoring the whole page so there's less repeated content) to another PR? :-)

betatim:
Can we keep the (repo2docker) instructions so that they can be used without installing a script?

I've been asking people who contribute to r2d if they also want to deploy the change and point them at the SRE guide for instructions. So far people have been happy with them and the success rate of people agreeing to do it and doing it is high. I don't know if this is because you can do all the steps from a browser but it feels like having to fork, clone, run a script, go to GitHub has more friction :-/ Or it would lead to people being more invested because they have made some effort setting stuff up :-/

Is there an easy compromise?

choldgraf:
I don't think this script is really "required", you can still look up the commit from the history, it's just meant to be a helper to make it easier for people who want to make comparison URLs. If people don't think this would make it easier, then I agree we shouldn't have the script up there...

betatim:
Ah ok. From a quick look at the diff this morning it looked like we were removing the steps for looking it up by hand. If we keep the "by hand" instructions my concerns are null and void.

The script definitely makes my life easier (so I want this script) was more thinking about new contributors and what a reasonable number of steps is to ask them do.

choldgraf:
@betatim I mean, we could always build a Binder that lets people run the script :-)  all that information is in the public anyway...

choldgraf:
mergable from my end

PR #624, opened 2018-06-02
Ban SamLau95/nbinteract-image

betatim:
This bans SamLau95/nbinteract-image as they are using up their 100 slots (and more) if we let them.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -6,6 +6,7 @@ binderhub:
       c.GitHubRepoProvider.banned_specs = [
         # e.g. '^org/repo.*',
         '^ines/spacy-binder.*',
+        '^SamLau95/nbinteract-image.*',
       ]
   service:
     type: ClusterIP


betatim:
This is the referrer causing the build requests: https://www.nbinteract.com/examples/examples_empirical_distributions.html

https://grafana.mybinder.org/render/d-solo/fZWsQmnmz/pod-activity?refresh=1m&orgId=1&panelId=1&from=1527811368023&to=1527897768023&width=1000&height=500&tz=UTC-07%3A00

It isn't causing any trouble yet but rather ban it now instead of have some problems over the weekend when everyone is away.

cc @samlau95

betatim:
For reference the output of `kubectl logs -f prod-nginx-ingress-controller-7b4fdbdcc8-6rf58 nginx-ingress-controller | grep "GET /build/gh/SamLau95/nbinteract-image"`

```
10.128.0.2 - [10.128.0.2] - - [01/Jun/2018:23:52:27 +0000] "GET /build/gh/SamLau95/nbinteract-image/master HTTP/2.0" 200 327 "https://www.nbinteract.com/examples/examples_empirical_distributions.html" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/59.0.3071.104 Safari/537.36" 48 0.878 [prod-binder-8585] 10.12.0.184:8585 344 0.878 200
```

betatim:
The weird thing is that there were no changes to the nbinteract code. So maybe a broken browser?

choldgraf:
well every now and then if a user of nbinteract leaves open their browser and the eventloop stays open, it'll just keep re-sending signals to binder to build, and this continues until the person closes their window :-/

betatim:
The behaviour you describe sounds like a bug in nbinteract. Not sure I understand why nbinteract should ever leave the event source open unless there is a bug in nbinteract or the browser wrt closing the event source.

SamLau95:
Heya @betatim @choldgraf , I've attempted to replicate the thebelab implementation of EventSource handling in https://github.com/SamLau95/nbinteract/pull/82. Can we unban SamLau95/nbinteract-image to see whether we get lower load?

choldgraf:
@samlau95 sounds good to me...I'll do this now

PR #631, opened 2018-06-07
[WIP] appendix: trust proxy headers and load config from env

minrk:
since the appendix goes in the image, any change to notebook config must be paired with a bump to our image prefix to force rebuilds. This adds support for notebook config in `EXTRA_NBCONFIG_` environment variables, so that we can change config for new launches of existing images by setting environment variables.


minrk:
Marking as WIP, as the trust_xheaders is more general, and is implemented in https://github.com/jupyterhub/binderhub/pull/585

I still think we should do the EXTRA_NBCONFIG_ so that we can add singleuser config without having to rebuild with the updated appendix, but it's a separate issue now.

minrk:
Closing in favor of #792 which gets all the benefits here with less shenanigans.

PR #636, opened 2018-06-08
Lower the guaranteed memory and upper limit on CPU

betatim:
From looking at typical usage pattern most pods use O(100MB) of memory
and hardly any CPU. This is an attempt to allow us to pack our nodes
more tightly.

To look at usage of pods run `kubectl top pods | grep jupyter-`


choldgraf:
I think this is a good idea!

PR #637, opened 2018-06-08
Update secret sharing instructions

betatim:
I struggled some to decrypt the key as my ssh key requires a passphrase (that for some reason isn't taken from `ssh-agent` or the keychain) by ssh-vault. I've updated the instructions to something that worked for me.

What do people think about the intermediate file? Is the warning good enough?

Also, I now have access to the secrets.


choldgraf:
@betatim made some small wording edits, I'm +1 if you're +1

PR #647, opened 2018-06-13
Redirect status.mybinder.org to the docs

choldgraf:
This adds a redirect rule for status.mybinder.org to go to https://mybinder.readthedocs.io/en/latest/status.html

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -138,3 +138,7 @@ redirector:
       host:
         from: playground.mybinder.org
         to: play.nteract.io
+    - type: host
+      host:
+        from: status.mybinder.org
+        to: mybinder.readthedocs.io/en/latest/status.html


betatim:
LGTM.

One weird thing, on the status page itself I see a graph up to 23:57h, but right now it is 7:33h local time. A shift of an integer number of hours I would get but this shift I don't quite understand.

screen shot 2018-06-14 at 07 32 41

minrk:
@betatim I think you're right. The date range is probably absolute in the page, rather than the 'last 12h'. That's an issue for the docs repo, though.

minrk:
https://github.com/jupyterhub/binder/pull/104 should fix the chart time windows

choldgraf:
does this redirect work for folks?

betatim:
No :( Is there a DNS record for https://status.mybinder.org  ?

choldgraf:
maybe this is cruft from my trying to fix this yesterday...a dig says there's a CNAME for status.mybinder.org

choldgraf:
I removed the CNAME from the DNS, let's see if it works once that propagates

PR #648, opened 2018-06-14
whitelist rsync, xroot ports

### LINKED ISSUE #614: open outbound rsync on port tcp/873
@chennesy and I have a use case that uses rsync to transfer data from a remote repository.

It would be helpful to allow outbound connections via the rsync protocol port (tcp/873) so that commands like this will work for rsync’ing data from a repository which requires using the rsync protocol (and that do not allow ssh-based transport) :
```
%%bash
htid2rsync --f vol_ids_5.txt | rsync -azv --files-from=- data.sharc.hathitrust.org::features/ local-folder/
```

You can see a real-life example accessing the HathiTrust Digital Library which provides access to millions of digitized books for use by researchers: https://github.com/aculich/htrc

Currently this example hangs until rsync has a timeout since it cannot reach hathitrust.org via tcp/873.

minrk:
this is not run on CI, firewall-rules must be run manually after merge to unblock the ports

closes #614

### DIFF
--- a/scripts/firewall-rules
+++ b/scripts/firewall-rules
@@ -181,16 +181,21 @@ def main(cluster, replace=False):
         replace=replace,
     )
 
-    # whitelist http(s) to the world, as the only access we have
+    # whitelist connections to the world
+    # this is currently:
+    # HTTP[S]: 80,443
+    # git: 9418
+    # rsync: 873
+    # xroot: 1094
     add_rule(
         prefix + 'allow-http',
         {
             'priority': 1000,
-            'description': f"Allow HTTP(S) access to the outside world for cluster {cluster}",
+            'description': f"Allow approved traffic to the outside world for cluster {cluster}",
             'direction': 'egress',
             'destination-ranges': WORLD_CIDR,
             'target-tags': ','.join(node_tags),
-            'allow': 'tcp:80,tcp:443,tcp:9418',
+            'allow': 'tcp:80,tcp:443,tcp:9418,tcp:873,tcp:1094',
         },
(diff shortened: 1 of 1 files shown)


lukasheinrich:
let me know if you need someone to test the xrootd port. Or I can send some instructions.

betatim:
For rsync and xrootd should we use a whitelist of hosts as well? Not sure I could propose a starting list so this is more a thought. Otherwise LGTM.

minrk:
> For rsync and xrootd should we use a whitelist of hosts as well?

I'd really like to, but we can't do this easily. Our firewall blocking occurs at the CIDR level (ip address ranges), so we would need to resolve hosts with DNS and constantly update the associated ip addresses in the filter.

betatim:
Should we merge this? I don't remember why it is sitting here, so 👍 on merging.

lukasheinrich:
is there a way to test this? an example xrootd path would be (CMS open data)

```
root://eospublic.cern.ch//eos/opendata/cms/Run2011A/ElectronHad/AOD/12Oct2013-v1/20001/001F9231-F141-E311-8F76-003048F00942.root
```

betatim:
Merged, once the deploy is done you should be able to test it on mybinder.org @lukasheinrich

minrk:
BTW, firewall-rules is *not* run on CI, but I just ran it so this should be in effect now.

aculich:
Thanks @minrk I've confirmed that this works for `rsync` with the original repository in question that motivated this change: https://github.com/aculich/htrc

image

PR #650, opened 2018-07-04
Updating BinderHub to bring in Kubespawner updates

betatim:

https://github.com/jupyterhub/binderhub/compare/856e3e6...dc2c517


betatim:
This will also bring us the nbviewer preview!

PR #654, opened 2018-07-05
[WIP] Load testing

betatim:
Add simple script to generate fake traffic to test a binderhub

### DIFF
--- a/scripts/load-test.py
+++ b/scripts/load-test.py
@@ -1,0 +1,99 @@
+import argparse
+import json
+import time
+import random
+import pickle
+
+from collections import namedtuple
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+import requests
+
+
+Launch = namedtuple('Launch', 'idx start end status')
+
+
+def build_binder(repo,
+                  ref='master',
+                  binder_url='https://mybinder.org'):
+    """Launch a binder"""
+    url = binder_url + '/build/gh/{repo}/{ref}'.format(repo=repo, ref=ref)
+    r = requests.get(url, stream=True)
+    r.raise_for_status()
+    for line in r.iter_lines():
+        line = line.decode('utf8', 'replace')
+        if line.startswith('data:'):
+            yield json.loads(line.split(':', 1)[1])
+
+
+def launch_binder(n, repo, ref='master', binder_url='https://mybinder.org'):
+    delay = random.random() * 10
+    time.sleep(delay)
+
+    launched_at = time.time()
+
+    for evt in build_binder(repo, ref=ref, binder_url=binder_url):
+        if 'message' in evt:
+            print(n, "[{phase}] {message}".format(
+                phase=evt.get('phase', ''),
+                message=evt['message'].rstrip(),
+            ))
+        if evt.get('phase') == 'ready':
+            url = "{url}?token={token}".format(**evt)
+            ready_at = time.time()
+            print(n, "ready at %s" % url)
+            return Launch(n, launched_at, ready_at, 'success')
+
+    else:
+        failed_at = time.time()
+        print(n, "never launched.")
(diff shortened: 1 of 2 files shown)


scripts/load-test.py
```diff
@@ -0,0 +1,150 @@
...
+    r.raise_for_status()
+    for line in r.iter_lines():
+        line = line.decode('utf8', 'replace')
+        if line.startswith('data:'):
+            yield json.loads(line.split(':', 1)[1])
+
+
+def launch_binder(n, repo, ref='master', filepath=None,
+                  binder_url='https://mybinder.org'):
+    delay = random.random() * 10
```
minrk: Probably good to have the interval be an input (e.g. users per minute, etc.)

scripts/load-test.py
```diff
@@ -0,0 +1,150 @@
...
+        '--binder',
+        default='https://mybinder.org',
+        help="""
+        The URL of the binder instance to use.
+        Use `http://localhost:8585` if you are doing local testing.
+    """)
+    opts = parser.parse_args()
+
+    gun_time = time.time()
+    with ThreadPoolExecutor(max_workers=10) as executor:
```
minrk: Make concurrency an input here as well

scripts/load-test.py
```diff
@@ -0,0 +1,150 @@
...
+                    print('Launch %r succeeded and took %is' %
+                          (idx, launch['end'] - launch['start'])
+                          )
+                else:
+                    print('Launch %r failed and took %is' %
+                          (idx, launch['end'] - launch['start'])
+                          )
+
+    with open(opts.results, 'wb') as f:
+        pickle.dump({'gun_time': gun_time,
```
minrk: csv or json might be a nicer dump format than pickle, since these are simple numbers.

scripts/load-test.py
betatim: Do you have any experience with how big we can make this number? On my test on Saturday morning with 100 launches and `max_workers=10` (or so) the launches ended up spread out over 100 or so seconds. I will see what happens if you set `max_workers=100` but if this is known to not be a useful avenue it might be worth switching to async'ing all the things instead of this threading malarky.

minrk:
I should also mention [hubtraf](https://github.com/yuvipanda/hubtraf) which has been used for hub load tests in the past, in case it's relevant here. It's just for JupyterHub, but might be something to consider building on in the future.

betatim:
Sorry for the slow response, I thought I had left comment saying: with this number of threads 100 launches are spread over 100s or so. Is it feasible to bump up the # of threads or should I just stop now and go async?

I hadn't seen hubtraf, I found a different load test by Yuvi but that was in JS and running on a k8s cluster so I thought I'd "quickly" create a small script that I can run locally, fetches some extra pages for more "network traffic", and is in Python. I'll re-evaluate.

betatim:
https://grafana.mybinder.org/d/nDQPwi7mk/node-activity?refresh=1m&orgId=1&from=1531543333345&to=1531554133345 shows two attempts. One launching 100 binders in a timespan of 10s, second launching 200 over 10s. During the second one we had to spawn a new node.

Interesting point: why so many docker image pulls? SHouldn't we be pulling the built image just once per node not `n_launches_before_image_present` times?

betatim:
https://gist.github.com/betatim/e7a8fb98730c9b40b31573d5c14fc489 results of the benchmark

betatim:
Can we merge this? I don't plan to do much more on this in the near future.

The most interesting result from this is that the time to launch keeps increasing (see last plot in the notebook). Another question is why there are so many [image pulls](https://grafana.mybinder.org/d/nDQPwi7mk/node-activity?refresh=1m&orgId=1&from=1531543333345&to=1531554133345) was expecting ~1 per node. To launch 200 instances in parallel I used two copies of the same repo. I think all the pulls are for the second repo.

PR #663, opened 2018-07-12
Update singleuser_extra_pod_config to extra_pod_config

betatim:
This updates our extra config snippet.

I am not sure why we need this at all and the proper fix is to [change this line](https://github.com/jupyterhub/kubespawner/blob/7ae99005409cb75684a0a1ec792dba4adcb4b4f2/kubespawner/objects.py#L160) in kubespawner directly. At least in my local tests setting `restartPolicy` does not influence the value of `restart_policy` in the string that `spec.to_str()` generates, but there is also no error when you set a incorrect property (seems like a bug in `kubernetes`).

A further mystery is that the values from `extra_pod_config` are put through [an attribute map](https://github.com/jupyterhub/kubespawner/blob/7ae99005409cb75684a0a1ec792dba4adcb4b4f2/kubespawner/objects.py#L245) so there it should work with the wrong name. But evidently doesn't, my hypothesis is that something with the deprecation of the old name of `singleuser_extra_pod_config` doesn't work properly.

Going to deploy this to see if it fixes up our current deployment, then consider a PR to kubespawner, and then discussion on why we have this extra config snippet in the first place.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -73,7 +73,7 @@ binderhub:
         cors: *cors
       extraConfig:
         neverRestart: |
-          c.KubeSpawner.singleuser_extra_pod_config.update({'restartPolicy': 'Never'})
+          c.KubeSpawner.extra_pod_config.update({'restart_policy': 'Never'})
     proxy:
       service:
         type: ClusterIP


betatim:
Getting this deployed took several attempts and a bit of apply brute force.

First travis timed out because the new hub pod could not take over a volume mount because the odl hub pod was not shutting down. Discussion in jupyterhub/zero-to-jupyterhub-k8s#757 and jupyterhub/zero-to-jupyterhub-k8s#754.

Previously a way to fix it was to delete the running hub pod, and it would all sort itself out. Did this but then ended up with two hub pods. To fix this I manually edited the old replicaset used for the hub deployment and set the `replicas` property to 0 for that. This seemed to sort out the number of hub pods that were trying to run.

Finally I restarted the travis build again which now succeeded.

I think we should invest (even more) effort into testing in the upstream projects like kubespawner and zero2jupyterhub so that we don't cause ourselves stress by having bugs creep into the deployment. Ideally the upstream projects would do this for us but I can see that extensive testing under all sorts of weird conditions isn't the highest priority of contributors there. Maybe we can ply the role of "boring old person" that try to slow down the pace of changes and balance the desire for fast paced changes from other contributors.

betatim:
More extensive testing with a "real" k8s setup is a great idea! I think donating or providing the resources for doing this would be an easy way for some of the many users of the z2jh guide to give back to the community.

The resources used to operate mybinder.org should be reserved for operating mybinder.org as they come from a grant (or in the future other resources) specifically for this.

choldgraf:
FWIW your patch seems to have really returned the pod behavior to a state of normalcy:

image

we had thought this might be a change in kubernetes packing behavior, but it looks like it was the restart policy *shrug*

PR #675, opened 2018-07-21
add notes about new node pools on prod

minrk:


docs/source/command_snippets.md
```diff
@@ -112,23 +122,43 @@ gcloud beta --project=binder-prod container node-pools create $new_pool \
     --num-nodes=2 \
     --enable-autoscaling \
     --enable-autorepair \
-    --min-nodes=2 \
+    --min-nodes=1 \
     --max-nodes=8
+
+old_core_pool=core-pool
```
choldgraf: should these be commented out to make it clear they're just for show and not actually used in the script?

choldgraf: ah nevermind, I see we use it later. Maybe first have a short explainer in its own code block like

First we'll create variables that point to our old and new node pools to make it clear when we're creating new things vs. deleting old things.

```
old_pool=ssd-pool
new_pool=hm32
```

docs/source/command_snippets.md
```diff
@@ -112,23 +122,43 @@ gcloud beta --project=binder-prod container node-pools create $new_pool \
...
+new_core_pool=core-pool-1-11
+
+gcloud beta --project=binder-prod container node-pools create $new_core_pool \
```
choldgraf: do we still need beta? just double-checking

docs/source/command_snippets.md
```diff
@@ -112,23 +122,43 @@ gcloud beta --project=binder-prod container node-pools create $new_pool \
     --num-nodes=2 \
     --enable-autoscaling \
     --enable-autorepair \
-    --min-nodes=2 \
+    --min-nodes=1 \
     --max-nodes=8
+
+old_core_pool=core-pool
+new_core_pool=core-pool-1-11
```
choldgraf: also these are just arbitrary names right? there's no way to upgrade w/o choosing a new name?

choldgraf:
just a couple nits, looks good to me! thanks for adding this in!

docs/source/command_snippets.md
consideRatio: Add this for improved future compatibility with optional features introduced in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/758
```
--node-labels=hub.jupyter.org/node-purpose=core \
```

docs/source/command_snippets.md
```diff
@@ -112,23 +122,43 @@ gcloud beta --project=binder-prod container node-pools create $new_pool \
     --num-nodes=2 \
```
consideRatio: Add this for improved future compatibility with optional features introduced in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/758
```
--node-labels=hub.jupyter.org/node-purpose=user \
```

minrk:
Added the notes about node-labels and removed the nodeSelector update step, since that's no longer needed because node-labels take care of that.

docs/source/production_environment.md
```diff
@@ -46,6 +46,24 @@ to each other as possible. We want all common config in `values.yaml` so testing
...
 
+## Deployment nodes and pools
+
+The staging cluster has one node pool, which makes things simple.
+The production cluster has two, one for "core" pods (the hub, etc.)
+and another dedicated to "user" pods (builds and user servers).
+This strategy helps protect our key services from potential issues caused by users and helps us drain user nodes when we need to.
+
+Since only user pods should be running on the user nodes,
+cordoning that node should result it in being drained and reclaimed
```
betatim: "it in" -> "in it"

docs/source/production_environment.md
```diff
@@ -46,6 +46,24 @@ to each other as possible. We want all common config in `values.yaml` so testing
...
+after the max-pod-age lifetime limit
+which often wouldn't happen without manual intervention.
+
+In the future, when we implement a pod packing strategy,
+nodes could get reclaimed
```
betatim: missing full stop or words?

betatim:
Merge away after the above two comments.

docs/source/production_environment.md
minrk: Indeed. Forgot to finish this section. Should be set now.

docs/source/production_environment.md
```diff
@@ -53,16 +53,23 @@ The production cluster has two, one for "core" pods (the hub, etc.)
...
+Since ~only user pods should be running on the user nodes,
+cordoning that node should result in it being drained and reclaimed
 after the max-pod-age lifetime limit
 which often wouldn't happen without manual intervention.
 
-In the future, when we implement a pod packing strategy,
-nodes could get reclaimed
+It is still *not quite true* that only user pods are running on the user nodes at this point.
+There can be some pods such as heapster and kube-dns that may run on user nodes,
+and need to be manually removed from the pod after cordoning before the autoscaler will allow culling.
```
consideRatio: The autoscaler will scale down nodes that does not contain any blocking pods.

Blocking pods include:
* pods without a controller, such as the pods created by KubeSpawner (They don't belong to a Deployment)
* system critical pods, such as kube-dns
* pod that has a disruption policy that does not tolerate the pod to be rescheduled or removed

The heapster pod is a pod created by a daemonset though right? It wont block scaling down. The kube-dns will though. If you want to avoid having the kube-dns schedule there you can taint the user nodes, but that will influence the heapster pods as well, I wonder if one can allow them anyhow somehow... 

When I have added some additional infrastructure in z2jh, it will be very easy to setup all user related pods (image pullers, actual users and the future user-placeholder and user-dummies) to tolerate a taint on user nodes as well as require them to schedule there.

consideRatio: Btw, a PDB can be created to allow the kube-dns pod to allow it to be rescheduled I think, but I'm not so confident since it has system-critical annotation on it still.

docs/source/production_environment.md
```diff
@@ -53,16 +53,23 @@ The production cluster has two, one for "core" pods (the hub, etc.)
...
+
+In the future, when we implement a pod packing strategy and node taints,
+nodes could get reclaimed truly automatically without any intervention,
+but we are not there yet.
 
 Users and core pods are assigned to their pools via a `nodeSelector` in `config/prod.yaml`.
-We have one pool of each, which has a name.
+We use a custom label `mybinder.org/node-purpose = core | user`
```
consideRatio: If you use `hub.jupyter.org/node-purpose = core | user` you will be aligned with features I've already developed for z2jh, even though they are not merged yet.

minrk: The node docs actually apply both labels for new pools, so we can get there. It's not time to do it yet, though.

choldgraf:
are we ready for a merge here? :-)

minrk:
> are we ready for a merge here? :-)

I think so!

PR #680, opened 2018-07-30
incident (jlab build takes over, 30 July 2018) and docs update

choldgraf:
an incident report for a recent outage where jupyterlab-demo builds were taking a long time and eating up all the binderhub CPU


choldgraf:
cc @yuvipanda @betatim @ian-r-rose @Carreau as y'all were in the thick of it figuring it all out together :-D

choldgraf:
fixed

PR #681, opened 2018-08-01
unban jlab demo

choldgraf:
This unbans the jupyterlab demo repository.

I'm a bit confused about banning now, since it seems that the `spacy` binder is also banned, but AFAIK that website still works. @minrk do you know what's up?


betatim:
This is a ban from the "old" repo they used initially. It was used by a version of https://github.com/ines/juniper that had a bug that causes lots of launches (open event stream) so we banned it. They switched to a [new repo](https://github.com/ines/spacy-io-binder) after fixing things. We kept the ban in place for a while because the buggy JS was on a CDN and didn't seem to expire. Might be safe to remove it now.

PR #685, opened 2018-08-07
Update repo2docker and bump r2d prefix

betatim:
This is a  repo2docker version bump. See the link

https://github.com/jupyterhub/repo2docker/compare/d4c9c88...aee37be


betatim:
This also bumps the r2d build prefix. Let's see how this goes. Buckle up ;)

PR #692, opened 2018-08-20
add drain-user-node script

minrk:
deletes replicaset-controlled pods on a node after cordoning

tries to ensure nothing on the node is going to prevent the autoscaler reclaiming it once user pods finish on the node.

cc @betatim

### DIFF
--- a/scripts/drain-user-node
+++ b/scripts/drain-user-node
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+"""
+Drain a cordoned node with user pods,
+leaving user pods to finish draining on their own.
+
+To be run after cordoning a user node,
+to increase the likelihood of that node being reclaimed promptly
+by the autoscaler.
+
+Drains replicaset-controlled pods *other than user pods* from a node
+so that the autoscaler can reclaim it when it's ready.
+
+This drains pods such as `kube-dns` that can end up blocking
+autoscaler reclamation without disrupting user pods.
+"""
+
+import argparse
+import sys
+from kubernetes import config, client
+
+# Setup our parameters
+argparser = argparse.ArgumentParser(description=__doc__)
+argparser.add_argument(
+    "--dry-run",
+    action="store_true",
+    help="Dry run (report what would happen, don't actually do anything)",
+)
+argparser.add_argument(
+    "-y", dest="answer_yes", action="store_true", help="Answer yes (skips confirmation)"
+)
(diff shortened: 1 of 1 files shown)


scripts/drain-user-node
```diff
@@ -0,0 +1,154 @@
...
+kube_context_help = (
+    "Context pointing to the cluster to use. To list the "
+    "current activated context, run `kubectl config get-contexts`. Default: currently active context"
+)
+argparser.add_argument("--kube-context", default=None, help=kube_context_help)
+args = argparser.parse_args()
+
+# prefix to avoid scaring people during output
+if args.dry_run:
+    dry_prefix = "(Not actually) "
```
betatim: 😂

scripts/drain-user-node
```diff
@@ -0,0 +1,154 @@
...
+            # after deletion
+            to_delete.append((pod, f"owned by {owner.kind}"))
+            continue
+        elif owner:
+            print(f"Skipping {name} owned by {owner.kind}")
+        else:
+            # allow deleting stopped pods
+            if pod.status.phase in {"Failed", "Completed"}:
+                to_delete.append((pod, f"Pod {pod.status.phase}"))
+            if not (
```
betatim: Ouf. When do we arrive here? The first time I read it I thought it was missing an `else` with reference to line 115.

Now I am thinking we will go here for all pods that aren't owned by anyone .. and if the `if` is true we skip adding the pod to the delete list, but if the clause is `False` we also don't do anything.

Maybe add a comment on why we are here/want the situation is when we get here.

scripts/drain-user-node
```diff
@@ -0,0 +1,154 @@
...
+        + "\n  ".join(
+            "{} {}".format(pod.metadata.name, reason) for pod, reason in to_delete
+        )
+    )
+    if not args.answer_yes and not args.dry_run:
+        ans = input(f"Delete {delete_count} pods [y/N? ")
+        if not ans.lower().startswith("y"):
+            print("aborting...")
+            sys.exit(0)
+    # actually do the delition
```
betatim: delition -> deletion

scripts/drain-user-node
minrk: Added a comment. It's checking if it's a user pod (build or server) and only logging ownerless pods that *aren't* those. The goal is to avoid a flood of "Not deleting jupyter-..." output

PR #712, opened 2018-09-09
Set up Matomo (piwik) for mybinder.org

yuvipanda:
Currently we use Google Analytics for analysis. This works great,
but has two problems:

1. We're leaking user data to Google. This has a host of long
   term privacy problems for our users.
2. Our analytics data is locked into Google Analytics. There is
   no way to automatically get long term usage data out of
   Google Analytics without paying a lot of money for it.

[Matomo](https://matomo.org/) (formerly Piwik) is pretty much
the only open source analytics game in town. It works reasonably
well, scales ok, has a nice UI not too different from Google Analytics,
and provides for data export.

It requires a MySQL database. We use the Google Cloud hosted
MySQL (https://cloud.google.com/sql/docs/) instead of running our
own. This has no lock-in (since it is just MySQL) and frees us
from having to run a database ourselves - win-win!

We should run Google Analytics & Piwik side-by-side for a while,
and turn off Google Analytics once we are comfortable with Matomo.

TODO:
- [x] Document creating of MySQL cloud instance & service accounts
- [x] Add Ingress object


yuvipanda:
Check out https://demo.matomo.org for a demo of the UI.

betatim:
What kind of data are we going to be collecting if we make the switch? Right now the GA settings are set to minimum collection, no cookies, etc. We should make sure we consciously change the amount of tracking we do. Mostly so we aren't surprised later on.

If we do increase the amount of information we collect and store we should revisit GDPR and friends to double check if we need to have a short blurb on the webpage to inform people or not.

yuvipanda:
@betatim yes, absolutely. We should:

1. List the questions we want answered
2. Collect just enough data to answer those questions
3. Export processed aggregated data publicly regularly
4. Not keep any (anonymized) raw data for more than 30 days

mybinder/templates/matomo/deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+  replicas: {{ .Values.matomo.replicas }}
+  selector:
+    matchLabels:
+      app: matomo
+  template:
+    metadata:
+      labels:
+        app: matomo
+    spec:
+      volumes:
```
minrk: This deployment needs a nodeSelector and/or pdb in order to avoid getting in the way of scale-down

mybinder/templates/matomo/deployment.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: matomo
+  labels:
+    app: matomo
```
minrk: missing release, heritage labels

mybinder/templates/matomo/deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+          - /cloud_sql_proxy
+          - -instances={{ .Values.matomo.db.instanceName }}=tcp:3306
+          - -credential_file=/secrets/cloudsql/credentials.json
+        securityContext:
+          runAsUser: 2  # non-root user
+          allowPrivilegeEscalation: false
+        volumeMounts:
+          - name: cloudsql-instance-credentials
+            mountPath: /secrets/cloudsql
+            readOnly: true
```
minrk: ingress.yaml below created an empty file?

mybinder/templates/matomo/deployment.yaml
yuvipanda: Added!

mybinder/templates/matomo/deployment.yaml
yuvipanda: Fixed!

yuvipanda:
@minrk thanks for the review! I've addressed the issues you raised, and added a little docs. I'm going to try get this going on Tuesday!

betatim:
I think the [matomo GDPR guide](https://matomo.org/docs/gdpr/) is going to be excellent in getting us up and running and compliant. In terms of the four questions/steps: I would follow the minimum of their GDPR guide that lets us avoid GDPR issues and start with that. Then see if we crave more detail and discuss that in a new issue (if and when it comes up).

A slight counter point to deleting data ASAP/within 30 days. Quickly being able to pull up stats, make graphs, get usage numbers is very helpful for conversations with others to whom we want to "demonstrate our impact". I'd be a big supporter of keeping the data in a place where we can make simple queries and pretty plots with a few clicks. For example yesterday at EPFL in response to "Does anyone here actually use this mybinder.org thing?" it was great to be able to pull up our GA and instantly answer that question: "looks like some people use mybinder.org from your region of the world, we'd have to look at these hundreds of repos to figure out who/what exactly they do, but here are two I know are researchers at your university." For this to work we need more than 30days of data though because in most places we aren't popular enough to make good statements based on 30days alone.

docs/source/index.rst
```diff
@@ -22,6 +22,7 @@ Components
    components/dashboards.md
    components/ingress.md
    components/cloud.md
+   components/matomo.rst
```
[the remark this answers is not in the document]
yuvipanda: Whoops, no! Git fuckup, will fix!

yuvipanda:
@betatim oh I agree! I want us to keep aggregated data forever - things like launches per repo, some super broad geographic data, etc. But we should explicitly figure out what kinda data we want long term, and store only that. Everything else should be considered 'raw' and purged.

choldgraf:
I'm +1 on just replicating exactly what data we're collecting from GA now, and solve the more nuanced question of "what kind of data *should* we be collecting from a strategic standpoint" in a separate thread (AKA I don't think answering that question needs to block switching from GA to matomo)

yuvipanda:
@choldgraf yes, no disagreements.

This PR doesn't actually collect any data even :) It just sets up Matomo. Need PRs to BinderHub before we can do that.

PR #714, opened 2018-09-13
apply coreNodeSelector to tiller

minrk:
avoids tiller landing on user nodes

`helm init` has a nice `--node-selectors` arg, but it only works on first init, not with `helm init --upgrade`, so we have to use `kubectl patch`.

[ref](https://github.com/helm/helm/issues/4063)


minrk:
This deploy failed because the patch wasn't quite right. Should be fixed in #716

Issue #725, opened 2018-09-18
Set up Matomo as alternative to Google Analytics


yuvipanda:
Tracking ticket for setting up [Matomo](https://matomo.org/) on mybinder.org.

- https://github.com/jupyterhub/mybinder.org-deploy/pull/712 set up initial matomo deployment.
- #724 fixed up syntax error in #712

yuvipanda:
Matomo is failing to start, with error:

```
chown: changing ownership of './config/config.ini.php': Read-only file system
```

It's trying to chown the config.ini.php file, but since we're projecting it with a configmap, it can't. I'd like it to not be able to change the config.ini.php file. I'll look around to see what we can do.

yuvipanda:
It's up and running now! \o/

Changes in admin panel seem to want to write changes to filesystem. This is a problem for us, since our filesystem is emphemeral. Need to figure out the right set of plugins & settings, and put that into our deployment system.

yuvipanda:
Next steps are:

- [ ] Figure out how to persist config changes over pod restarts
- [ ] Observe reported data between Google Analytics & this to watch for discrepancies
- [ ] Watch resource usage of pod (CPU, RAM) to see where we should set limits & requests
- [ ] Run at least 3 copies of Matomo, so we provide uninterrupted service

yuvipanda:
Currently, Matomo is underreporting visitors by about 2x. Need to figure out what's going on!

yuvipanda:
@betatim helped me realize that I was looking at the wrong metric, and Matomo is reporting things appropriately!

Now we need to tweak some privacy settings, and then see how metrics track between GA and Matomo over a few months before we can consider turning GA off.

PR #737, opened 2018-09-19
Add a bit more docs on matomo setup

yuvipanda:
Ref #725


betatim:
Good work on docs in real time!

https://mybinder-sre.readthedocs.io/en/latest/components/matomo.html#initial-installation there is some formatting mishap :-/

PR #738, opened 2018-09-19
Add Cloud Firewall Rules for Matomo

yuvipanda:
Ref #725 

Currently, these have been hand created. That should be removed &
the script should be run under supervision.


yuvipanda:
@minrk I've never run the firewall script manually. Would be awesome if you have time to merge this and run it!

minrk:
Merged and run. After #699, the tight egress firewall should be redundant with network policies, so it can be removed. I'll look into that shortly.

PR #755, opened 2018-10-02
Increase matomo replicas to 3 & make at least 1 available

yuvipanda:
We don't wanna lose metrics if possible. So we run 3 copies
of matomo, and make sure at least 1 is available during
scaling and other changes.

Ref #725

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -167,6 +167,7 @@ redirector:
         to: mybinder.readthedocs.io/en/latest/status.html
 
 matomo:
+  replicas: 3
   nodeSelector: *coreNodeSelector
   db:
     instanceName: binder-prod:us-central1:matomo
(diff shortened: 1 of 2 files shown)


betatim:
This looks reasonable, but my pdb knowledge is small.

Is there something in here to make sure the three pods are not all on the same node? A problem with the node seems like the most likely reason for matomo getting into trouble. This also means we need to discuss the core vs user pool story again (see #752). We only have one core node right now.

yuvipanda:
@betatim there isn't, but since there is only one core node there isn't much we can do there. I would like us to eventually move away from the concept of core nodes if possible, but that requires that all our components are fully HA.

yuvipanda:
@minrk @betatim I think with this, we can probably move matomo from core pool pinning to no pinning?

Issue #766, opened 2018-10-14
https://docs.mybinder.org/ cert is invalid


betatim:
Firefox tells me:
```
docs.mybinder.org uses an invalid security certificate.
The certificate is not trusted because it is self-signed.
The certificate is only valid for ingress.local. 
Error code: MOZILLA_PKIX_ERROR_SELF_SIGNED_CERT
```

betatim:
We seem to be constantly requesting certificates for our redirect domains: https://crt.sh/?q=%25.mybinder.org

minrk:
This is fixed by #770

PR #767, opened 2018-10-14
Remove status.mybinder.org redirect

betatim:
The redirect currently fails and we have triggered the Let's encrypt
rate limit which means docs.mybinder.org has become unavailable. For the
moment removing status.mybinder.org to test if we can recover.

Related to #766

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -161,10 +161,6 @@ redirector:
       host:
         from: playground.mybinder.org
         to: play.nteract.io
-    - type: host
-      host:
-        from: status.mybinder.org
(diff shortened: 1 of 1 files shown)


betatim:
This was a/the problem with status.mybinder.org:

```
time="2018-10-14T13:10:53Z" level=warning msg="authorization failed after 1m0s: reachability test failed: Get http://status.mybinder.org/.well-known/acme-challenge/_selftest: dial tcp: lookup status.mybinder.org on 10.15.240.10:53: no such host" context=acme domain=status.mybinder.org
```

From the kubelego logs and confirmed by trying it in my browser. I suspect we broke the ingress config at some point?

PR #768, opened 2018-10-14
Upgrade kube-lego version

betatim:
While investigating our HTTPS setup I noticed that we run an old version of kube-lego. This takes us to the latest available version before it became unmaintained.

I am not sure if this will fix the problem in #766 or not but it seems like it is worth a shot.

Would be good to get @minrk or @yuvipanda to take a look at this before deploying it. We have been rate limited for seven days with let's encrypt so not in a mega hurry to get it fixed as it is going to be a while before we get unbanned anyway. At least that is my understanding.

### DIFF
--- a/mybinder/requirements.yaml
+++ b/mybinder/requirements.yaml
@@ -9,7 +9,7 @@ dependencies:
    version: 1.2.0
    repository: https://kubernetes-charts.storage.googleapis.com
  - name: kube-lego
-   version: 0.1.12
+   version: 0.4.2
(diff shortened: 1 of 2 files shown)


minrk:
I believe this will fix the problem, but not immediately. From the updated logs, it would appear that latest kube-lego knows about the rate limit and how to wait, but we still need to wait for the current rate limit to run out before we will get the new certs for the redirector.

I'm additionally going to try #770 to see if a different domain combination allows us to refresh without the rate limit

Issue #760, opened 2018-10-05
Mount /etc/jupyter as a volume for custom config


minrk:
It would be useful to load extra Jupyter config from a *volume* rather than relying on the appendix. We can do this by mounting `/etc/jupyter` as a volume, e.g. from a configmap.

Two reasons:

1. it gets applied at launch time instead of build time (doesn't require rebuild to get it on all images)
2. it gets applied to custom Dockerfile images, which skip the appendix

The only downside of this is that images can technically put config in /etc/jupyter (only custom Dockerfile images should have the ability to do this), so there is a slim technical possibility that we could be overriding the intended behavior of the image.

For example, the Jupyter docker-stacks [do this](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/jupyter_notebook_config.py), but only to set various docker-appropriate defaults that are irrelevant for Binder, so it would be no problem to ignore it in that case.

PR #788, opened 2018-10-29
update Notebook UI via appendix

bitnik:
Hi, sorry I had had to do this PR before as discussed in https://github.com/jupyterhub/binderhub/issues/674. If this change is not needed anymore, we can close this PR.

This PR updates appendix in order to update Notebook UI:
1. Removes logout button (I think this button is confusing, since users are not logged in and there is now Quit button.)
2. Adds binder buttons:
  - `Go to repo`: opens the source repo url in new tab
  - `Copy binder link`: copies the binder launch link into clipboard
  - `Copy session link`: copies the binder session link into clipboard. When this link is shared with another user, that user will reach to the same binder session. It will not start a new launch.

I can remove or rename buttons according to your feedback.

### DIFF
--- a/appendix/run-appendix
+++ b/appendix/run-appendix
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -euo pipefail
+set -x
+appendix_dir="$(dirname "$0")"
+# add custom.js for notebook
+# it is executed when the notebook app starts and adds binder buttons next to Quit button in the UI
+sed -i 's|{binder_url}|'"$BINDER_URL"'|g' "${appendix_dir}/static/custom.js"
+sed -i 's|{repo_url}|'"$REPO_URL"'|g' "${appendix_dir}/static/custom.js"
+mkdir -p ~/.jupyter/custom
+cat "${appendix_dir}/static/custom.js" >> ~/.jupyter/custom/custom.js
+chown -R $NB_USER:$NB_USER $HOME/.jupyter
--- a/appendix/static/custom.js
+++ b/appendix/static/custom.js
@@ -1,0 +1,35 @@
+function copy_link_into_clipboard(b) {
+    var $temp = $("<input>");
+    $(b).parent().append($temp);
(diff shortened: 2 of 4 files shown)


minrk:
Waiting a bit on this one, since implementing #760 (#792) will remove the appendix altogether, in favor of a better implementation, I think.

betatim:
Do you want to refresh this PR now that those PRs have landed?

What do people think about having the "copy session link"? I feel a bit uneasy because it would encourage people to share secrets with others and gives them a chance to copy the wrong link to twitter. However it is a useful feature that I use ... at the moment I would say we should not show it and leave it as a "expert user" feature that you can share sessions.

bitnik:
I just realised that after https://github.com/jupyterhub/mybinder.org-deploy/pull/792, `binder_url` in [login template of notebook](https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/files/etc/jupyter/templates/login.html) is always `https://mybinder.org` because there is no more `BINDER_URL` env variable (see https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/files/etc/jupyter/jupyter_notebook_config.py#L4).

I didn't intend to do it but this PR should fix it too. Even we don't merge this PR, I think `BINDER_URL` env variable should be defined in appendix, so each image has its binder url.

bitnik:
@minrk  @betatim I rebased this once again :)

minrk:
After #792, we shouldn't be using the appendix at all anymore, so this needs to be accomplished to achieve the same goal without use of the appendix.

In particular, I think it should be in the form of an nbextension, not custom.js, which can be loaded from /etc/jupyter.

bitnik:
Alright, I misunderstood. I am closing this PR. I will think about how to do the same without using appendix at all as @minrk suggested.

choldgraf:
@bitnik thanks for the first pass! sorry for the noise :-/

PR #792, opened 2018-10-31
move custom config to etc/jupyter

### LINKED ISSUE #760: Mount /etc/jupyter as a volume for custom config

minrk:
- remove appendix
- mount config at /etc/jupyter via configmap
- use files for jupyter_notebook_config.py, templates
- load etcJupyter.jupyter_notebook_config.json as custom config

This makes adding custom config for servers much easier (edit values.yaml), and enables config for all images, regardless of build (previously, Dockerfiles would skip our appendix), and doesn't require killing the image cache to get new values for existing builds

closes #760

I'd like to see how this goes here, but at least some of this (namely singleuser extra config mounted at /etc/jupyter_notebook_config.json) is generic enough that it should probably be promoted to z2jh


mybinder/templates/user-configmap.yaml
```diff
@@ -0,0 +1,16 @@
...
+  {{ $name }}: {{ $cfg | toJson | quote }}
+  {{- end }}
+  {{- (.Files.Glob "etc/jupyter/*").AsConfig | nindent 2 }}
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: user-etc-jupyter-templates
+data:
+  {{- (.Files.Glob "etc/jupyter/templates/*").AsConfig | nindent 2 }}
```
minrk: Need one configmap per directory because configmaps cannot contain subdirectories

mybinder/etc/jupyter/jupyter_notebook_config.py
```diff
@@ -0,0 +1,5 @@
+import os
```
yuvipanda: Can we put this file under a 'files/' subdir? Not sure if helm supports that.

mybinder/templates/user-configmap.yaml
```diff
@@ -0,0 +1,21 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: user-etc-jupyter
```
yuvipanda: Should probably add labels to the configmaps too.

PR #814, opened 2018-11-05
script cleaning up images on old prefixes

### LINKED ISSUE #800: Delete images from registry if they aren't using current prefix
We spend about 500$ a month on our image storage. We have kept all images ever, and I suspect that if we delete images that don't match the current prefix, that cost will go down a lot.

minrk:
This is a script that talks to the docker registry api at gcr.io. It is not run automatically, and probably shouldn't be because we don't want to delete old images the instant we bump the prefix. Maybe wait a week after a bump before we do that, which is harder to automate.

Anyone should be able to run it, though, since it only consumes the credentials in this repo.

Testing it out right now. This takes a really long time.

closes #800

### DIFF
--- a/scripts/delete-old-images.py
+++ b/scripts/delete-old-images.py
@@ -2,79 +2,72 @@
 """Cleanup images that don't match the current image prefix"""
 
 import asyncio
-from asyncio.subprocess import create_subprocess_exec, PIPE
+import atexit
 import json
-
-jsonmod = json
+from functools import partial
 from multiprocessing import cpu_count
 import os
 import pipes
-from subprocess import check_call
 import sys
+import time
 
+import aiohttp
 import tqdm
 import yaml
 
 HERE = os.path.dirname(__file__)
 
-# limit on the number of concurrent outstanding requests
-# gcloud api uses a surprising large amount of CPU,
-# so don't oversubscribe by default
-process_pool = asyncio.BoundedSemaphore(cpu_count())
 
-
-def setup_auth(release):
-    return check_call(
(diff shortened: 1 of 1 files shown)


minrk:
This script is now pretty stable, and has deleted ~30k old builds.

After lots of partial runs stopping due to timeouts, etc. the latest run has taken two hours to delete 18k builds, with 4k left.

minrk:
This script should be good to go. 

Further things we can explore in subsequent PRs:

- remove old builds of images that have lots (e.g. limit an image to its most recent 5 or 10 builds)

choldgraf:
this is great! one question: where is it documented? :-)

If you give a short-hand example of how it's run, I can make a PR to the SRE guide, if that's helpful

minrk:
@choldgraf good catch! Added better docstrings to the script. A PR to the SRE guide with some notes would be awesome.

scripts/delete-old-images.py
```diff
@@ -0,0 +1,266 @@
...
+        total_images = 0
+        async for image in list_images(session, project):
+            total_images += 1
+            if f"gcr.io/{image}".startswith(prefix) and not any(
+                ci_string in image for ci_string in CI_STRINGS
+            ):
+                matches += 1
+                continue
+            # don't call ensure_future here
+            # because we don't want to kick off everything before
```
betatim: this still puzzles me, it says that we don't use ensure_future, but we do on L173. Is there somewhere else that `ensure_future` could be used but isn't in the next lines?

PR #841, opened 2018-11-11
binderhub bump 80ac52c...6578e4d

choldgraf:

https://github.com/jupyterhub/binderhub/compare/80ac52c...6578e4d

### DIFF
--- a/mybinder/requirements.yaml
+++ b/mybinder/requirements.yaml
@@ -12,5 +12,5 @@ dependencies:
    version: 0.4.2
    repository: https://kubernetes-charts.storage.googleapis.com
  - name: binderhub
-   version: 0.2.0-80ac52c
+   version: 0.2.0-6578e4d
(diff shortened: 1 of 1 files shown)


choldgraf:
cc @betatim

betatim:
@minrk should we have our wits about us when deploying the change to the CORS config? Otherwise I'd deploy this change this morning expecting that nothing happens beyond the badge changin.

minrk:
There aren't CORS changes to deploy here. Sometimes the diffs linked from chart updates will include merge commits for the *previous* deploy that don't contain actual changes because the chart versions are based on the 'last change commit,' which is typically the commit that actually changed the file, not the merge when it landed in master. When in doubt, check the file changes, not the commit list for what's to be deployed.

This should be comfortably deployed now if you want to hit the button.

betatim:
Thanks! I thought it was weird to only see the merge commit and no actual changes :-/

Deploy underway.

PR #842, opened 2018-11-12
give user pods access to https on the ingress controller

### LINKED ISSUE #838: Launch event archives not accessible from inside mybinder.org
The launch activity archives aren't accessible from a (notebook) pod running on mybinder.org:

```
$ wget https://archive.analytics.mybinder.org/events-2018-11-03.jsonl
--2018-11-10 20:51:32-- 
https://archive.analytics.mybinder.org/events-2018-11-03.jsonl
Resolving archive.analytics.mybinder.org (archive.analytics.mybinder.org)... 35.202.202.188
Connecting to archive.analytics.mybinder.org (archive.analytics.mybinder.org)|35.202.202.188|:443...
```

will eventually time out but works as expected from your laptop/computer outside the cluster.

https://mybinder.org is also not accessible so I think this is a network policy/firewall configuration issue. I think we should keep mybinder.org inaccessible from pods as it makes it harder to use a binder to launch binders.

ref #789

minrk:
grants access to the public entrypoints of the cluster

closes #838

### DIFF
--- a/mybinder/templates/netpol.yaml
+++ b/mybinder/templates/netpol.yaml
@@ -56,4 +56,16 @@ spec:
             - {{ $ipOrCidr}}/32
             {{- end }}
           {{- end }}
+    # allow https access to the ingress controller
+    # (needed to fetch public urls hosted on this cluster,
+    # e.g. analytics.mybinder.org)
+    - ports:
+        - port: 443
+          protocol: TCP
+      to:
+        - podSelector:
+            matchLabels:
+              app: nginx-ingress
+              component: controller
+
 {{- end }}


minrk:
I was able to test this manually by deploying an extra network policy with this rule on staging with `kubectl apply -f`.

It would be good to have some scaffolding to be able to run tests inside a user pod, so we could verify things like network access.

betatim:
You think it was the preceding rule that allows access to various ports for IPs in "the world but not the cluster" got used and these requests (even though they were to a public IP) got stopped by it?

Thoughts about blocking access to mybinder.org's endpoints that allow launching new binder instances?

Otherwise, merge time I think.

minrk:
> You think it was the preceding rule that allows access to various ports for IPs in "the world but not the cluster" got used and these requests (even though they were to a public IP) got stopped by it?

Yes, precisely. The `except: 10.0.0.0/8` prohibits direct egress from the user pod to anything in the cluster not whitelisted (this includes the ingress controller). Apparently, the network setup has a short-circuit such that requests to the public urls of binder are routed directly to the ingress controller instead of going out to the wilds of the Internet and then back through a public interface.


> Thoughts about blocking access to mybinder.org's endpoints that allow launching new binder instances?

I don't think this is feasible at the moment. It's so trivial to do mischief via the mybinder.org api from anywhere, that allowing it from inside binder doesn't really make a difference.

betatim:
Merged!

PR #849, opened 2018-11-15
Automatically publish cloud compute costs

### LINKED ISSUE #345: Automate cloud compute financial transparency
In #205 we discussed making our data about Google Cloud costs publicly-available. That's been done in https://github.com/jupyterhub/binder-billing and https://github.com/jupyterhub/binder-data, though not in an automated fashion. `binder-data` requires a manual `gsutil` run to bring in the latest set of data. This should be:

* changed to use the google cloud API in python
* updated to happen automatically

yuvipanda:
- Works with CSV & JSON exports
- Exports aggregate data to Google Cloud Storage

TODO:

- [x] Run automatically 
- [x] Add documentation on using and interpreting this
- [ ] Publicize this in the index page

Fixes #345

### DIFF
--- a/docs/source/analytics/cloud-costs.rst
+++ b/docs/source/analytics/cloud-costs.rst
@@ -1,0 +1,32 @@
+.. _analytics/cloud-costs:
+
+================
+Cloud Costs Data
+================
+
+In an effort to be more transparent about how we use our funds,
+we publish the amount of money spent each day in cloud
+compute costs for running mybinder.org.
+
+Interpreting the data
+=====================
+
+You can find the data in the `Analytics Archive
+<https://archive.analytics.mybinder.org>`_ at `cloud-costs.jsonl
+<https://archive.analytics.mybinder.org/cloud-costs.jsonl>`_. Each line in
+the file is a JSON object, with the following keys:
+
+#. **startTime** and **endTime**
+
+   The start and end of the billing period this item represents. These
+   times are inclusive, and in pacific time observing DST (so PDT or PST).
+   The timezone choice is unfortunate, but unfortunately our cloud provider
+   (Google Cloud Platform) provides detailed billing reports in this timezone
+   only.
+
+#. **cost**
+
+   The cost of all cloud compute resources used during this time period. This
+   is denominated in US Dollars.
+
+The lines are sorted by ``startTime``.
--- a/docs/source/analytics/cloud-costs.rst
+++ b/docs/source/analytics/cloud-costs.rst
@@ -4,7 +4,7 @@
 Cloud Costs Data
 ================
 
-In an effort to be more transparent about how we use our funds,
+In an effort to be transparent about how we use our funds,
 we publish the amount of money spent each day in cloud
 compute costs for running mybinder.org.
 
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -202,9 +202,11 @@ matomo:
 
 analyticsPublisher:
   project: binder-prod
+  destinationBucket: mybinder-events-archive
   events:
     sourceBucket: mybinder-events-raw-export
-    destinationBucket: mybinder-events-archive
+  billing:
+    sourceBucket: binder-billing
 
 gcsProxy:
   buckets:
--- a/config/staging.yaml
+++ b/config/staging.yaml
@@ -97,9 +97,9 @@ matomo:
 
 analyticsPublisher:
   project: binder-staging
+  destinationBucket: mybinder-staging-events-archive
   events:
     sourceBucket: mybinder-staging-events-raw-export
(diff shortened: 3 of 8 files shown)


yuvipanda:
I now think it's ok to push this out as is, see how it goes for a few days and *then* publicise this in the index page.

betatim:
> publicise this in the index page.

what do you mean with this?

docs/source/analytics/cloud-costs.rst
```diff
@@ -0,0 +1,32 @@
+.. _analytics/cloud-costs:
+
+================
+Cloud Costs Data
+================
+
+In an effort to be more transparent about how we use our funds,
```
betatim:
```suggestion
In an effort to be transparent about how we use our funds,
```

docs/source/analytics/cloud-costs.rst
```diff
@@ -0,0 +1,32 @@
...
+
+You can find the data in the `Analytics Archive
+<https://archive.analytics.mybinder.org>`_ at `cloud-costs.jsonl
+<https://archive.analytics.mybinder.org/cloud-costs.jsonl>`_. Each line in
+the file is a JSON object, with the following keys:
+
+#. **startTime** and **endTime**
+
+   The start and end of the billing period this item represents. These
+   times are inclusive, and in pacific time observing DST (so PDT or PST).
```
betatim: How hard is it to have all times in UTC instead of local timezones that have summer vs winter time? Would IMHO make it easier to look at the data if it is in UTC always

yuvipanda: @betatim unfortunately not easy, since google cloud provides us daily aggregates in PDT / PST only. I don't think we can really extrapolate that into UTC.

yuvipanda:
> > publicise this in the index page.
> 
> what do you mean with this?

I mean add a link on https://archive.analytics.mybinder.org/

docs/source/analytics/cloud-costs.rst
betatim: Ah, that is annoying. We can't change the "billing timezone" of the project or some such? (Never thought I'd use the word billing timezone)

yuvipanda: @betatim unfortunately, nope :'(

minrk: What do the timestamps look like? If they are reasonably normal timesetamps and have tzinfo in them, converting to utc can be done:

```python
from datetime import timezone
from dateutil.parser import parse as parse_date

def as_utc(timestamp):
    """convert string timestamp in any timezone to iso timestamp in utc"""
    return parse_date(timestamp).astimezone(timezone.utc).isoformat()
```

yuvipanda: @minrk @betatim ah, so we can convert them to UTC this way, but then they don't start / end on day boundaries. Is that ok?

betatim: What kind of data do we get at what level of resolution? The way I understood it is that we get one data point per day (from midnight PST to midnight PST).

betatim:
How does the billing info get into the input storage bucket in the first place? Is that written down somewhere and I'm overlooking it?

docs/source/analytics/cloud-costs.rst
minrk: If the timestamps are ISO8601 with UTC offset, whether they are UTC or not doesn't make a huge difference because anything that understands numerical timezone offsets (dateutil, pandas, python datetime, javascript) should handle them fine. If it's one data point per day, I'd probably leave them alone. I don't have a strong preference for UTC with odd start/end vs local timezone with natural start/end, so I'm happy to go with whatever other folks prefer, and I'd lean toward doing less if nobody else has a strong preference.

Rather than saying that it's "unfortunate" that this is the case and declaring the current timezone, we can just say that these are ISO 8601 timestamps with UTC offsets. That's easy enough for anyone to deal with. I would only really call it unfortunate if they were local timestamps without offsets, which would cause headaches.

betatim: Agreed. If we get one "data point" per day I wouldn't bother moving them around and then end up with a day that starts not at midnight.

yuvipanda: Cool! They already have UTC offsets, will amend documentation.

minrk:
Billing export is documented [here](https://cloud.google.com/billing/docs/how-to/export-data-file). It looks like the user gets to pick the prefix, so maybe that should be an input?

minrk:
Billing export is documented [here](https://cloud.google.com/billing/docs/how-to/export-data-file). It looks like the user gets to pick the prefix, so maybe that should be an input?

PR #851, opened 2018-11-22
bump binderhub f80738c...5be4d08

minrk:
This is a Binderhub version bump. See the link below for a diff of new changes:

https://github.com/jupyterhub/binderhub/compare/f80738c...5be4d08


minrk:
Deploying this one while staging has been upgraded to 1.11 before upgrading prod to 1.11

PR #859, opened 2018-11-26
adding about page text

choldgraf:
Now that https://github.com/jupyterhub/binderhub/pull/729 is in mybinder.org (mybinder.org/about) this attempts to configure the page to add the extra text that @betatim suggested. I used the documentation that was provided in that PR, and I *think* this is where the text needs to go in the helm chart but I'm not really sure. @betatim do you have any guidance there?

To try and demo the CSS and wording locally, I followed the pattern in the [binder docs](https://github.com/jupyterhub/binderhub/blob/master/CONTRIBUTING.md#pure-html--css--js-development) and copied a config file for local stuff. I added it to the repo in case it's useful to use for local development when we don't have access to minikube (I don't have it on my windows machine). Is there any way that we can import values into the localonly config *from* the helm chart? (e.g. everything that isn't explicitly over-written by the localonly config?) Maybe @yuvipanda or @minrk have ideas on that one?

Either way, here is what it looks like when I serve locally:

image

### DIFF
--- a/config/localonly/binderhub_config.py
+++ b/config/localonly/binderhub_config.py
@@ -1,0 +1,13 @@
+from binderhub.repoproviders import FakeProvider
+
+c.BinderHub.use_registry = False
+c.BinderHub.builder_required = False
+c.BinderHub.repo_providers = {'gh': FakeProvider}
+c.BinderHub.tornado_settings.update({'fake_build':True})
+
+c.BinderHub.about_message = ('<br /><p style="width: 650px; margin: 0px auto;">mybinder.org is public infrastructure operated by the <a href="https://jupyterhub-team-compass.readthedocs.io/en/latest/team.html#binder-team">Binder Project team</a>.<br /><br />'
+                             'The Binder Project is a member of <a href="https://jupyter.org">Project Jupyter</a>, which is a fiscally '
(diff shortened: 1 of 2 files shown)


config/localonly/binderhub_config.py
```diff
@@ -0,0 +1,13 @@
+from binderhub.repoproviders import FakeProvider
```
minrk: Is this file meant to be added? I don't think it will be used.

choldgraf:
+1 if you're ready for it

betatim:
Merged. Prepare to see text appear on the about page :)

betatim:
See the revert PR for another log snippet. I think we will need a second attempt at getting this deployed.

With https://github.com/jupyterhub/binderhub/pull/680 I think we can even move this to an entry here https://github.com/jupyterhub/mybinder.org-deploy/blob/2a668ca47a095948728a451766459b3bda3b5136/mybinder/values.yaml#L49:
```
BinderHub:
  about_message: ...
```

choldgraf:
hmmmm ok - it sounds like I just need to replace the `"` with `'`, yeah? I can give that a shot tomorrow AM

betatim:
That and/or moving it to a different place in the `values.yaml`.

PR #860, opened 2018-11-27
bump binderhub 5be4d08...2b368e4 and re-enable pod priority

minrk:
https://github.com/jupyterhub/binderhub/compare/5be4d08...2b368e4

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -172,9 +172,9 @@ binderhub:
         enabled: true
         replicas: 2
       podPriority:
-        enabled: false
+        enabled: true
       userPlaceholder:
-        enabled: false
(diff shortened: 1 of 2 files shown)


mybinder/values.yaml
```diff
@@ -172,9 +172,9 @@ binderhub:
         enabled: true
         replicas: 2
       podPriority:
-        enabled: false
+        enabled: true
       userPlaceholder:
-        enabled: false
+        enabled: true
         replicas: 0
```
consideRatio: 0 replicas makes the enabling of userPlaceholders pointless, but we can scale it manually though.

```
kubectl scale -n prod sts/user-placeholder --replicas 15
```

PR #861, opened 2018-11-28
25 placeholders on prod

minrk:
since it takes ~2 minutes to get a new node and we launch ~10 users per minute. This should trigger autoscale ~1 minute sooner than the current 10.

we can check [this chart](https://grafana.mybinder.org/d/nDQPwi7mk/node-activity?orgId=1&from=now-24h&to=now&panelId=43&fullscreen) to see if this seems like enough, and turn it up more, if we want.


betatim:
Deployed, because why not :)

PR #863, opened 2018-12-03
about message w passthrough

choldgraf:
Here's a second attempt at the about message...@betatim do you think this would get around the problem w/ the quotes? I changed all the start/stop string quotes to be single-spaced, and everything inside is double-spaced.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -50,6 +50,13 @@ binderhub:
       use_registry: true
       build_image: jupyter/repo2docker:9f081a2
       per_repo_quota: 100
+      about_message: |
+        <p>mybinder.org is public infrastructure operated by the <a href="https://jupyterhub-team-compass.readthedocs.io/en/latest/team.html#binder-team">Binder Project team</a>.<br /><br />
+        The Binder Project is a member of <a href="https://jupyter.org">Project Jupyter</a>, which is a fiscally
+        sponsored project of <a href="https://numfocus.org/">NumFocus</a>, a US 501c3 non-profit.<br /><br />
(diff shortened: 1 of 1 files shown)


mybinder/values.yaml
```diff
@@ -50,6 +50,13 @@ binderhub:
       use_registry: true
       build_image: jupyter/repo2docker:9f081a2
       per_repo_quota: 100
+      about_message: |
+        ('<p>mybinder.org is public infrastructure operated by the <a href="https://jupyterhub-team-compass.readthedocs.io/en/latest/team.html#binder-team">Binder Project team</a>.<br /><br />'
```
betatim: I think we can put one long string here. It doesn't have to be valid Python, when it arrives in Python it will be a string.

You can experiment a bit with `import yaml` but I think:
```
about_message: |
  <p>mybinder.org is public infra ...
  The Binder Project is a member ...

```
should work. So no `(` and `)` nor `'` around the lines.

choldgraf:
@betatim how does the latest push look?

mybinder/values.yaml
```diff
@@ -50,6 +50,13 @@ binderhub:
       use_registry: true
       build_image: jupyter/repo2docker:9f081a2
       per_repo_quota: 100
+      about_message: |
+        <p>mybinder.org is public infrastructure operated by the <a href="https://jupyterhub-team-compass.readthedocs.io/en/latest/team.html#binder-team">Binder Project team</a>.<br /><br />
+        The Binder Project is a member of <a href="https://jupyter.org">Project Jupyter</a>, which is a fiscally
```
choldgraf: will it be smart about adding a space for linebreaks that are between 2 words?

choldgraf: Woooo!

PR #870, opened 2018-12-13
bump repo2docker image b61e374...72d7634

minrk:
this is the 0.7.0 release

this is a repo2docker version bump. See the link below for changes

https://github.com/jupyter/repo2docker/compare/b61e374...72d7634


betatim:
Deployed. Let's see what happens and then decide if we should also bump the r2d prefix (maybe on Saturday morning when traffic is low?)

minrk:
That's a good idea

betatim:
Done.

PR #873, opened 2018-12-18
updating about link

choldgraf:
the about page has a redirect problem so this just makes the link point directly to the right page


consideRatio:
Merged and created #874

PR #878, opened 2019-01-12
Add a DNS based ban for a newly discovered mining pool

betatim:
Deals with the pool discovered just now being used by `cnrig` in the jupyterlab repo to mine coins.

I think our [SRE guide on bans](https://github.com/jupyterhub/mybinder.org-deploy/blob/9126f56838e7332929626f5c6dbf264b3f8c1a32/docs/source/command_snippets.md#banning-traffic) is out of date since https://github.com/jupyterhub/mybinder.org-deploy/pull/740. Something for a new PR.


betatim:
Related question: can we ban `*.example.com`? I think we could safely ban the root domain in the PR and all its subdomains.

choldgraf:
I'm generally +1 on this if we know it's a bad actor and it's not something like `*.google.com`.

Issue #879, opened 2019-01-14
Improve the SRE guide to include common patterns / tips to spot miners


choldgraf:
We [recently had another miner incident](https://gitter.im/jupyterhub/mybinder.org-deploy?at=5c3b61ad39596c372bd62783). As mybinder.org grows in popularity, this will only become more common.

Could we add a page to the SRE guide that describes how to identify and deal with mining behavior? Perhaps somebody that has dealt with this already could sit down w/ me and we can hash out an outline that I can turn into a docs PR?

PR #880, opened 2019-01-14
[WIP] Add script to list pods with high CPU processes

betatim:
This script comes from https://gist.github.com/minrk/325699d85ba904000b125f4de706353c
with some small modifications.

Use this to find potential miners or other abusive pods.

Making a start on #879

### DIFF
--- a/scripts/expensive-processes.py
+++ b/scripts/expensive-processes.py
@@ -8,7 +8,7 @@ from operator import attrgetter
 import os
 import pipes
 import re
-from subprocess import check_output
+from subprocess import check_output, run
 import sys
 
 import kubernetes.client
@@ -52,7 +52,13 @@ def ssh(instance, ssh_cmd):
 def get_instances():
     """Get the instance names"""
     instance_json = gcloud("compute", "instances", "list")
-    return [i["name"] for i in instance_json if '6h5b' not in i['name']]
+    instances = []
+    for i in instance_json:
+        name = i["name"]
+        if any(x in name for x in ('g47m', '6h5b')):
+            continue
+        instances.append(name)
+    return instances
 
 
 def parse_t(timestring):
@@ -189,6 +195,12 @@ def suspicious_cmd(cmd):
         return True
 
     if "socat" in cmd:
+        return True
+
+    if "onion" in cmd:
+        return True
+
+    if "base64" in cmd:
         return True
 
     if "./python" in cmd:
--- a/scripts/expensive-processes.py
+++ b/scripts/expensive-processes.py
@@ -188,6 +188,9 @@ def suspicious_cmd(cmd):
     if "xmrig" in cmd:
         return True
 
+    if "socat" in cmd:
+        return True
+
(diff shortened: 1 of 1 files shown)


scripts/expensive-processes.py
```diff
@@ -0,0 +1,253 @@
...
+import pipes
+import re
+from subprocess import check_output
+import sys
+
+import kubernetes.client
+import kubernetes.config
+
+project = "binder-prod"
+kube_ctx = "prod"
```
betatim: I removed the kube context bits because for me the context had a different name. Is there a way we can make this work across systems? I think this would reduce the chances of executing a command in the wrong cluster.

Should also be removed.

scripts/expensive-processes.py
```diff
@@ -0,0 +1,253 @@
...
+    """execute a command via ssh on the remote host"""
+    if isinstance(ssh_cmd, list):
+        cmd = " ".join(map(pipes.quote, ssh_cmd))
+    cmd = [
+        "gcloud",
+        f"--project={project}",
+        "compute",
+        "ssh",
+        instance,
+        "--zone=us-central1-a",
```
betatim: Had to add this to make the command work. It it weird because I can run `gcloud --project=binder-prod compute ssh <nodename>` to get a ssh connection. However this command needs the zone specification otherwise there is an error that things aren't well enough specified.

scripts/expensive-processes.py
```diff
@@ -0,0 +1,253 @@
...
+    #     stdout=True,
+    #     stderr=False,
+    #     stdin=False,
+    #     tty=True,
+    # )
+    except Exception as e:
+        return f"Error reporting on {pod.metadata.name}: {e}"
+
+
+async def report_pod(pod, instance):
```
betatim: Also pass in the instance name because it was interesting to see if the suspicious pods were on the same node or not.

choldgraf:
just a note that this is sort of covered in the mybinder-tools repo here: https://github.com/jupyterhub/mybinder-tools/blob/master/metrics.ipynb

(search for the "Pod Resources" section)

scripts/expensive-processes.py
minrk: > Is there a way we can make this work across systems?

Leaving them out and relying the current kubectl configuration is probably the only portable mechanism.

scripts/expensive-processes.py
minrk: might have to do with defaults. This can be set in your gcloud config and can be any value.

minrk:
@choldgraf true! This does a little more, because it doesn't just point to which pods are using cpu (usually normal busy users), it shows what processes inside those pods, which is where we typically find our miners.

But that points to something potentially useful: with the busy-pod implementation in that notebook, maybe we can remove the `gcloud` calls altogether from this script, which should simplify things a bunch.

minrk:
I think we can close this one in favor of minesweeper, which does this continuously (and more)

Issue #892, opened 2019-02-04
Did we change the BinderHub image name generation?


choldgraf:
I just tried to update the BinderHub image that we use on mybinder.org using the `list_new_commits.py` script, but this didn't seem to output the correct image hash. (see #890 and #891). Have we changed the way that image names get generated in BinderHub? Or perhaps there's a bug in `list_new_commits.py` now?

PR #893, opened 2019-02-04
updating r2d and commit script for r2d

choldgraf:

https://github.com/jupyter/repo2docker/compare/ff894f44...9c2c7b6a

also partially addresses #892 though I think we still need to figure out the binderhub/jupyterhub side of that script

### DIFF
--- a/scripts/list_new_commits.py
+++ b/scripts/list_new_commits.py
@@ -38,7 +38,7 @@ for i_repo, i_url in repos.items():
     latest_hash[i_repo] = resp.json()[1]['sha']
 
 url_bhub = 'https://github.com/jupyterhub/binderhub/compare/{}...{}'.format(bhub_live, latest_hash['binderhub'][:7])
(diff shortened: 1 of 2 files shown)


scripts/list_new_commits.py
```diff
@@ -38,7 +38,7 @@
     latest_hash[i_repo] = resp.json()[1]['sha']
 
 url_bhub = 'https://github.com/jupyterhub/binderhub/compare/{}...{}'.format(bhub_live, latest_hash['binderhub'][:7])
-url_r2d = 'https://github.com/jupyter/repo2docker/compare/{}...{}'.format(r2d_live, r2d_master[:7])
+url_r2d = 'https://github.com/jupyter/repo2docker/compare/{}...{}'.format(r2d_live, r2d_master[:8])
```
choldgraf: it pulls 8 characters not 7

choldgraf:
@rgbkrk I'm giving this another shot! I think that it may have been because we were using one-too-few characters for the r2d image name 🤞

PR #895, opened 2019-02-10
Bump repo2docker to fix conda bug

betatim:
This updates repo2docker to a version that correctly pins conda. We previously weren't pinning it properly and as a result got hit by a recently introduced bug in conda

https://github.com/jupyter/repo2docker/compare/9c2c7b6a...9766c954


betatim:
In particular https://github.com/jupyter/repo2docker/pull/576

PR #902, opened 2019-02-20
incident report for today

minrk:
`kubectl logs` stopped working due to a problem

This is

1. a likely bug in GKE, where network info is lost for a node
2. a likely bug in kubernetes master (not sure what component)
   where one node with messed up network status prevents
   ssh tunnels from being created to any node

Reported [to GKE](https://issuetracker.google.com/issues/124813114).

The root cause remains unknown, but we learned a bit about how to track down the problem:

1. `kubectl get node -o json | jq '.items[] | [.status.addresses[]]'` gives suspicious output for the sad node (at least this time)
2. you can get apiserver logs even for a managed cluster with

   ```bash
   kubectl proxy
   curl -O http://localhost:8001/logs/kube-apiserver.log
   ```


choldgraf:
nice! thanks for the writeup :-)

so it sounds like as a general rule, when the "no ssh tunnel" issue pops up, the solution is to `describe` to find the node w/ a wonky IP situation and/or start recycling nodes, yeah?

docs/source/incident-reports/2019-02-20-no-logs.md
```diff
@@ -0,0 +1,162 @@
+# incident date: 2019-02-20, kubectl logs unavailable
+
+## Summary
+
+Quick summary of what user impact was, how long it was, and how we fixed it.
```
betatim:
```suggestion
For new builds users did not see the build logs. Builds did happen normally though. Fixed by removing node that had no IP address in its metadata.
```

docs/source/incident_reporting.rst
```diff
@@ -26,12 +26,13 @@ Incident history
    :maxdepth: 1
    :caption: Incidents
 
+   incident-reports/2019-02-20-no-logs
```
betatim:
```suggestion
   incident-reports/2019-02-20-no-logs.md
```

betatim:
Good to merge after the two suggestions are decided on.

Issue #885, opened 2019-01-24
Cordoning node with ingress controller causes outage


minrk:
According to https://github.com/kubernetes/kubernetes/issues/65013, load-balancers exclude cordoned nodes from targets because a cordoned node is presumed to be cordoned as a precursor to removal, rather than a temporary condition. When we clean images, we cordon the cleaning node because new launches tend to fail while image cleaning is occurring due to very high disk load. Together, this means that image-cleaning on the node with the ingress-controller causes an outage as the load-balancer ignores the node with the ingress-controller as a valid routing target.

The quickest solution is to move the ingress controllers to the core nodes, which don't tend to get cordoned.

A second and probably better solution that's perhaps more appropriate is to use a different node taint that `kubectl cordon` to prevent user pods and builds from being assigned and use nodeSelectors/tolerances to accomplish this.

minrk:
This should also be covered in the SRE for node pool upgrades, because we need to be careful to not leave cordoned nodes with the ingress controller pods during upgrade.

betatim:
We could also add affinity rules to the controllers so they don't both end up on the same node.

betatim:
Closing as we now have the controllers setup with pod anti affinity

PR #904, opened 2019-02-23
Add anti-affinity rules for nginx-ingress

betatim:
This is a proposal for fixing some of #885.

Using anti-affinity rules we can organise for the nginx-ingress
controller pods to be spread over multiple nodes. This prevents an
outage from happening when that particular node gets cordoned.
nginx-ingress controllers are not selected when they are on a node
that is cordoned.

It is only a partial fix as we'd still end up with an outage if two nodes get cordoned at the same time and they both contain ingress controllers. I think spreading the pods across nodes is a good idea independent of fixing the cordoning problem fully.

I've not tested this locally because I only have one node in my minikube. This means we should deploy this when we have people around to check that it actually works.

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -145,10 +145,8 @@ prometheus:
 
 nginx-ingress:
   controller:
-    nodeSelector: *coreNodeSelector
     service:
       loadBalancerIP: 35.202.202.188
-    replicaCount: 2
 
 static:
   ingress:
--- a/config/staging.yaml
+++ b/config/staging.yaml
@@ -63,7 +63,6 @@ nginx-ingress:
   controller:
     service:
       loadBalancerIP: 104.197.11.66
-    replicaCount: 2
 
 static:
   ingress:
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -244,7 +244,23 @@ nginx-ingress:
         prometheus.io/scrape: 'true'
         prometheus.io/port: '10254'
   controller:
-    replicaCount: 5
+    affinity:
+      podAntiAffinity:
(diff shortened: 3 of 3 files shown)


config/prod.yaml
```diff
@@ -145,10 +145,8 @@ prometheus:
 
 nginx-ingress:
   controller:
-    nodeSelector: *coreNodeSelector
```
consideRatio: Note that we can have both a nodeSelector and an podAntiAffinity rule, if we dont have the coreNodeSelector I assume this pod can end up on all kinds of nodes, which I guess could be fine if it is allowed to be disrupted.

consideRatio: How many core nodes does mybinder.org run with?

betatim: Good point about disruption. Without it a node running one of these pods would never get "scaled away" right? I will check what needs doing.

We only have one core node which is why I removed this selector. It was added as a workaround for both controllers running on the same non-core node that then got cordoned for image clean up purposes which then lead to an outage for mybinder.org because both controllers were unavailable (controllers on cordoned nodes are treated as unavailable and don't get any traffic).

betatim: The default value for `controller.minAvailable` is 1 in the [nginx-ingress chart](https://github.com/helm/charts/tree/master/stable/nginx-ingress#configuration) and we have two pods. I think this means that one pod can be moved at a time.

consideRatio: This will work out good, and better than before, i think!

betatim:
I'll deploy this tomorrow (Tuesday) morning EU time and see what happens.

betatim:
Deployed and the two controllers are running on different nodes now.

PR #908, opened 2019-03-15
Decrease total number of pods per node

betatim:
This increases the memory guaranteed to each user pod as a way to reduce
the number of pods per node.

The motivation for doing this is to see if we can lower the load average per node a bit to make things fell "snappier". I think it is easy to compute how much this will cost us extra but I have no idea how much we'd have to reduce the load to get that "snappier" feeling. hence just trying some values.

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -42,7 +42,7 @@ binderhub:
     singleuser:
       nodeSelector: *userNodeSelector
       memory:
-        guarantee: 450M
+        guarantee: 550M
         limit: 2G
       cpu:
         guarantee: 0.01


choldgraf:
shall we just merge this and see what happens? :-)

betatim:
Sounds good to me. Mostly was looking for someone else to offer an opinion so merge away.

betatim:
https://grafana.mybinder.org/d/nDQPwi7mk/node-activity?refresh=1m&panelId=26&fullscreen&orgId=1&from=1552718023934&to=1552804423934

We are now running at about 66 instead of 95 user pods per node. This means we will have about 30% more nodes at any given time. Let's keep an eye on if performance "feels better" so we can decide if we want to keep it or not.

The [Load average](https://grafana.mybinder.org/d/nDQPwi7mk/node-activity?refresh=1m&panelId=38&orgId=1&from=1552718023934&to=1552804423934&fullscreen) plot isn't super obviously showing "less load since the deploy" (the yellow line).

To see how much more this is costing us we keep an eye on https://grafana.mybinder.org/d/nDQPwi7mk/node-activity?refresh=1m&panelId=40&orgId=1&from=now-30d&to=now&fullscreen

PR #917, opened 2019-03-25
Add incident report for repo2docker upgrade and build cache wipe

betatim:
Incident report for yesterday's repo2docker upgrade bonanza.


betatim:
Need to reflect on our todo's from this that should be listed at the end of this report before merging.

choldgraf:
I'm curious what it meant to add two new nodes, but as a new "node pool" instead of just adding nodes to the current setup. Was there a specific reason for that?

betatim:
The node pool we have is set to auto-scale. This means even if you manually increase the minimum size of the cluster the auto-scaler won't actually deploy those nodes unless it also thinks they are needed. This means if your cluster is "empty" as determined by the auto-scaler it won't grow the cluster.

In this case I wanted to have extra nodes that could help out with builds.

betatim:
Merged this now as I won't get around to seriously thinking about other action items.

PR #924, opened 2019-04-03
[MRG] Add report on node pool upgrade incident

betatim:
Report on today's incident during the node pool upgrade.


choldgraf:
the action item for docs is now also merged in #923 , thanks @betatim

betatim:
Unfortunately #923 only contains the hint in the "staging" cluster part of the docs. Reflecting that I still remembered this when I worked on staging but had forgotten by the time I cordoned the last few nodes on production (after lunch...). So the action item is still open :-(

choldgraf:
well we can merge this one in either way!

PR #937, opened 2019-04-09
operators team info and slight reorg

choldgraf:
This is a slight refactor of the SRE docs, and adds more mention of the operator team.


docs/source/index.rst
```diff
@@ -1,45 +1,58 @@
 Site Reliability Guide for mybinder.org
 =======================================
 
-Introduction
-------------
+What is the mybinder.org operations team?
+-----------------------------------------
 
-This site is a collection of wisdom, tools, and other helpful information to assist in
```
betatim: I would keep this sentence before the heading (L4) to give people who land here an idea where they are. Even before I care what the ops team is I care about "where the heck am I here and why is there an ops team?"

docs/source/index.rst
```diff
@@ -1,45 +1,58 @@
...
 
-If you're looking for the Binder docs, `see this link <https://docs.mybinder.org>`_. If you're looking
-for information on deploying your own BinderHub, `see this link <https://binderhub.readthedocs.io>`_.
+This site is a collection of wisdom, tools, and other helpful information that the
+mybinder.org operations team uses for maintenance and team-processes
+around the BinderHub deployment at `mybinder.org <https://mybinder.org>`_.
+
+**If you're interested in helping the mybinder.org operations team**, first
+check out `this post on what an operator does <https://discourse.jupyter.org/t/the-operators-no-binder-isnt-forming-a-rock-band/694>`_.
+If that still sounds interesting to you, please reach out to the operations team
```
betatim: I wouldn't use "if this still sounds interesting...", how about "To show your interest please post [here](link to specific thread) or stop by the [operations gitter channel](link here) and introduce yourself." or something?

docs/source/index.rst
```diff
@@ -1,45 +1,58 @@
...
+via the `Jupyter Discourse <https://discourse.jupyter.org>`_.
+
+Note that if you're not interested in information about the mybinder.org BinderHub deployment,
+you may find more useful information elsewhere. If you're looking for the
+Binder docs, `see this link <https://docs.mybinder.org>`_. If you're looking
+for information on deploying your own BinderHub
+`see this link <https://binderhub.readthedocs.io>`_.
```
betatim: I'd remove the first sentence, the start of the other sentences makes it clear that this is about "other places for information" IMHO. Should we move this to the top before the "what is the ops team" heading? That way people who are in the wrong place get redirected early.

docs/source/index.rst
```diff
@@ -1,45 +1,58 @@
...
+mybinder.org operations team uses for maintenance and team-processes
+around the BinderHub deployment at `mybinder.org <https://mybinder.org>`_.
+
+**If you're interested in helping the mybinder.org operations team**, first
+check out `this post on what an operator does <https://discourse.jupyter.org/t/the-operators-no-binder-isnt-forming-a-rock-band/694>`_.
+If that still sounds interesting to you, please reach out to the operations team
+via the `Jupyter Discourse <https://discourse.jupyter.org>`_.
+
+Note that if you're not interested in information about the mybinder.org BinderHub deployment,
+you may find more useful information elsewhere. If you're looking for the
```
betatim:
```suggestion
you may find more useful information elsewhere. If you are looking for the
```
I prefer no contractions in written English ... just personal taste

docs/source/index.rst
```diff
@@ -1,45 +1,58 @@
...
+Binder docs, `see this link <https://docs.mybinder.org>`_. If you're looking
```
betatim:
```suggestion
documentation on how to use mybinder.org, `see the user facing documentation <https://docs.mybinder.org>`_. If you're looking
```
"Binder" is an overloaded term so I tried to use something else. If you accept this change we should also remove the `the` on the previous line (I think).

docs/source/deployment/how.md
```diff
@@ -11,6 +11,18 @@ change deployed to mybinder.org.
 The first two sections of this page cover how to upgrade either `repo2docker` or
 `BinderHub`.
 
+## Deployment policy
+
+Deployments to mybinder.org should be:
```
betatim: Should we add a point 0: "**Timely** We deploy changes to repo2docker and BinderHub within a few days of them being merged into master"?

betatim:
I like it. Especially the points that describe the deployment policy.

I was wondering if we should formulate them as aspirational policies ("deployments should be...") or if we should change it to "Our guiding principles for deployments to mybinder.org are:". Then add a sentence at the end saying something like "Achieving this state will require continuous improvement of our tools and processes."

Actually as I tried to write the above paragraph I struggled to find a good formulation for what I had in my mind. Maybe we leave it as is then?

docs/source/index.rst
choldgraf: ok, added in a new thread to Discourse and update the language here, LMKWYT

choldgraf:
@betatim addressed! (I think) let me know how that looks to you

PR #940, opened 2019-04-10
Temporarily ban the RasaHQ repo

betatim:
This repo fails to build and receives a lot of launch requests. Banning
it till the repo is fixed upstream.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -46,6 +46,7 @@ binderhub:
         # e.g. '^org/repo.*'
         - ^ines/spacy-binder.*
         - ^soft4voip/rak.*
+        - ^RasaHQ/rasa_core.*
     BinderHub:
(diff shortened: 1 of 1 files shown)


choldgraf:
how will they know if their repository builds if it's banned?

betatim:
Using repo2docker locally?

I posted an issue on [their new repo](https://github.com/RasaHQ/rasa_nlu/issues/3134) about two weeks ago and progress has been slow. Meanwhile they've been using up quite a bit of CPU time building and rebuilding a failing repo. At least that is my conclusion from staring at the CPU load charts on grafana for a bit and trying to work out what causes the very large loads. RasaHQ isn't the only cause for them but they seem to cause a good chunk of it.

choldgraf:
cool - as long as they know about the existence of this problem then I think it's fine

PR #944, opened 2019-04-14
update list_new_commits to change source of binderhub and repo2docker SHA

henchc:
We can't reliably get helm chart updates from the binderhub repo commits, this checks the helm chart yaml instead.

Also looks at dockerhub for repo2docker instead of repo.

### DIFF
--- a/scripts/list_new_commits.py
+++ b/scripts/list_new_commits.py
@@ -22,14 +22,14 @@ r2d_live = helm_chart['binderhub']['config']['BinderHub']['build_image'].split('
 
 print('Fetching latest commit SHA for BinderHub and repo2docker...')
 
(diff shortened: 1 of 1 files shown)


betatim:
The problem is that not every commit in the BidnerHub or repo2docker repo leads to a new chart/image. In particular the binderhub helm chart doesn't get built for merge commits nor for commits that don't change the code.

For repo2docker we should be building an image "all the time". However it is worth checking if there is actually an image available on docker hub because that is what will cause a deploy to fail. For mega bonus points (new PR maybe) we could have the script give a warning when there is a newer commit in the r2d repo and no image on docker hub.

choldgraf:
thanks for the patch @henchc !

PR #963, opened 2019-05-08
analytics-publisher: Bump jinja2 version

yuvipanda:
Fix for a sandbox security escape (https://nvd.nist.gov/vuln/detail/CVE-2019-10906)
in jinja2. We are not affected by this particular vulnerability
but is good to upgrade regardless.


choldgraf:
Any reason this is now >= instead of ==? Could a jinja 2 update create some kinda confusing result if it isn't pinned? I guess we're only using this jinja within any person's environment so it wouldn't break reproducibility

yuvipanda:
@choldgraf good call. Done!

PR #970, opened 2019-05-21
Adding new configuration for OVH cluster

mael-le-gal:
The aim of this PR is to provide the configurations for the `OVH K8s deployments`.

In this PR I made the enabling of `analytics-publisher` and `gcs-proxy` available because I don't need these 2 in the ovh deployments.
I also hid all the `ovh specific secrets` for now until someone send me the encryption key.

The instance of the binder stack resulting of this configuration is available here : https://binder.mybinder.ovh/

This a **[WIP]**, don't hesitate to tell me if I did something wrong.

I still need to fix the following points :  

- The image cacheing doesn't seem to work. It re-builds every time.
- I need to deploy the same dashboards in this grafana instance than yours here https://grafana.mybinder.org


meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/caching-problem-on-ovh-k8s-cluster-deployment/1151/1

betatim:
Nice work!

@mael-le-gal can you email me at betatim@gmail.com and I will reply with the ssh-vault'ed secret for git-crypt.

@yuvipanda do you have an idea how we can collect events from both instance in one place or will we have to write a new analytics publisher for OVH or ...? I think we need to keep the ability to collect the launch events as this is (I think) our source of truth for how much mybinder.org is used, what repos get launched, etc. It is an interesting data stream and I think we need to dig out these kinds of stats for bragging and reporting on people who give us money etc.

config/binder-ovh.yaml
```diff
@@ -0,0 +1,156 @@
+projectName: binder-ovh
```
betatim: can we call this file `config/ovh.yaml` instead of `config/binder-ovh.yaml`? Not a big deal but it would please my OCD :)

betatim:
Do you want to add to the deploy script that runs on PR merge and deploys things as well? I had a look at our setup and these are some things that we should have on the radar:

1. we need to build and push to a public registry [here](https://github.com/jupyterhub/mybinder.org-deploy/blob/cbc66897db5f12883770192ab02a3dfa0cb1040e/.travis.yml#L64-L69) or push also to a registry accessible from inside the OVH cluster. I think we should continue to keep the mybinder.org docker registry private to avoid abuse and large network traffic bills

1. our [`deploy.py`](https://github.com/jupyterhub/mybinder.org-deploy/blob/master/deploy.py) needs to learn to load the keys needed to auth to the OVH cluster when doing a deploy there

1. we need to add [these two steps for OVH](https://github.com/jupyterhub/mybinder.org-deploy/blob/cbc66897db5f12883770192ab02a3dfa0cb1040e/.travis.yml#L87-L92)

config/binder-ovh.yaml
mael-le-gal: yes of course, I'll  change that

mael-le-gal:
### I just pushed a new version of the PR, here are the updates : 

- `secrets/config/ovh.yaml` and `secrets/ovh-kubeconfig.yml` are now encrypted with your `git-crypt` key and contain all needed secrets for the `ovh stack`

- In the documentation I add myself in the list of user who know the `git-crypt key`

- I fix the image caching problem, it should now stop rebuilding at each time. The resolution is explained [here](https://discourse.jupyter.org/t/caching-problem-on-ovh-k8s-cluster-deployment/1151)

- As you suggest @betatim I update the `deploy.py` script. You can know launch a deploy on the `ovh stack` simply by running : `python deploy.py ovh binder-ovh`.

- I also change the `travis.yml` file so that your CDCI worflow will now automatically deploy on the `ovh stack` once the deployment of `gcp stack` is done
 
- I copy/paste all your grafanas dashboard so that we have the same in the `ovh stack`

### Summary of OVH Stack access points : 

- binder : https://binder.mybinder.ovh
- hub : https://hub-binder.mybinder.ovh
- grafana : https://grafana-mybinder.mybinder.ovh
- redirected : https://playground-mybinder.mybinder.ovh -> https://play.nteract.io
- redirected : https://docs-mybinder.mybinder.ovh -> https://mybinder.readthedocs.io

### Remaining point 1 : 

> we need to build and push to a public registry here or push also to a registry accessible from inside the OVH cluster. I think we should continue to keep the mybinder.org docker registry private to avoid abuse and large network traffic bills

If I understand correctly the `chartpress.yaml`, the step you describe is related to `analyticsPublisher` and that part is not yet used in the `ovh stack`, is it ?

### Remaining point 2 : 

I should do a last check on all allocated resources for pods (limits/requested) because for now it is only default one.

Is there some specifics stuff that we need to be aware of ? For example pods that usually consume most ressources ? pods that need to be spawned on separated nodes ?

### Remaining point 3 : 

We should discuss the supervision process. As we are providing the k8s cluster it seems normal that we will supervise it and fix it when broken. However as you will have the control on deployments and on the source code of `binder` I suppose this last point can't be a part of our responsabililty.

It's an open question and I only provide my own point of view, maybe a further meeting whith all the related people would be great to talk about this.

betatim:
re: point 1: you are right. I hadn't looked inside the chartpress config in this repo.

point 2: all the pods deployed by this helm chart should get the same limits as on the GKE production deployment no? For the GKE deployment we operate two node pools: one for user spawned pods and one for the core services. This is controlled by the node selector. You removed that from the OVH config. I think for getting started this should be fine. We can look at that in a new PR. The main reason for doing this is to make it possible for the cluster auto scaler to scale down. Otherwise you end up with pods that can't be disrupted on a node and the auto scaler won't remove that node.

point 3: can we make it so that (part of) us can also log in to the dashboard/UI that controls the cluster? Then we can help with that. If we can work out the commands ([like here](https://mybinder-sre.readthedocs.io/en/latest/command_snippets.html#cluster-management)) for k8s updates, adding/removing node pools/node upgrades we can also help with that. I think the most frequent "maintenance" task at the moment is `kubectl cordon <nodename>` as we seem to get "broken" nodes sometimes. Again, if we know how to do this someone from the team can do it and we can share the workload. (Generally any and all help maintaining the cluster(s) is welcome :) ). If you want to I'd be happy to add your Google account to the GKE cluster so you can help out there if you want to?

betatim:
Given point 3 is about admin stuff: should we merge this and see what happens?

I am excited to see it all in action (or maybe fail in some way we didn't expect :) )

mael-le-gal:
> point 2: all the pods deployed by this helm chart should get the same limits as on the GKE production deployment no? For the GKE deployment we operate two node pools: one for user spawned pods and one for the core services. This is controlled by the node selector. You removed that from the OVH config. I think for getting started this should be fine. We can look at that in a new PR. The main reason for doing this is to make it possible for the cluster auto scaler to scale down. Otherwise you end up with pods that can't be disrupted on a node and the auto scaler won't remove that node.

Thanks for these precisions, I saw the `coreNodeSelector` and `userNodeSelector` in the production configuration but I choose to ignore them at first. Now I understand better. I will do a 2nd PR in the future for using the same resource configuration than your current production.

> point 3: can we make it so that (part of) us can also log in to the dashboard/UI that controls the cluster? Then we can help with that. If we can work out the commands (like here) for k8s updates, adding/removing node pools/node upgrades we can also help with that. I think the most frequent "maintenance" task at the moment is kubectl cordon <nodename> as we seem to get "broken" nodes sometimes. Again, if we know how to do this someone from the team can do it and we can share the workload. (Generally any and all help maintaining the cluster(s) is welcome :) ).

By default you are admin on the k8s cluster with the `kubectl` command and the `secrets/ovh-kubeconfig.yml` so you can do everything you are able to do with `kubectl`. There is no k8s Dashboard/UI installed by default but you can easilly install one with [a helm chart ](https://github.com/helm/charts/tree/master/stable/kubernetes-dashboard) if some of you prefer a Web UI than a client. 
If you'll need to add/remove nodes in the k8s cluster or administrate it you'll need our help in a first place. But we will be happy to work with you getting an auto-scale cluster depending on your load. That last point will need some work from our side, but it feasible I guess.

> If you want to I'd be happy to add your Google account to the GKE cluster so you can help out there if you want to?

I'll send you my Google account by mail. I can't assure you that I'll have plenty of time but from time to time i'll be happy to help :)

> Given point 3 is about admin stuff: should we merge this and see what happens? I am excited to see it all in action (or maybe fail in some way we didn't expect :) )

Yes I think we can merge this first PR. Is there a URL to your travis workflow so I can see the advancement of deployment and investigate if failure ?

choldgraf:
This is all very exciting :-)

@mael-le-gal re: keeping track of what's happening at deployment, could you check out this page: https://mybinder-sre.readthedocs.io/en/latest/deployment/how.html#deployment-policy which has some ideas for how to deploy an upgrade and how to watch progress?

A few questions

* These two deployments are using separate registries, right? So if somebody builds an image on one cluster, it will still need to be re-built on another? (https://github.com/jupyterhub/mybinder.org-deploy/pull/970/files#diff-44ac241a10470437b10db3c789f48695R14)
* This will deploy to GCP+OVH in serial, one after another, yes? (https://github.com/jupyterhub/mybinder.org-deploy/pull/970/files#diff-354f30a63fb0907d4ad57269548329e3R97). I think that's fine for now but want to flag this for future discussion because it won't scale well if/when we add more clusters.

Finally, just a note that we should really write up a blog post about this once the cluster is merged! I'm happy to help with that once the time is right.

betatim:
@mael-le-gal all the questions/answers are fine with me. We don't use the k8s dashboard (at least I don't know of someone who does) so that is fine as is. I was indeed talking about the "dashboard" that allows us to update k8s versions, add nodes, etc. Also fine if we do that via you for the moment.

Currently the CI only runs when the PR is merged. The best way to find the output and status of the jobs is to click the badge next to the word "deployment" [here](https://github.com/jupyterhub/mybinder.org-deploy/#mybinderorg-deploy). (One day it would be cool to run some tests already on the PR but as always: "so many things to do").

> These two deployments are using separate registries, right? 

Correct.I donn't even know if we can share the GKE registry only with certain other external IPs. We'd also have to pay for the egress traffic from the Google cloud, which is usually some of the most expensive traffic. I think because these two clusters will always run the same version of bhub and repo2docker this is a small problem.

> This will deploy to GCP+OVH in serial, one after another, yes?

Correct. Making it run in parallel would be nice, for debugging at the start I think it will be easier to not have the log output intermingled.

betatim:
To double check: right now the cluster is not setup to auto scale correct? Wanted to check before sending (lots of) traffic to it.

mael-le-gal:
It look like the travis workflow was a success ? 👍 

> To double check: right now the cluster is not setup to auto scale correct? Wanted to check before sending (lots of) traffic to it.

No the OVH cluster is not set to auto scale for now. I'm working for adding more nodes by default. (currently you have only one ... that's few)

minrk:
We should probably add some 'capacity' api to binder, so we can easily see how many users a deployment currently has, and how many it claims to support. We should probably also explore failure modes of attempting to deploy on a 'full' cluster. Right now, I think this results in 'stuck' spawns as pods fail to be scheduled and eventually time out at the hub level. On one level, this is the right thing to do because if space frees up, the pod will launch. But it's not very informative for the user who has no idea that the deployment is full and they need to try somewhere else or another time.

baaastijn:
Hi there, i'm product manager at OVH, in Mael team.

I'd love to help on this topic, especially on 3 points : 

1. Sizing correctly our K8s cluster, by knowing the expected volumetry (amount of containers to run, size of containers, etc)
2. horizontal scaling for the K8s cluster. We don't have auto scaling but i think that we can automate something easily together, since we provide metrics for nodes.
3. Communication about it. We share the same values and wewill write a blog post about Jupyter, their community, helm + jupyterhub, etc. You may also have ideas or ongoing actions about it.

Tell me if we continue the discussion here or discourse or whatever. Thanks folks !

betatim:
Hey @baaastijn!

**Cluster size:** we currently operate a cluster that auto-scales so there is no functionality in BinderHub (the software that is running) to deal with "the cluster is full". There is an issue for that https://github.com/jupyterhub/binderhub/issues/861 (contributions welcome).

The current cluster behind mybinder.org scales between 4 and 7 nodes during a week. Each node has 8 vCPUs and 52 GB memory. There is one additional node that is always on for core services (4 vCPUs, 26 GB memory). However if the cluster doesn't auto-scale I think you don't need the dedicated node. We have it because it helps with the auto-scaler reliably scaling down the cluster.

In terms of cluster sizing I'd go with a setup that is enough to handle 5-10% of mybinder.org's traffic (so 10% of the resources, about one or two nodes??). Then we run that setup for a bit and see how it goes. After that either auto-scaling is working or we size the cluster based on what we learnt so that it can handle 50%(??) of the traffic.

The traffic distributor we plan to use is already up at https://my.bndr.it. Right now it uses a weighting of 4:1 between mybinder.org and binder.mybinder.ovh. You can try it via https://my.bndr.it/v2/gh/binder-examples/requirements/master (it has a 10s memory so you need to wait a moment between clicks to get a new roll of the random number generator). The plan is to move this to mybinder.org once we are confident that everything works.

We run around 300-400 pods at any moment. The resources are set [here](https://github.com/jupyterhub/mybinder.org-deploy/blob/43c3dda1b801d8a6b4c5526c66742ab69622679b/config/prod.yaml#L44-L49). We heavily oversubscribe our nodes, because most of the time a notebook just idles.

What do you think of this as a plan?

**Communication about it.** We are very interested in writing a blog post to announce that mybinder.org is now backed by two clusters and that OVH is the first additional member in the mybinder.org federation. Do we want to write one post and then post it to the OVH and Jupyter blogs? Or two separate blog posts? I think I am Ok with either or the option that will lead to more exposure/publicity.

baaastijn:
Thanks @betatim 

Let's go for 2 nodes for a start then :)
ACK for the autoscaling part

For the blog post, i'd prefer to create 2 separate contents who will achieve 2 differents goals

1. On OVH side, we will explain that we are proud to help you,, that it runs on OVH K8s (and you can try to create a hub with helm), and that OVH has a portfolio of products for data lovers. we will also explain what is jupyter for sure.
2. on Jupyter side i guess you will be more focused on explaining that you now add members to mybinder, starting with OVH. If you need additional info about us (the open source project we help, the legislation we have or data product we propose) don't hesitate !

betatim:
Sounds good.

PR #974, opened 2019-05-23
Ban shishirchoudharygic/mltraining

betatim:
shishirchoudharygic/mltraining is launching and launching and launching new pods but never seems to do anything with them. I posted on their repo earlier today shishirchoudharygic/mltraining#1 asking if they knew what is happening. Banning for now to see if that provokes a reaction.

My assumption is that it is some browser/script/JS/event stream thing gone awry.


choldgraf:
hmm you don't suppose this is related to the other repo that was wonky as well? The one that was unclear what exactly it was for?

PR #980, opened 2019-05-24
make copyright year dynamic

pdurbin:
Instead of 2017 (two years ago), let's make the year dynamic.

I didn't test this but it worked at https://github.com/IQSS/dataverse/commit/d8f1b98a6776c25b79312e44f1e6ea7baa24a1e0 from what I understand. 😄

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -60,7 +60,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'Site Reliability Guide for mybinder.org'
-copyright = '%d, Binder Team' % datetime.now().year
+copyright = '2017 - %d, Binder Team' % datetime.now().year
 author = 'Binder Team'
 
 # The version info for the project you're documenting, acts as replacement for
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -59,7 +60,7 @@ def setup(app):
 
 # General information about the project.
 project = 'Site Reliability Guide for mybinder.org'
-copyright = '2017, Binder Team'
+copyright = '%d, Binder Team' % datetime.now().year
```
betatim:
```suggestion
copyright = '2017 - %d, Binder Team' % datetime.now().year
```

IANAL. This way we preserve the starting date which is nice

choldgraf: I tried committing this but it wouldn't let me :-/

choldgraf:
@pdurbin I believe that we don't have permission to push to your branch, can you decide to accept @betatim's suggestion above (or not) so we can get this in? :-)

pdurbin:
@choldgraf done! I added 469fb34 by @betatim into my branch. Thanks for the heads up.

choldgraf:
wooo, thanks!

PR #988, opened 2019-05-27
Add README for federation redirector

betatim:
Adds a basic README for the redirector.


choldgraf:
this is great, thanks for adding the docs. One question: is this discoverable in the SRE guide at all?

betatim:
We have to link to it I think, or add content there too.

Issue #999, opened 2019-06-04
Federation redirector setup


betatim:
Here a proposal/plan for how to setup the domain names we have and the redirector.

`mybinder.org` is a A record that points to `35.202.202.188` (current prod cluster IP). The redirector will pick up requests and send them to either `gke.mybinder.org` or `ovh.mybinder.org`.

`ovh.mybinder.org` is a CNAME that points to a hostname representing the prod cluster hosted on OVH.

`gke.mybinder.org` is a CNAME that points to `mybinder.org` as a way to point to `35.202.202.188`.

We will configure both the gke and ovh binderhub to use `mybinder.org` as the host in the generated badges.

I think we already have everything in place for this setup regarding DNS entries. What is missing is a switch of which `Host` BinderHub answers on and which host the redirector answers on.

I think it would be good to practice that switch using our staging cluster and I will try that now. Setting up staging.mybinder.org to be served by the federation-redirector (or F'ederick, Frederick with a silent R obviously is what you get from making one word out of federation-redirector ... ;) ) and sending traffic to gke.staging.mybinder.org and ovh.mybinder.org (to have a second host).

betatim:
I've made a few PRs and changes. We now have a redirector setup for staging. This means if you visit https://staging.mybinder.org you will be redirected to one of our staging "clusters". We only have one so we simulate this by having two hostnames `gke.staging.mybinder.org` and `gke2.staging.mybinder.org`.

The switch over from direct access to BinderHub and going via the proxy worked well. I was worried there'd be issues with the certificates but none materialised. I also found a few bugs in the redirector in the process that are now fixed. One good thing was that this didn't actually require any DNS changes (yay for wildcards!).

Overall I think we are in a good place. I think we should  attempt to swap in the federation redirector for prod in the morning European time. Maybe at first without sending traffic to OVH but just the GKE cluster. The point of this is to see what the redirector does when it is hit with a real amount of traffic. If that goes send about 5-10% of the traffic to OVH.

Two things that need pondering:

We should check first: can we recover staging to the "no redirector" setup without causing downtime in case something weird happens when we activate the redirector.

Should we publish a blog post explaining why people are being sent to ovh.mybinder.org/gke.mybinder.org now at the same time as enabling this? Or can we rely on the fact that most people won't notice for a few days so we can test this without having to have a blog post ready?

choldgraf:
This is awesome, thanks so much for everybody's hard work!

[...]

PR #1005, opened 2019-06-05
Disable redirector on staging

betatim:
This PR will disable the redirector on staging. The goal of this exercise is to find out if we can do this without causing a big interruption. Knowing we can back out of the switch to using a redirector will be useful when we enable it on production.

ref: #999


betatim:
This worked. Was running `curl` in a loop every few seconds and requests switched from receiving a 302 to being served the bhub landing page at some point during the deploy. No HTTPS related errors.

PR #1007, opened 2019-06-06
Enable federation for mybinder.org

betatim:
This PR switches mybinder.org to be a federation. Traffic will be sent to the existing GKE cluster and the OVH cluster. The weights are set to 98 and 2 respectively. This means we will send 2% of the traffic to OVH to start with.

I built https://mybinder.org/v2/gh/ipython/ipython-in-depth/master?filepath=binder/Index.ipynb and https://mybinder.org/v2/gh/jupyterlab/jupyterlab-demo/master?urlpath=lab%2Ftree%2Fdemo%2FLorenz.ipynb "by hand" on the OVH cluster so that the images are available. These are the most popular repos this is why I picked them for pre-building.

ref: #999 

To back-out of this we can revert this PR and/or look at #1005 where we did this for staging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,13 +89,13 @@ before_deploy:
   python ./deploy.py prod prod-a
 - |
   # Stage 5, Step 3: Verify production works
-  travis_retry py.test -vx -n 2 --binder-url=https://mybinder.org --hub-url=https://hub.mybinder.org
+  travis_retry py.test -vx -n 2 --binder-url=https://gke.mybinder.org --hub-url=https://hub.gke.mybinder.org
 - |
   # Stage 5, Step 4: Deploy to production on ovh k8s
   python ./deploy.py ovh binder-ovh
 - |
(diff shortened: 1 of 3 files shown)


betatim:
I realised just before pressing merge that if we rename the hub's hostname users will loose connection to their currently running pods. Instead I will add a second hostname to the hub. I didn't rehearse this on staging but I think this will work. Fingers crossed.

betatim:
The binder I launched before pressing merge kept running and being available to me. This means that people using mybinder.org at the time shouldn't have noticed anything.

The deploy script failed at first because we had to wait for the HTTPS certificate so validating that the host was up failed for a short time. This means that maybe a few launches during that time window also failed but should have been retried(?). If we do this again we might want to add the new hostname before to get the certificate first, then send traffic there.

https://grafana-mybinder.mybinder.ovh/d/fZWsQmnmz/pod-activity?orgId=1&from=1559797490030&to=1559800991504 is the grafana for OVH. Can you see when we started sending traffic there?

PR #1010, opened 2019-06-06
Automatically deploy ovh ingress for the 'ovh.mybinder.org' certificate

mael-le-gal:
On the OVH k8s instance we need to automatically deploy a specific ingress for serving the correct certificate if user are coming from the `ovh.mybinder.org` url.

I choose to do this outside the helm deployment because it is very specific to the ovh deployment.


betatim:
Should we write down in the SRE guide why we can't use the "normal" way of getting the certificate so that in a few months we can remember?

Otherwise looks good to me.

mael-le-gal:
You're right, I added a small part about that in the SRE guide.

PR #1020, opened 2019-06-11
Lower traffic fraction sent to OVH

betatim:
The current cluster can't schedule more than 40 user pods so for the
moment we reduce the amount of traffic sent there.

Screen Shot 2019-06-11 at 15 50 17

One small mystery is why there are over 40 pods here in the first place as there are only ~300 pods on GKE and I thought we are sending 10% of the traffic ... :-)

Screen Shot 2019-06-11 at 15 50 24

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -401,9 +401,9 @@ federationRedirect:
   hosts:
     gke:
       url: https://gke.mybinder.org
-      weight: 90
+      weight: 95
       health: https://gke.mybinder.org/versions
     ovh:
       url: https://ovh.mybinder.org
-      weight: 10
+      weight: 5
       health: https://ovh.mybinder.org/versions


choldgraf:
yeah I was wondering the same thing - none of us have credentials to inspect the OVH cluster right? We just need to ping them? We should work out some high-bandwidth channel for mybinder.org operators to talk with each other.

betatim:
What do you mean with inspect?

We can run all the `kubectl` commands but we don't have access to the equivalent of https://console.cloud.google.com/kubernetes/clusters/details/us-central1-a/prod-a?project=binder-prod&tab=details&persistent_volumes_tablesize=50&storage_class_tablesize=50&nodes_tablesize=50 to increase the cluster.

Ideally we'd get a few people access to that so we can reboot nodes, increase the cluster, etc.

PR #1042, opened 2019-06-21
Handle 404s for the federation redirector

### LINKED ISSUE #1041: 404s on mybinder.org return a stack trace
Going to https://mybinder.org/test returns a 404 with the following:

```
  File "/usr/local/lib/python3.7/site-packages/tornado/web.py", line 1676, in _execute
    result = self.prepare()
  File "/usr/local/lib/python3.7/site-packages/tornado/web.py", line 2431, in prepare
    raise HTTPError(self._status_code)
tornado.web.HTTPError: HTTP 404: Not Found
```

jtpio:
Fixes #1041 

In #1041 we were talking about adding a template to properly handle errors by copying it from BinderHub (or even using `importlib.resources`). 

But we can also just add another redirect and let `gke` or `ovh` handle the 404s?

In the end the changes are:

- Set `debug=False`
- Add one more redirect handler

cc @betatim

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -164,6 +164,7 @@ def make_app():
                 {"url": "https://gke.mybinder.org/about", "permanent": True},
             ),
             (r"/", RedirectHandler),
+            (r".*", RedirectHandler),
         ],
         hosts=hosts,
         cookie_secret="get-me-dynamically",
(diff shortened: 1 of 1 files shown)


betatim:
I was  unsure if sending all (undeterminend URI) traffic to one of the hosts inn the federation makes sense. For example should mybinder.org/versions redirect you to one of the hosts or is it better if we make it an error to try and access it? I still don't know what a good answer is so let's merge this and keep thinking, we can always make a second PR :)

PR #1054, opened 2019-06-26
Changing the OVH registry to use

mael-le-gal:
As describe in the issues [here](https://github.com/jupyterhub/mybinder.org-deploy/issues/1016) and [here](https://github.com/jupyterhub/mybinder.org-deploy/issues/1031) we had some problem of reliability with the current docker registry used by the `ovh` stack.

As we just launched a new service of managed docker registry in `beta` it would be a good idea to switch on it and see if it improve anything.

I ordered a registry for you and here is the [URL to the admin interface](https://dlqpwel7.gra5.container-registry.ovh.net/harbor/sign-in) 

This service is powered by the open source project [Harbor](https://github.com/goharbor/harbor) and you can log in with the credentials that you'll find on this PR.

I hope that my configuration is `ok` as I didn't have the chance to test it. ⚠️ 
Maybe I should switch the OVH traffic to 0% also on this PR in order for me to test that everything is OK before switching again to 15%. **What do you think ?**

PR #1058, opened 2019-06-27
Changing the OVH registry to use (second try)

mael-le-gal:
Second try for https://github.com/jupyterhub/mybinder.org-deploy/pull/1054

This time we also added a token url like so :  

```
DockerRegistry:
    token_url: https://dlqpwel7.gra5.container-registry.ovh.net/service/token?service=harbor-registry
```

### DIFF
--- a/config/ovh.yaml
+++ b/config/ovh.yaml
@@ -6,6 +6,8 @@ binderhub:
       hub_url: https://hub-binder.mybinder.ovh
       badge_base_url: https://mybinder.org
       image_prefix: dlqpwel7.gra5.container-registry.ovh.net/binder/r2d-f18835fd-
+    DockerRegistry:
+      token_url: https://dlqpwel7.gra5.container-registry.ovh.net/service/token?service=harbor-registry
   registry:
     url: https://dlqpwel7.gra5.container-registry.ovh.net
     username: cLxohgOONW


config/ovh.yaml
```diff
@@ -5,12 +5,12 @@ binderhub:
     BinderHub:
       hub_url: https://hub-binder.mybinder.ovh
       badge_base_url: https://mybinder.org
-      image_prefix: binder-registry.mybinder.ovh/binder-ovh/r2d-f18835fd-
+      image_prefix: dlqpwel7.gra5.container-registry.ovh.net/binder/r2d-f18835fd-
     DockerRegistry:
-      token_url: https://auth-binder-registry.mybinder.ovh/auth?service=binder-registry
+      token_url: https://dlqpwel7.gra5.container-registry.ovh.net/service/token?service=harbor-registry
```
betatim: Do you think we need the `?service=harbor-registry`?

mael-le-gal: When I was trying to make the first registry work I remember for sure that I had a problem because I didn't specify the `service=binder-registry`.

So I guess that it will be the same this time with `service=harbor-registry`

Moreover, in the documentation example [here](https://binderhub.readthedocs.io/en/latest/setup-binderhub.html) the `service=...` is also given

choldgraf:
Since ovh is at 0% right now, Im gonna give this a merge and see what happens!

betatim:
Looks like this worked. I built and launched an image. When I then tried to launch it again it was fetched from the registry instead of rebuilding!

mael-le-gal:
Everything seems to be ok ✅ 
registry

🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

Issue #1023, opened 2019-06-13
Deploy to all production clusters, then test them


betatim:
Currently we deploy to GKE, then test GKE, then deploy to OVH, then test OVH.

Sometimes the test of GKE fails (for no good reason) and we end up not deploying to OVH.

I think it would be good to first do the two deploys and then test both of them. This way we keep the two clusters more consistent. If someone else also thinks this is a good idea/would make our life easier start a PR :)

This change would be in `.travis.yml` and a good way to learn a bit more about how the automatic deploy setup works.

henchc:
This had been merged! https://github.com/jupyterhub/mybinder.org-deploy/pull/1072

PR #1072, opened 2019-07-04
[MRG] deploy both production clusters then test both

henchc:
Thought I would take the opportunity for what seemed to me like a good entry-PR on the build process here addressing issue #1023

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,11 +88,11 @@ before_deploy:
   # Stage 5, Step 2: Deploy to production
   python ./deploy.py prod prod-a
 - |
-  # Stage 5, Step 3: Verify production works
-  travis_retry py.test -vx -n 2 --binder-url=https://gke.mybinder.org --hub-url=https://hub.gke.mybinder.org
-- |
-  # Stage 5, Step 4: Deploy to production on ovh k8s
+  # Stage 5, Step 3: Deploy to production on ovh k8s
   python ./deploy.py ovh binder-ovh
+- |
+  # Stage 5, Step 4: Verify production works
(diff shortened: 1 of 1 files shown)


betatim:
Looks good to me. Time to merge or do you want to keep working on it ([WIP] in the title)?

henchc:
Great! Just wanted to make sure I didn't miss something :) .

choldgraf:
The change looks good to me! Could you quickly summarize the main benefit of this? Is the goal to minimize the time over out-of-sync-ness between the clusters?

betatim:
Sometimes the test of GKE fails (for no good reason) and we end up not deploying to OVH.

I think it would be good to first do the two deploys and then test both of them. This way we keep the two clusters more consistent. If there is something actually wrong with the setup we will still get a failure that we have to deal with.

Just feels like we get flakey tests that lead to a difference in deployed versions/config more often than an actual test failure on GKE that saves us from also breaking OVH.

And unless we make a catastrophic configuration error things would work well enough for the federation proxy to keep sending traffic to the broken GKE cluster. So we don't get the benefit of "well, let's stop deploying before we hose everything". We should revisit this deploy-deploy-test-test choice when we make the federation redirector smarter.

henchc:
I would +1 this, I've gotten odd timeout failures that may not be the upgrades fault in the first place, and now with the other hub it would deploy one and not the other. I don't think this is a major change for that, but at least you would fully upgrade/revert both together. But I would defer to @betatim given his experience here.

betatim:
Merged. If it starts being annoying or doing the wrong thing we can easily switch back to the old behaviour.

PR #1074, opened 2019-07-08
Adding useful grafana plots to SRE guidelines

sgibson91:
This PR adds a new page to the SRE guidelines which lists some grafana charts chosen by the mybinder.org team. A link and explanation as to what they show as well as some context on when it's interesting to check them is given.

will fix https://github.com/jupyterhub/team-compass/issues/182


minrk:
Great! This is something we can add to over time that will really help folks see what's going on.

PR #1077, opened 2019-07-10
upping quota for jupyter organization repos

choldgraf:
This (I think) bumps the quota for jupyter-specific organizations. Rather than just doing this for JupyterLab, it seemed reasonable to do so for several other jupyter repositories as well...happy to keep it to jupyterlab only if we think that's a good idea....

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -50,6 +50,12 @@ binderhub:
         - ^hmharshit/cn-ait.*
         - ^shishirchoudharygic/mltraining.*
         - ^hmharshit/mltraining.*
+      high_quota_specs:
+        - ^jupyterlab/.*
+        - ^jupyter/.*
+        - ^jupyterhub/.*
+        - ^jupyter-widgets/.*
+
     GitRepoProvider:
       banned_specs:
         - ^https%3A%2F%2Fbitbucket.org%2Fnikiubel%2Fnikiubel.bitbucket.io.git/.*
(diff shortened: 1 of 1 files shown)


betatim:
I think we should pick a limit less than 1000, maybe 200? 1000 is so high that it is equal to having no limit. So maybe 200 and someone who checks once in a while (and then ups the limit) how close the jupyterlab repo is to the limit is the way to go. We should make sure that mybinder.org remains usable by all the other people going about their day to day activities (or workshops, tutorials, etc) during this. I think making some people who want to try out JupyterLab 1.0 wait is better than causing unplanned stress for those who are planning on using mybinder.org for something else.

At the same time as deploying this we should increase the maximum number of nodes in the user node pool. I think currently it is 10 and we routinely use 6 when it is busy. We don't have a good plan/UX for the cluster being out of resources so we should make sure we don't run out :)

mybinder/values.yaml
```diff
@@ -50,13 +50,21 @@ binderhub:
...
+        - ^jupyter-widgets/.*
+
     GitRepoProvider:
       banned_specs:
         - ^https%3A%2F%2Fbitbucket.org%2Fnikiubel%2Fnikiubel.bitbucket.io.git/.*
     BinderHub:
       use_registry: true
       build_image: jupyter/repo2docker:b16b6101
       per_repo_quota: 100
+      per_repo_quota_higher: 1000
```
choldgraf:
```suggestion
      per_repo_quota_higher: 200
```

choldgraf:
lol that's a good point, I dunno why I picked an order-of-magnitude difference, I think I had too much coffee at scipy :-)

betatim:
Too much or not enough?

betatim:
Deployed now. Let's keep an eye on things with the goal of maybe reducing the limit again when the JupyterLab 1.0 rush is over?

betatim:
I bumped the size of the user pool to 10 (from 8).

Issue #1066, opened 2019-07-01
Load balancing strategies/customg build pod scheduling


betatim:
Triggered by https://github.com/jupyter/repo2docker/pull/718 (install R packages before copying the repo contents to allow fast rebuilds when only the README changed) and question about "hey I just built my repo but now I have to build it again" (caused by the user ending up on GKE vs OVH) I did some research on load balancing strategies.

Properties we need or would like to have:
* as little state as possible
* base decision only on information available in a request
* deal with changing number of clusters (or bins in LB lingo)
* simple to implement and debug

The standard in terms of making the LB strategy "sticky" (without explicit state) is to use [consistent hashing](https://en.wikipedia.org/wiki/Consistent_hashing). This strategy can deal with the number of clusters changing and only has to reassign a (small) fraction of sessions when the number of clusters changes.

With consistent hashing ([an example implementation](https://gist.github.com/betatim/304dbcb1b239f5775b16be563468f7d7)) we would always send a particular repo (`spec` in BinderHub terms) to the same cluster. This is good for all the kinds of caching we do. We'd stop assigning requests at random to a cluster and use this based on the spec.

One problem with consistent hashing is that the same key always goes to the same cluster. This means all the load for very popular spec (e.g. ipython-in-depth repo) would end up on one cluster, potentially exceeding the cluster's resources.

There are some clever approaches to solving this like https://arxiv.org/abs/1608.01350 but they require an estimate of the current load on each cluster. A simple one comes from Etsy ([here](https://codeascraft.com/2017/11/30/how-etsy-caches/)): add a single random character to the spec before computing which cluster it goes to. If you do this for "hot" keys they get spread around while "cold" keys always go to the same cluster.

How do we determine "hot" repos? I'd propose something like:
```python
# simulate a sequence of launches
launch_sequence = ["repo number %i" % i for i in random.choices(range(20),
                                                                range(1, 21),
                                                                k=10000)]
# list of the last N launches
last_launches = deque([], 1024)

# simulate launches
for n, launch in enumerate(launch_sequence):
    last_launches.append(launch)

print("Launch %i, top 5:" % n, Counter(last_launches).most_common(5))
```
this is super simple to implement, only takes into account recent launches and gives us an easy parameter to tune (how big do we make the top N?) to decide which repositories are "hot". If this doesn't work well we can always go to the big guns of using a count min sketch to estimate launches (but we'd have to teach it to forget).

Questions:
[...]

PR #1082, opened 2019-07-14
Proxy some routes instead of redirecting

betatim:
This changes the federation redirector so that most pages a user interacts with are proxied to gke.mybinder.org instead of redirecting people.

This means that as a user you see mybinder.org in the address bar when you visit the homepage and when waiting for a repo to launch. After this you are redirected.

You can try it out locally by starting `app.py` and then visiting http://localhost:8080/. What do people think of this change of behaviour? I think it makes it harder for people to copy&paste the "wrong" URL and with the new banner, the subdomains working and users seeing the hoster's name in the redirected URL we have a better balance between usability and exposing who is donating compute resources.


minrk:
Seems like it accomplishes the right thing. I'm still thinking about the implementation. What happens if a user refreshes the loading page? Will they get routed to a new cluster, or is there anything sticky that keeps them on the given cluster? What about sequential requests, e.g. a reconnecting eventstream? Will they reliably go to the same host?

betatim:
The redirect handler uses a cookie to make things "sticky" but that cookie only lasts for 10s. It gets set each time you make a request so if you make a (relevant) request every 8s you will always get the same cluster.

I picked 10s because it seemed like a good starting guess to balance "distribute people" and "keep people who hit refresh a bunch on the same connection".

Does a event stream reconnect? I thought that at the moment when I visit https://gke.mybinder.org/v2/gh/binder-examples/requirements/master (so no proxy, straight up binderhub) and press refresh I trigger a new launch and that the same thing would happen if my connection dropped for some reason.

I was thinking of working on https://github.com/jupyterhub/mybinder.org-deploy/issues/1066 (the title is wrong, should mention the federation) a bit more and through that solving the issue of sticky-ness. It would route you to a cluster based on which repo you are launching (modulo hot repos that somehow get distributed across clusters).

betatim:
Idea from the weekend: use a cookie per path (`/build/<spec>`) to remember which cluster we assigned the user to. This means we can use a really long (or infinite?) timeout as long as we check that the cluster is currently available. We still balance users across clusters. Need to check that the `path` attribute of cookies works as I think it does (cookie only gets sent if a request is for the (sub)path set in the cookie.

images/federation-redirect/app.py
```diff
@@ -60,7 +111,7 @@ def set_default_headers(self):
         # make sure the host is a valid choice and considered healthy
         if host_name not in self.host_names:
             host_name = random.choices(self.host_names, self.host_weights)[0]
-        self.set_cookie("host", host_name, max_age=10)
+        self.set_cookie("host", host_name, path=uri)
```
betatim: By setting the `path` parameter this cookie will only be sent back by the browser if a user visits the same path. So for repeated calls to http://example.com/v2/build/binder-examples/requirements/master we will see this cookie again and send the user to the same cluster again.

However a request to http://example.com/v2/build/binder-examples/r/master will not send the `host` cookie so we will pick a new cluster, which gets stored in a cookie again.

betatim:
@minrk what do you think of the idea of using a path specific cookie? I think it solves the "stickyness" problem.

It would be even nicer if we could do this on a repo basis (ignoring the branch/SHA1/etc) but I don't see a way to do that without having to put logic that understands the repo provider spec for each repo provider into the proxy.

betatim:
I'll merge this today as I'll have some time to watch what happens and revert it.

minrk:
I think we should take this into consideration when designing API v3. Routing cookies are definitely a standard way to do this, but I think we might be able to get away from that if we bake it into the multi-request build/launch API.

PR #1093, opened 2019-07-26
Adding a 'remote write' on prometheus for ovh configura…

mael-le-gal:
Add the following configuration for the `prometheus.yml` configuration in the  prometheus pod :

```yaml
remote_write:
- url: "<secret>"
  remote_timeout: 120s
  bearer_token: "<secret>"
  queue_config:
    max_shards: 1000
    min_shards: 4
  write_relabel_configs:
  # Filter metrics to keep
  - action: keep
    source_labels: [__name__]
    regex: "binderhub_launch_time_seconds_count.*|\
        binderhub_build_time_seconds_count.*"
```

⚠️  As the `remote_write` parameter is a `list` I was forced to put all the change for this PR on a `secret file` so you will need to pull locally my branch to validate it. Sorry for that.


betatim:
My prometheus-fu is low: what does this do? My guess is it writes to a remote storage specified by the URL?

So the end result is that the metrics listed at the end of the snippet are sent to the remote storage? Which you then use for alerting.

I'll merge this now and if we need to change/update something we can make a new PR.

mael-le-gal:
yes that's exactly that, it push the metric `binderhub_launch_time_seconds_count` and `binderhub_build_time_seconds_count` to our timeseries backend accessible on `remote_write.url`

That's the timeseries backend that we use in my team for all our internal alerts (the token has been generated for binder)

PR #1101, opened 2019-07-30
Treat weight zero hosts as if they didn't exist

betatim:
The federation proxy would treat hosts with weight zero as if they were
still part of the rotation when checking if a cookie based decision was
still valid. This PR fixes that and adds OVH back with weight zero.


choldgraf:
nice - seems good to me, maybe @willingc can test this on the link she clicked recently that was defaulting to the OVH cluster?

betatim:
Tricky because she didn't have a cookie. I'll test it once it is deployed by setting a cookie by hand to OVH.

PR #1103, opened 2019-07-30
Bump BinderHub

betatim:

https://github.com/jupyterhub/binderhub/compare/822f7c9...899fed2


betatim:
I got here before the henchbot!! :D

PR #1126, opened 2019-08-14
Issue template

choldgraf:
This adds an issue template directing people to the binder repository instead of this one. cc @betatim

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,11 @@
+<!--
+Welcome!
+
+✨✨✨✨
+This is the **deployment repository** for mybinder.org. If you've got
+a question or comment about the Binder Project, then please open an issue
+at https://github.com/jupyterhub/binder/issues/new.
+
+Thanks!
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+<!--
+Welcome!
+
+✨✨✨✨
+This is the **deployment repository** for mybinder.org. If you've got
+a question or comment about the Binder Project, then please open an issue
+at https://github.com/jupyterhub/binder/issues/new.
```
betatim: The "binder" repo is just as off topic for most things as this one is :-/ and it contains a pointer back to this repo in its issue template.

I think people incorrectly post to this repo so rarely that having a template come up that gets in the way of the mostly "legit" users and redirects those who are lost to a different (very likely also incorrect) repo is probably not worth it.

choldgraf: That's fair - I guess this is too much of an over-fit to this particular situation. I'll close this one because I don't think it'll add that much benefit

Issue #1144, opened 2019-08-30
Federation proxy, consider capacity when picking a target


bitnik:
from https://github.com/jupyterhub/mybinder.org-deploy/issues/1104#issuecomment-526112754:
> Thinking what else could be a good starting point for learning. The health endpoint now includes information about how many pods are running and a hub can optionally announce what capacity it has (the pod quota part in https://gke.mybinder.org/health). One thing to do is to extend the federation proxy https://github.com/jupyterhub/mybinder.org-deploy/blob/master/images/federation-redirect/app.py to take into account whether or not a hub is above its capacity when picking a target. We'd also have to configure a capacity for OVH, I'd suggest 50 to start with as that seems to work as maximum and sometimes we go above it (-> we can observe the limiting works).

bitnik:
Closing this (https://github.com/jupyterhub/mybinder.org-deploy/pull/1145).

PR #1145, opened 2019-08-30
Federation proxy, consider capacity when picking a target

bitnik:
For #1144:

- [ ] extend the federation proxy to take into account whether or not a hub is above its capacity when picking a target
- [x] configure pod quota of OVH as 50
- [ ] Update README.md of federation-redirect (https://github.com/jupyterhub/mybinder.org-deploy/blob/master/images/federation-redirect/README.rst)

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -23,13 +23,13 @@ CONFIG = {
         "gke": dict(
             url="https://gke.mybinder.org",
             weight=1,
-            health="https://gke.mybinder.org/versions",
+            health="https://gke.mybinder.org/health",
             prime=True,
         ),
         "ovh": dict(
             url="https://ovh.mybinder.org",
             weight=1,
-            health="https://ovh.mybinder.org/versions",
+            health="https://ovh.mybinder.org/health",
             # health="https://httpbin.org/status/404",
         ),
     },
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -125,18 +125,18 @@ async def health_check(host, active_hosts):
     try:
         for n in range(check_config["retries"]):
             try:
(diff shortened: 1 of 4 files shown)


choldgraf:
This looks like a great idea imo!

bitnik:
@betatim can you review the changes? Now it considers the target as unhealty if it is above its capacity. But now I am thinking that probably you meant to decrease it's weight (or something similar) while picking a new target? (not removing it from rotation until it has less running pods than pod_quota)

images/federation-redirect/app.py
```diff
@@ -125,10 +125,19 @@ def set_default_headers(self):
     try:
         for n in range(check_config["retries"]):
             try:
-                await client.fetch(
+                response = await client.fetch(
```
betatim: Because of https://github.com/jupyterhub/binderhub/blob/f746e500b73645d0774765d6e6e1261aa84f0fd1/binderhub/health.py#L174-L180

I think what happens is that if the service is unhealthy the status code is 50x which means this call raises an exception. However the pod quota is considered a "soft check" (better name needed?) so being above quota doesn't set the HTTP status call to 50x. This means I'd only check the pod quota health in the next few lines and let the `except` block deal with all the other unhealthy conditions.

betatim:
I think removing the host from the rotation for a while is fine if it is over its quota. I think right now if you get removed you will be retried in 5minutes time. My hunch (unverified by data) is that if you are above the quota now, in 5min you will still be above quota or have gone down a bit. We don't need to make it so you are always "at the quota". I think it is better to let usage drop a bit below the quota before sending traffic again.

images/federation-redirect/app.py
bitnik: Done, can you check it again?

betatim:
LGTM. Merging to see it in action

PR #1162, opened 2019-09-16
[WIP] Upgrade to cert-manager from kube-lego

### LINKED ISSUE #1148: Migrate mybinder.org from kube-lego to cert-manager for LetsEncrypt
Hi all,

I've been chatting with @consideRatio and @minrk at the workshop about enabling LetsEncrypt/HTTPS for the Turing private BinderHub. The common theme seems to be that kube-lego is deprecated and cert-manager is the new way forward. Min invited me to upgrade mybinder.org to cert-manager as a learning exercise and produce some documentation of the process. [Erik's NeurIPS deployment](https://github.com/consideRatio/neurips.mybinder.org-deploy) uses cert-manager so I'll look into that too.

Which clusters have been migrated:

- [x] GKE staging (PR: #1406)
- [x] GKE prod (PR: #1415)

sgibson91:
Hi all,

This is my attempt to update mybinder.org to using [`cert-manager`](https://github.com/jetstack/cert-manager) instead of `kube-lego`, which is deprecated.

So far, I've attempted (but not deployed):

* adding `clusterissuer.yaml` and `_helpers.tpl` files,
* renaming secrets that contained references to `kube-lego`,
* updating the staging config.

Any comments appreciated!

Fixes #1148


manics:
In the past I've had problems with CRDs and uses of CRDs. Basically the implementation of the CRD has to be installed _before_ anything that defines an instance of the CRD.
https://github.com/helm/helm/blob/master/docs/chart_best_practices/custom_resource_definitions.md

The docs for the cert-manager Helm chart suggest it's even more confusing 😟:
https://docs.cert-manager.io/en/latest/getting-started/install/kubernetes.html#steps

sgibson91:
Thanks @manics, yes @consideRatio and I ran into this issue when installing cert-manager on the Turing's BinderHub when we were in Oslo 😊 I'm more asking if I've removed anything important from the config 😱

mybinder/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
...
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "hub23-chart.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
```
consideRatio: Replaced the template definition names in this file

config/staging.yaml
```diff
@@ -119,3 +124,7 @@ federationRedirect:
       url: https://gke2.staging.mybinder.org
       weight: 1
       health: https://gke2.staging.mybinder.org/versions
+
+cert-manager:
+  ingressShim:
+    defaultIssuerName: "staging"
```
consideRatio: Will this be set to production someplace also? It should not be left unset for any deployment I think.

It probably makes sense to have it set to staging by default in values.yaml and then overridden to production elsewhere.

sgibson91: Agreed. Also my plan is to upgrade the staging cluster first before deploying onto the prod cluster.

sgibson91:
> I think renaming all the ingress TLS termination secrets at once could leave us throttled or temporary banned with lets encrypt as there were more than 4 or so. You could copy the existing secrets to the new name if you like to mitigate this issue, or look up lets encrypt documentation details about this as I'm certainly not certain about this being an issue or how it will behave if it is an issue.

I've reverted my renaming of secrets instead.

> BTW: How is labels typically set in this helm charts kubernetes resources? Perhaps we should mimic that rather than introducing helpers to set the labels? I think we should aim for consistency with the repo rather than consistency with other Helm charts if there is many other resources aligned to follow a certain practice within this meta helm chart.

I'm not sure that I follow this? @betatim @minrk @choldgraf?

mybinder/templates/_helpers.tpl
sgibson91: Done!

config/staging.yaml
sgibson91: Set staging as default in values.yaml

mybinder/requirements.yaml
```diff
@@ -8,9 +8,9 @@ dependencies:
  - name: grafana
    version: 1.18.0
    repository: https://kubernetes-charts.storage.googleapis.com
- - name: kube-lego
-   version: 0.4.2
-   repository: https://kubernetes-charts.storage.googleapis.com
+ - name: cert-manager
+   repository: https://charts.jetstack.io
+   version: v0.10.0
```
betatim: If we do this won't we also lose kube-lego for production? Not sure what the solution is to have kube-lego and cert-manager available in parallel but I think we need to figure that out.

sgibson91: Interesting question. Does an upgrade also uninstall dependent charts that have been removed or just apply the diff? I can add the dependency back in for now.

betatim: Not sure, but I'd expect it to uninstall things as well. Needs trying out for a definitive answer.

consideRatio: Removing it from requirements.yaml will remove it from the k8s cluster on deploy.

ref: https://github.com/helm/helm/pull/2850

sgibson91:
As per https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1448, if anyone wants to help out/take over this PR then they should feel free!

sgibson91:
Closing due to a new plan of attack, see https://github.com/jupyterhub/mybinder.org-deploy/issues/1148#issuecomment-545419571

PR #1163, opened 2019-09-16
[ISSUE-1160] Switching on ovh traffic

mael-le-gal:
Upgraded the ovh cluster from `1.11.x` to `1.14.x` => everything is green on my side so I think we can switch on the traffic on `ovh` again :)


choldgraf:
Let's give it a whirl!

PR #1186, opened 2019-10-04
reduce traffic to GESIS to 0 temporarily for 07.10

bitnik:
Hi, on 07.10 there will be maintenance on one of our servers. We don't expect downtime but since it is our big server, we will lose big amount of resources temporarily (less capacity of having user pods) and we expect that everything will be slow (the other server has to download docker images from hub).

It would be perfect if we could have this PR merged during Sunday. When the maintenance is done, I will make another PR to increase it again. Maybe that time we even increase it to 5% :)

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -420,7 +420,7 @@ federationRedirect:
   hosts:
     gke:
       url: https://gke.mybinder.org
-      weight: 80
+      weight: 81
       health: https://gke.mybinder.org/health
       prime: true
     ovh:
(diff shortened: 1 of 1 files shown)


betatim:
Is it ok if we merge it later today? Otherwise it might get forgotten over the weekend :-/

bitnik:
yes! Btw actually there is another option is that we can set the `pod_quota` to 0 before the maintenance starts.

We can go with the solution that you prefer :)

betatim:
I like the `pod_quota` idea. It is meant to give power to the cluster operator so lets use it :)

Issue #1188, opened 2019-10-08
Federation redirector - versions check


bitnik:
Hi, in September team meeting we were discussing about how to keep binder instances in sync. And one idea was

> if a member of the federation is “too far behind” we automatically stop sending traffic

For now this is a case only for GESIS Binder because we do the upgrades through our own repo and sometimes they are delayed.

So first questions is do we want to add versions check into federation redirector? If yes, I would like to do that. I think can start with a PR "if an instance doesn't have the same versions as mybinder.org, remove it from list of healthy binders". What do you think?

betatim:
Sounds like a good plan to me. Like you suggested I would check the version(s) of the tools on https://gke.mybinder.org/versions (this should be marked as "prime" in the redirector so we don't have to hard code it) and then mark every other host that doesn't have these exact versions as "unhealthy'.

manics:
As a long-term goal it might be nice to allow other binderhubs to be upgraded to newer minor versions (so backwards compatible according to semver), before the main binderhub is upgraded. This would allow full capacity to be maintained.

choldgraf:
Since the major/minor versions etc don't mean a ton in BinderHub, what if we also included the date in the `/versions` endpoint? So it'd serve up something like:

```
{
 "builder": "jupyter/repo2docker:0.10.0-100.g0dba611",
 "builder_date": "2019-02-02",
 "binderhub": "0.1.0+406.ga168d06.dirty",
 "binderhub_date": "2019-08-01"
}
```

In https://github.com/jupyterhub/binderhub/pull/863/files I couldn't figure out how to send a request to the GitHub API to figure out the date because requests were blocked, but if the dates were in there it'd be trivial to figure out "how many days is XXX BinderHub behind the master BinderHub".

PR #1191, opened 2019-10-10
Federation redirector: check versions

bitnik:
This is a simple attempt for https://github.com/jupyterhub/mybinder.org-deploy/issues/1188

We can start with this or based on feedback improve it.

One thing I wanted to do but I was not sure was increasing the number of retries. Because there are 2 checks in the same loop.

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -19,7 +19,7 @@ from tornado.web import RequestHandler
 
 # Config for local testing
 CONFIG = {
-    "check": {"period": 10, "jitter": 0.1, "retries": 3, "timeout": 2},
+    "check": {"period": 10, "jitter": 0.1, "retries": 5, "timeout": 2},
     "hosts": {
         "gke": dict(
             url="https://gke.mybinder.org",
@@ -135,6 +135,7 @@ async def health_check(host, active_hosts):
     try:
         for n in range(check_config["retries"]):
             try:
+                # TODO we could use `asyncio.gather()` and fetch health and versions in parallel
                 # raises an `HTTPError` if the request returned a non-200 response code
                 # health url returns 503 if a (hard check) service is unhealthy
                 response = await client.fetch(
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -153,9 +153,13 @@ async def health_check(host, active_hosts):
                     all_hosts[host]["versions"], request_timeout=check_config["timeout"]
                 )
                 versions = json.loads(response.body)
+                # if this is the prime host store the versions so we can compare to them later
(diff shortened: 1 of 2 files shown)


betatim:
I'd be Ok with increasing the number of retries a bit, maybe 5? I think the only downside would be that a complete check now takes longer than the time between scheduled checks. Though it is probably not much of a problem as we only schedule the next check once the first one is done.

betatim:
Actually, having re-read the code: forget my comment about checks taking too long (lets go with five retries?) and also the comment about `gather` (I think it makes things more complex with very little benefit).

bitnik:
I think 5 retries is good. I just increased it and added a comment about gather.

images/federation-redirect/app.py
```diff
@@ -137,6 +148,16 @@ def set_default_headers(self):
                         if not check["ok"]:
                             raise Exception("{} is over its quota: {}".format(host, check))
                         break
+                # check versions
+                response = await client.fetch(
+                    all_hosts[host]["versions"], request_timeout=check_config["timeout"]
+                )
+                versions = json.loads(response.body)
```
betatim:
```suggestion
                versions = json.loads(response.body)
                # if this is the prime host store the versions so we can compare to them later
```

images/federation-redirect/app.py
```diff
@@ -137,6 +148,16 @@ def set_default_headers(self):
...
+                if all_hosts[host].get("prime", False):
+                    all_hosts["versions"] = versions
+                elif versions != all_hosts.get("versions", versions):
```
betatim:
```suggestion
                # check if this cluster is on the same versions as the prime
                # w/o information about the prime's version we allow each
                # cluster to be on its own versions
                if versions != all_hosts.get("versions", versions):
```

betatim: Changes from `elif` to `if`

betatim:
What do you think of the two suggestions? For me they make it easier to think about which case goes down which branch of the if statements.

Otherwise this looks good to me and I'll attempt to deploy it later today to see what happens.

images/federation-redirect/app.py
bitnik: changing from `elif` to `if` is programmatically okay but logically i don't understand it

bitnik: if a host is not prime, than check its versions with versions of prime

betatim: For me it is easier to think about "is this a prime? then fetch the version" and "is this host on the right version?" as two separate things. This is why I thought splitting it up into two blocks instead of using `if ... elif ...`.

bitnik:
Thanks for the suggestions :)

PR #1193, opened 2019-10-11
handler to show active hosts

bitnik:
It would be nice to see current list of active hosts in federation. For example at the beginning of this week, I set `pod_quota` to 0 but I was not sure if GESIS Binder was really removed from active hosts, because the number of launched pods were are not decreased as we expected.

Now I am thinking it may be even nicer to serve this data in a form that we can display in grafana.

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -132,8 +132,7 @@ class ActiveHostsHandler(RequestHandler):
         self.active_hosts = active_hosts
 
     async def get(self):
-        self.set_header("Content-type", "application/json")
-        self.write(json.dumps({"active_hosts": self.active_hosts}))
+        self.write({"active_hosts": self.active_hosts})
 
 
 async def health_check(host, active_hosts):
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -126,6 +126,16 @@ class RedirectHandler(RequestHandler):
         self.redirect(url)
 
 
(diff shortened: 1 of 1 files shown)


images/federation-redirect/app.py
```diff
@@ -126,6 +126,16 @@ def set_default_headers(self):
...
 
 
+class ActiveHostsHandler(RequestHandler):
+    """Serve information about active hosts"""
+    def initialize(self, active_hosts):
+        self.active_hosts = active_hosts
+
+    async def get(self):
+        self.set_header("Content-type", "application/json")
+        self.write(json.dumps({"active_hosts": self.active_hosts}))
```
betatim:
```suggestion
        self.write({"active_hosts": self.active_hosts})
```
I learnt we can write dictionaries directly to the response and tornado takes care of headers and the like.

images/federation-redirect/app.py
```diff
@@ -126,6 +126,16 @@ def set_default_headers(self):
         self.redirect(url)
 
 
+class ActiveHostsHandler(RequestHandler):
+    """Serve information about active hosts"""
+    def initialize(self, active_hosts):
+        self.active_hosts = active_hosts
+
+    async def get(self):
+        self.set_header("Content-type", "application/json")
```
betatim:
```suggestion
```

betatim:
Good idea. Will merge after the two suggestions.

bitnik:
Thanks, done.

PR #1202, opened 2019-10-15
adding binder buttons to Notebook UI

bitnik:
for https://github.com/jupyterhub/binderhub/issues/674

This PR adds 2 buttons into Notebook UI:
- Visit repo: opens the repo url in new tab
- Copy Binder link: copies the binder launch url into clipboard

In `jupyter_notebook_config.py` there are 4 template variables defined with different urls, here is an example:
- `repo_url`: https://github.com/bitnik/simple-binder-repo-2
- `ref_url`: https://github.com/bitnik/simple-binder-repo-2/tree/79540e5802824cdf6432d4bc22ad7849ed4cd6b0
- `binder_url`: https://mybinder.org/v2/gh/bitnik/simple-binder-repo-2/master
- `persistent_binder_url`: https://mybinder.org/v2/gh/bitnik/simple-binder-repo-2/79540e5802824cdf6432d4bc22ad7849ed4cd6b0

These template variables takes values from env variabels. For that I used `appendix` of BinderHub. But I have an idea to get rid of appendix. I think we can define these env variables during build process in binder. I will make a PR for that.

I updated notebook templates (they are mounted into user pods as cofigmaps) to add buttons, so `custom.js` is not used anymore. And I used `ref_url` for "Visit repo" button and `persistent_binder_url` for "Copy Binder link" button. In [login.html](https://github.com/gesiscss/mybinder.org-deploy/blob/57be8dfd1ad5f6a5ea2640994d413cabd5e04e79/mybinder/files/etc/jupyter/templates/login.html#L14-L30) `binder_url` was used and I didn't change it. Should we use `repo_url` and `binder_url` pair or `ref_url` and `persistent_binder_url` pair or a mix of them?

### DIFF
--- a/mybinder/files/etc/jupyter/jupyter_notebook_config.py
+++ b/mybinder/files/etc/jupyter/jupyter_notebook_config.py
@@ -1,13 +1,17 @@
 import os
 c.NotebookApp.extra_template_paths.append('/etc/jupyter/templates')
 
-federation_host = 'https://mybinder.org'
-current_hosts = ['https://gke.mybinder.org', 'https://ovh.mybinder.org']
-binder_url = os.environ.get('BINDER_URL', federation_host)
-persistent_binder_url = os.environ.get('PERSISTENT_BINDER_URL', '')
-for h in current_hosts:
-    binder_url = binder_url.replace(h, federation_host)
-    persistent_binder_url = persistent_binder_url.replace(h, federation_host)
+
+def make_federation_url(url):
+    federation_host = 'https://mybinder.org'
+    if not url:
+        return federation_host
+    url_parts = url.split('/v2/', 1)
+    return federation_host + '/v2/' + url_parts[-1]
+
+
+binder_url = make_federation_url(os.environ.get('BINDER_URL', ''))
+persistent_binder_url = make_federation_url(os.environ.get('PERSISTENT_BINDER_URL', ''))
 
 c.NotebookApp.jinja_template_vars.update({
     'binder_url': binder_url,
--- a/mybinder/files/etc/jupyter/jupyter_notebook_config.py
+++ b/mybinder/files/etc/jupyter/jupyter_notebook_config.py
@@ -1,5 +1,17 @@
 import os
 c.NotebookApp.extra_template_paths.append('/etc/jupyter/templates')
+
+federation_host = 'https://mybinder.org'
+current_hosts = ['https://gke.mybinder.org', 'https://ovh.mybinder.org']
+binder_url = os.environ.get('BINDER_URL', federation_host)
+persistent_binder_url = os.environ.get('PERSISTENT_BINDER_URL', '')
+for h in current_hosts:
+    binder_url = binder_url.replace(h, federation_host)
+    persistent_binder_url = persistent_binder_url.replace(h, federation_host)
+
 c.NotebookApp.jinja_template_vars.update({
-    'binder_url': os.environ.get('BINDER_URL', 'https://mybinder.org'),
+    'binder_url': binder_url,
+    'persistent_binder_url': persistent_binder_url,
+    'repo_url': os.environ.get('REPO_URL', ''),
+    'ref_url': os.environ.get('REF_URL', ''),
 })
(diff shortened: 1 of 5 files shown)


bitnik:
> But I have an idea to get rid of appendix. I think we can define these env variables during build process in binder. I will make a PR for that.
My solution didn't work. Maybe someone else has an idea how to do that?

bitnik:
@betatim could you make a review?

bitnik:
Btw I already deployed on GESIS Binder. Here you can check how buttons look like and work: https://notebooks.gesis.org/binder/v2/gh/bitnik/simple-binder-repo-2/master

mybinder/files/etc/jupyter/jupyter_notebook_config.py
```diff
@@ -1,5 +1,17 @@
 import os
 c.NotebookApp.extra_template_paths.append('/etc/jupyter/templates')
+
+federation_host = 'https://mybinder.org'
+current_hosts = ['https://gke.mybinder.org', 'https://ovh.mybinder.org']
```
betatim: Is there a way we can get this from the list of federation clusters in `values.yaml`? Having to check in several places when we change the members is going to lead to mistakes. Or maybe we don't need an explicit list? Can we remove everything before `/user` in `binder_url` maybe?

betatim:
Sorry for being so slow. I wanted to see if I could make a better suggestion for the comment I made but didn't find time to think about it yet. Hence just a comment with no good suggestion on how to improve it :-/

mybinder/files/etc/jupyter/jupyter_notebook_config.py
bitnik: that's true, these urls will most probably be forgotten. based on your suggestion, what about using this function?
```python
def make_federation_url(url):
    federation_host = 'https://mybinder.org'
    if not url:
        return federation_host
    url_parts = url.split('/v2/', 1)
    return federation_host + '/v2/' + url_parts[-1]
```

betatim: What could possibly go wrong? :D -> gets my support. We could probably still improve it but yeah.

Why check if `url` is empty? As in, when would it happen that it is empty?

bitnik: :) that is the case for `dockerfile` repos, they will mostly have `BINDER_URL` and `PERSISTENT_BINDER_URL` env variables not defined. because these env variables are set in appendix.

bitnik:
[login.html](https://github.com/gesiscss/mybinder.org-deploy/blob/57be8dfd1ad5f6a5ea2640994d413cabd5e04e79/mybinder/files/etc/jupyter/templates/login.html#L14-L30), which is mostly displayed to users when they share url of their binder instance, uses `binder_url`, should I change it to `persistent_binder_url`. Or should I replace `persistent_binder_url` with `binder_url` in html for buttons? This I can't decide :)

bitnik:
Now I think that it should stay as how it is now. `binder_url` in "login.html" and `persistent_binder_url` in button.

betatim:
Merged!

choldgraf:
This is awesome! Just to confirm, what's the expected behavior here? I clicked on the button, but it just copied the website "https://mybinder.org" and that's it...

betatim:
There should be two buttons. One takes you to the repo and one copies the launch URL into your clipboard. Which repo did you try it on?

choldgraf:
That's the repository that I tried as well, maybe a repo-dependent problem?

manics:
Presumably `appendix` only takes effect on a rebuild, so existing repos built before this change won't have the new environment variables until the repo is updated?

bitnik:
> Presumably appendix only takes effect on a rebuild, so existing repos built before this change won't have the new environment variables until the repo is updated?

yes, correct

> Could the environment variables be injected at runtime instead?

That would be much better. I remember that I tried to do it but I couldn't find a way.

manics:
Turns out passing `repo_url` is easy, it's already passed! https://github.com/jupyterhub/binderhub/pull/1010

PR #1205, opened 2019-10-16
[MRG] Increase memory request and limit of nginx-ingress

betatim:
Since the switch to the new node pool our nginx pods are using more
memory and as a result keep getting killed because they exceed their
limit.

Screenshot 2019-10-16 at 07 14 59


betatim:
The three deploys marked are (reruns of) https://github.com/jupyterhub/mybinder.org-deploy/pull/1197/files

PR #1211, opened 2019-10-23
cert manager install: step 1 - adding files

sgibson91:
This PR is adding relevant files and installing cert-manager without affecting any clusters.

Related issue: #1148

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,11 @@ before_deploy:
   # Stage 2, Step 2: Set up helm!
   helm init --client-only
   helm repo add jupyterhub https://jupyterhub.github.io/helm-chart
+  helm repo add cert-manager https://charts.jetstack.io
   helm repo update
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -59,8 +59,11 @@ before_deploy:
   # Stage 2, Step 2: Set up helm!
   helm init --client-only
   helm repo add jupyterhub https://jupyterhub.github.io/helm-chart
+  helm repo add cert-manager https://charts.jetstack.io
   helm repo update
   (cd mybinder && helm dep up)
+  # Add CustomResourceDefinitions for cert-manager
+  kubectl apply -f https://raw.githubusercontent.com/jetstack/cert-manager/release-0.10/deploy/manifests/00-crds.yaml
```
betatim: Needs access to the cluster so we need to run this as part of `deploy.py` where the cluster is selected and auth is setup.

PR #1227, opened 2019-10-24
Keep empty string when there is no env variable

betatim:
Follow up to #1202 that keeps the empty string for situations where the environment doesn't contain any information for the buttons.

### DIFF
--- a/mybinder/files/etc/jupyter/jupyter_notebook_config.py
+++ b/mybinder/files/etc/jupyter/jupyter_notebook_config.py
@@ -5,7 +5,7 @@ c.NotebookApp.extra_template_paths.append('/etc/jupyter/templates')
 def make_federation_url(url):
     federation_host = 'https://mybinder.org'
     if not url:
-        return federation_host
+        return ''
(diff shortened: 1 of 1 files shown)


betatim:
cc @choldgraf

bitnik:
@betatim thanks for the follow-up! 

the problem was that already built images don't have required env variables for buttons. this PR fixes this problem. but now there is a problem in [login.html](https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/files/etc/jupyter/templates/login.html), because same images now have `binder_url`, which is used in login.html, as empty string. For example check this link: https://hub.gke.mybinder.org/user/binder-examples-requirements-6xssa9qw/tree - this is link of the binder session I just launched. I think we just need another PR to update login.html, so it displays a generic message if there is no url, what do you think?

betatim:
Sounds like a good idea.

bitnik:
I created an issue for this, https://github.com/jupyterhub/mybinder.org-deploy/issues/1232

PR #1236, opened 2019-10-30
Change Gesis hostname and increase quota

betatim:
This changes the hostname used for the Gesis cluster to gesis.mybinder.org and increases its weight to 5% of all traffic.

### DIFF
--- a/docs/source/federation.rst
+++ b/docs/source/federation.rst
@@ -14,7 +14,7 @@ the BinderHub federation, who is in it, how to join it, etc, see
 ==========================  ========  ===============  ==============  =========
 gke.mybinder.org
 ovh.mybinder.org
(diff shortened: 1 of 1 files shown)


choldgraf:
ugh it's annoying that circle is erroring, I didn't realize it'd use the checks if no `.circleci` folder was there. I'll make a PR to add this in

betatim:
Reverting this again as we need to change the CORS settings (I think):
Screenshot 2019-10-30 at 21 55 04

betatim:
cc @bitnik

bitnik:
@betatim https://github.com/gesiscss/orc/commit/ecaeea74929be1fdf3af213b0dac08ad7233bb9b should fix this problem. can we try it again?

PR #1241, opened 2019-11-02
Fix pod antiaffinity spec, add unschedulable toleration

betatim:
Let the pod reschedule itself when the node it is on gets cordoned.

I think the way the pod anti-affinity was specified so far was wrong and hence had no effect.

The new toleration causes the pod to reschedule itself when the node gets cordoned.


betatim:
Our GKE cluster is running with `TaintBasedEvictions=false` which means the taint based eviction that the toleration relies on is turned off. You can check the state of the feature gates with `kubectl cluster-info dump | grep feature-gate`.

PR #1249, opened 2019-11-06
more general banner message

arnim:
This PR would make the banner message more suitable for members of the federation whose support is not limited to "sponsoring our computers". @betatim


betatim:
Works for me.

PR #1250, opened 2019-11-06
Revert "compatibility with notebook < 5.1.0"

betatim:
Reverts jupyterhub/mybinder.org-deploy#1247

Seeing the following exception when starting our CI repo:
```
Traceback (most recent call last):
  File "/srv/conda/envs/notebook/bin/jupyter-notebook", line 11, in <module>
    sys.exit(main())
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/jupyter_core/application.py", line 266, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "</srv/conda/envs/notebook/lib/python3.7/site-packages/decorator.py:decorator-gen-7>", line 2, in initialize
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/notebook/notebookapp.py", line 1630, in initialize
    self.init_webapp()
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/notebook/notebookapp.py", line 1378, in init_webapp
    self.jinja_environment_options,
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/notebook/notebookapp.py", line 159, in __init__
    default_url, settings_overrides, jinja_env_options)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/notebook/notebookapp.py", line 181, in init_settings
    env = Environment(loader=FileSystemLoader(template_path), extensions=['jinja2.ext.i18n'], **jenv_opt)
TypeError: type object got multiple values for keyword argument 'extensions'
```

So reverting for the moment to investigate.

PR #1247, opened 2019-11-05
compatibility with notebook < 5.1.0

btel:
This PR fixes the issue with compatibility with notebook servers < 5.1.0. 

The issue was that notebook server < 5.1.0 crashed when rendering the tree view with the templates injected by mybinder.org. The problem was due to the missing jinja2 extension in nb 5.0.0.

For details, see the thread at discourse: 

[Binder displays 500 internal server error when launching a built image](https://discourse.jupyter.org/t/binder-displays-500-internal-server-error-when-launching-a-built-image/2597/7)

PR #1259, opened 2019-11-08
[MRG] Enable translation Jinja extension for old notebooks

betatim:
Following up #1250 and #1247

This checks the version of the notebook and only explicitly adds the translation extension for older versions.


btel:
I confirm that the fix works for my repo (with notebook 5.0.0). Thanks.

https://mybinder.org/v2/gh/btel/jupyter-dashboards-binder/9bf038c253d4466547952a47b2fbe182b0b2c548

PR #1273, opened 2019-11-22
Migrating to cert-manager: Step 1 - adding new config files

sgibson91:
- `clusterissuer.yaml` is added
- `_helpers.tpl` is added
- Let's Encrypt contact email added to `values.yaml`
- `cert-manager` added to helm chart requirements
- `cert-manager` custom resource definitions installed in `deploy.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ mybinder/requirements.lock
 docs/_build
 travis/crypt-key
 env
+
+.vscode/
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -13,3 +13,5 @@ mybinder/requirements.lock
 docs/_build
 travis/crypt-key
 env
+
+.vscode/
```
betatim: Please don't add these personal files and directories. Instead use a global place to configure your ignores like `~/.gitignore`. This is so we don't end up with 101 different, personal ignore patterns here.

PR #1278, opened 2019-11-25
Reducing OVH quota further

betatim:
It seems like the failed launches are increasing, so testing what happens if we reduce the load

It could be that it looks more dramatic than it is (if there are only 5 launches then one failure will put you straight at 80% success rate). Needs more investigating.

### DIFF
--- a/config/ovh.yaml
+++ b/config/ovh.yaml
@@ -3,7 +3,7 @@ projectName: ovh
 binderhub:
   config:
     BinderHub:
-      pod_quota: 80
+      pod_quota: 20
       hub_url: https://hub-binder.mybinder.ovh
       badge_base_url: https://mybinder.org
       image_prefix: dlqpwel7.gra5.container-registry.ovh.net/binder/r2d-f18835fd-


betatim:
This deploy initially failed because tiller on staging was at v2.14 but we use v2.11. This change happened around 2d15h ago (now nearly 8am on Monday). Does anyone know what caused this upgrade?

I changed staging back to v2.11 by running `helm init --upgrade --force-upgrade` from my local machine after looking at the tiller pod with ` kubectl get pods -n kube-system`.

Restarting the build now

consideRatio:
@sgibson91 bumped tiller for cert-manager stuff i believe.

What was the error of using tiller 2.14.x?

betatim:
The error was that the local version is older than the remote one. I think it is fixable by upgrading the one we use on the CI as well. More a question of how we got into a state where we can't deploy.

consideRatio:
If you use the latest 2.15.x you will not get issues, but if you use 2.16.x you cannot use it with too old tillers as well, so if you want something robust atm you can go for the latest of 2.15.x on tiller+helm client

PR #1282, opened 2019-11-29
Manually bump BinderHub chart

betatim:
This is a manual BinderHub chart bump. It seems @henchbot is out of action at the moment (too much turkey? :D )

### DIFF
--- a/mybinder/requirements.yaml
+++ b/mybinder/requirements.yaml
@@ -12,5 +12,5 @@ dependencies:
    version: 0.4.2
    repository: https://kubernetes-charts.storage.googleapis.com
  - name: binderhub
-   version: 0.1.0-456.7e32ac0
+   version: 0.2.0-013.dace2ee
    repository: https://jupyterhub.github.io/helm-chart


henchc:
Ya I think the new hash system confused it @betatim , I will look into it this weekend!

consideRatio:
@henchc I'll try getting the documentation on how chartpress works now, but in short, the versions will look like:

0.2.0
0.2.0-006.shaa
0.3.0-beta.1
0.3.0-beta.1.009.asdf123
0.3.0-beta.1.020.sdfg234
0.3.0

Note that SHA's are min 4 chars, but typically 7, and that the build suffix will always be to append either a . or - depending on if there has already been a `-`, three or more digits representing the commits since the tag, and a commit sha.

betatim:
Thanks for taking a look. It seems like henchbot has recovered since then as there has been an automated PR.

It also made me realised how much I rely on henchbot working. When it was broken I was "ah, so, how do we do this manually again?" took a good few minutes to remember :D

consideRatio:
:)

Btw docs updated in chartpress, release not made yet.

henchc:
Sorry for the late reply. I'm catching up here but I think you all figured out that the bot died when the version was moved from 0.2.0 to 0.1.0 here: https://github.com/jupyterhub/mybinder.org-deploy/commit/1a65ea6ab20b77508758ccba2a7d5a6605280be1

Because I had hard-coded that version for some reason (can't remember why now)...

Anyway, I've removed the hard-coding of the version so we *should* be ok now after this: https://github.com/henchbot/mybinder.org-upgrades/commit/4bc630746b4ccfb82a8c742ad36bd74d6ad44eee

PR #1284, opened 2019-11-29
remove appendix and use new env vars from runtime

bitnik:
ref: https://github.com/jupyterhub/binderhub/pull/1010

### DIFF
--- a/mybinder/files/etc/jupyter/jupyter_notebook_config.py
+++ b/mybinder/files/etc/jupyter/jupyter_notebook_config.py
@@ -13,20 +13,13 @@ if V(notebook.__version__) < V('5.1.0'):
     c.NotebookApp.jinja_environment_options = {'extensions': ['jinja2.ext.i18n']}
 
 
-def make_federation_url(url):
-    federation_host = 'https://mybinder.org'
-    if not url:
-        return ''
-    url_parts = url.split('/v2/', 1)
-    return federation_host + '/v2/' + url_parts[-1]
-
-
-binder_url = make_federation_url(os.environ.get('BINDER_URL', ''))
-persistent_binder_url = make_federation_url(os.environ.get('PERSISTENT_BINDER_URL', ''))
+binder_launch_host = os.environ.get('BINDER_LAUNCH_HOST', '')
+binder_request = os.environ.get('BINDER_REQUEST', '')
+binder_persistent_request = os.environ.get('BINDER_PERSISTENT_REQUEST', '')
 
 c.NotebookApp.jinja_template_vars.update({
-    'binder_url': binder_url,
-    'persistent_binder_url': persistent_binder_url,
-    'repo_url': os.environ.get('REPO_URL', ''),
-    'ref_url': os.environ.get('REF_URL', ''),
+    'binder_url': binder_launch_host+binder_request,
+    'persistent_binder_url': binder_launch_host+binder_persistent_request,
+    'repo_url': os.environ.get('BINDER_REPO_URL', ''),
+    'ref_url': os.environ.get('BINDER_REF_URL', ''),
 })
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -68,13 +68,6 @@ binderhub:
       build_image: jupyter/repo2docker:0.10.0-117.g3337eaa
       per_repo_quota: 100
       per_repo_quota_higher: 200
-      appendix: |
-        USER root
-        ENV BINDER_URL={binder_url}
-        ENV PERSISTENT_BINDER_URL={persistent_binder_url}
-        ENV REPO_URL={repo_url}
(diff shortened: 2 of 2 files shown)


mybinder/files/etc/jupyter/jupyter_notebook_config.py
```diff
@@ -13,20 +13,13 @@
...
+binder_launch_host = os.environ.get('BINDER_LAUNCH_HOST', '')
+binder_request = os.environ.get('BINDER_REQUEST', '')
+binder_persistent_request = os.environ.get('BINDER_PERSISTENT_REQUEST', '')
 
 c.NotebookApp.jinja_template_vars.update({
-    'binder_url': binder_url,
-    'persistent_binder_url': persistent_binder_url,
-    'repo_url': os.environ.get('REPO_URL', ''),
-    'ref_url': os.environ.get('REF_URL', ''),
+    'binder_url': binder_launch_host+binder_request,
```
betatim:
```suggestion
    'binder_url': binder_launch_host + binder_request,
```

manics:
Does Gesis have a `badge_base_url` setting?
- https://github.com/jupyterhub/mybinder.org-deploy/search?q=badge_base_url&unscoped_q=badge_base_url
- https://github.com/gesiscss/mybinder.org-deploy/search?q=badge_base_url&unscoped_q=badge_base_url

E.g. https://mybinder.org/v2/gh/manics/jupyter-offlinenotebook/master?filepath=example.ipynb currently goes to Gesis:
`BINDER_LAUNCH_HOST	https://notebooks.gesis.org/binder/`

If it's forced onto gke https://gke.mybinder.org/v2/gh/manics/jupyter-offlinenotebook/master?filepath=example.ipynb
`BINDER_LAUNCH_HOST	https://mybinder.org/`

bitnik:
No we don't have that setting. Btw repo of GESIS Binder is https://github.com/gesiscss/orc/tree/master/gesisbinder

manics:
Thanks! Has the behaviour of Gesis changed with this PR, or did it always have `https://notebooks.gesis.org/binder/` in all links?

betatim:
Could we start using `badge_base_url`? Or some other way that links generated by people who came via mybinder.org don't reference the particular cluster?

When we had the individual hosts exposed to normal users the result was people sharing links to individual clusters and then complaining when that cluster was down/unavailable.

bitnik:
@manics no, this PR didn't change behaviour of GESIS Binder. 

Links in buttons already had `mybinder.org` as host, because we replace it in notebook config (see https://github.com/gesiscss/orc/blob/master/gesisbinder/gesisbinder/files/etc/jupyter/jupyter_notebook_config.py#L16-L21). But `BINDER_LAUNCH_HOST` is https://notebooks.gesis.org/binder/ because we don't want to set `badge_base_url` config.

The reason why we don't use `badge_base_url` is because it changes badge and share links on binder form (actually this was the purpose of this setting). Maybe we should add a new config for `binder_launch_host` (which has by default same value as `badge_base_url`, or the other way around).

@arnim

manics:
short-term fix, ignore `BINDER_LAUNCH_HOST` and hard-code https://github.com/gesiscss/orc/blob/0388c1138417af359ef023d7ec845582bb41eed6/gesisbinder/gesisbinder/files/etc/jupyter/jupyter_notebook_config.py#L24 to `binder_launch_host = 'https://mybinder.org/'`?

arnim:
Hi @manics, hi @betatim 

> When we had the individual hosts exposed to normal users the result was people sharing links to individual clusters and then complaining when that cluster was down/unavailable.

No links exposed to normal users (i.e. via mybinder.org) contain any deployment-specific references to the Gesis deployment. In other words, no user coming from the mybinder redirector is exposed to any Gesis-specific UI so this should, as far as I can see, not be an issue for now. 

However, in the long run, there is obviously a lot of value in visible promoting a universal https://mybinder.org/* launch URL for all deployments. I'm promoting it and this is / should be our goal. 

As far as I can currently see, there are two things that are keeping us away from this. On the technical side, we have users that require a specific env (currently higher limits and access to DBs that are IP-restricted). We need a way for them to generate launch links that ensure that they launch on-premise. On the organizational side, we need to justify our commitment. Pointing out that if our hub goes down everything still works as expected is a huge plus for all of us but difficult to sell. Pushing some more formal association ahead could help with this and (with @betatim) we have already discussed this briefly.

manics:
@bitnik @arnim Would it help if the redirector on mybinder.org passed a parameter like `launch_host` to the backend binderhub, so if `launch_host` is set (and in a whitelist to prevent redirection attacks) that would be used to set `LAUNCH_HOST` inside the user pods, and if not it would default to the current behaviour?

bitnik:
@manics I think that would work.

manics:
Cool, I'll look into it

PR #1295, opened 2019-12-02
Switch to using docker hub for OVH as registry

betatim:
The OVH registry is currently unavailable so taking the chance to use Docker Hub. It seems to work well for the GESIS cluster so let's see how it goes here. Maybe we can use it for OVH or if we go back to using the OVH registry we can use this as a way to gain some experience with using Docker Hub as registry.


betatim:
cc @mael-le-gal

PR #1304, opened 2019-12-10
[WIP] Configuring cert-manager on GKE staging cluster

sgibson91:


config/staging.yaml
```diff
@@ -124,3 +141,9 @@ federationRedirect:
       versions: https://gke2.staging.mybinder.org/versions
     # unset the gesis entry
     gesis: null
+
+certmanager:
+  ingressShim:
+    defaultIssuerName: "staging"
```
consideRatio: This should reference a name of a ClusterIssuer resource, and will be the default value if no annotations like this one exist on the k8s ingress resource.

```yaml
      cert-manager.io/cluster-issuer: letsencrypt-staging
```

I think it should say letsencrypt-staging on this default value. With it in place, we can remove annotations on the ingress resources except the `kubernetes.io/tls-acme: 'true'` one that indicates for cert-manager that it should act on this ingress resource at all.

config/staging.yaml
```diff
@@ -22,6 +37,8 @@ binderhub:
         guarantee: 0.1
         limit: 0.5
     ingress:
+      annotations:
+        cert-manager.io/cluster-issuer: letsencrypt-staging
```
consideRatio: This ingress miss `kubernetes.io/tls-acme: 'true'`, and I think it should rely on the default values for cluster-issuer in the certmanager ingressShim configuration

sgibson91:
Closing and starting afresh

PR #1307, opened 2019-12-14
Set X-Binder-Launch-Host header or binder_launch_host parameter

manics:
Set the `X-Binder-Launch-Host` header when pass-through proxying or the `binder_launch_host` query parameter when redirecting.

See https://github.com/jupyterhub/binderhub/pull/1031 for the corresponding BinderHub PR.

The last step of this is to add
```
    def badge_base_url(handler):
        launch_host = (
            handler.request.headers.get('X-Binder-Launch-Host', '') or
            handler.get_argument('binder_launch_host', ''))
        if launch_host and not launch_host.endswith('/'):
            launch_host += '/'
        if launch_host == 'https://mybinder.org/':
            return launch_host
        return ''
    c.BinderHub.badge_base_url = badge_base_url
```
to the Gesis BinderHub configuration

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -14,7 +14,7 @@ from tornado.ioloop import IOLoop
 from tornado.log import enable_pretty_logging, app_log
 
 from tornado.httpclient import AsyncHTTPClient, HTTPError, HTTPRequest
-from tornado.httputil import HTTPHeaders
+from tornado.httputil import HTTPHeaders, url_concat
 from tornado.web import RequestHandler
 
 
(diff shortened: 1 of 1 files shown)


choldgraf:
should this wait for https://github.com/jupyterhub/binderhub/pull/1031 before merging? it seems reasonable to me though I don't fully understand the implications since I'm not following that PR closely

bitnik:
yes, you are right, this has no effect if https://github.com/jupyterhub/binderhub/pull/1031 is not merged

manics:
@choldgraf If you're happy with the code here I think the final signoff for this and jupyterhub/binderhub#1031 together should be for @bitnik since it's to help customise the Gesis mybinder links. It should be safe to merge them in either order since neither will have any effect unless there's also a binderhub config change, and that's only needed for Gesis.

choldgraf:
Sounds good - in that case I am +1 when @bitnik is 1

bitnik:
I think we can merge this.

PR #1317, opened 2020-01-16
[RVW] Add a version check before helm upgrade happens

sgibson91:
This PR adds a helm version check into the `setup_helm` function in `deploy.py`. This will hopefully avoid the issue I got myself into where I had accidentally upgraded helm on staging to v2.14 when travis uses v2.11 - the mismatch causes helm upgrades to fail. It also implements some checks to see if the script is being run locally or not. Most often, the helm version problem occurs when the script is run locally so we ask the user if they're really sure they want to do this. We also add a `--local` flag so this input step is skipped.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -130,6 +130,7 @@ def setup_helm(release):
         )
     elif (client_version == "v2.11.0") and (client_version == server_version):
         # All is good! Perform normal helm init command.
+        # We use the --client-only flag so that the Tiller installation is not affected.
         subprocess.check_call(['helm', 'init', '--client-only'])
     else:
         # This is a catch-all exception. Hopefully this doesn't execute!
--- a/deploy.py
+++ b/deploy.py
@@ -107,11 +107,10 @@ def setup_helm(release):
     server_version = subprocess.check_output(server_helm_cmd
         ).decode('utf-8').split(":")[1].split("+")[0].strip()
 
-    print(
-        "Client version: {}, Server version: {}".format(
-            client_version,
-            server_version,
-        )
+    print(BOLD + GREEN +
+        f"Client version: {client_version}, Server version: {server_version}" +
+        NC,
+        flush=True
     )
 
     # Now check if the version of helm matches v2.11.0 which travis is expecting
--- a/deploy.py
+++ b/deploy.py
@@ -127,7 +127,7 @@ def setup_helm(release):
         # to bring the server side back to v2.11.0
         raise Exception(
             "Helm client and server versions do not match. Performing a force upgrade often resolves this issue." +
-            "Please run the following command and re-execute this script.\n\n" +
+            "Please run the following command and re-execute this script.\n\n\t" +
             "helm init --upgrade --force-upgrade"
         )
     elif (client_version == "v2.11.0") and (client_version == server_version):
--- a/deploy.py
+++ b/deploy.py
@@ -265,8 +265,34 @@ def main():
 
     args = argparser.parse_args()
 
-    # Check if the script is being run on travis or hand the --local flag set
-    if (cwd.startswith('/home/travis')) and (not args.local):
+    # Check if the local flag is set
+    if not args.local:
+        # Check if the script is being run on travis
+        if  not (cwd.startswith('/home/travis')):
+            # Catch the case where the script is running locally but the --local flag
+            # has not been set. Check that the user is sure that they want to do this!
+            print(
+                "You do not seem to be running on Travis but have not set the --local flag."
+            )
+
+            # Use regex to match user input
+            regex_no = re.compile("^[n|N][o|O]$")
+            regex_yes = re.compile("^[y|Y][e|E][s|S]$")
+            response = input("Are you sure you want to execute this script? [yes/no]: ")
+
+            if regex_no.match(response):
+                # User isn't sure - exit script
+                print("Exiting script.")
+                sys.exit()
+            elif regex_yes.match(response):
+                # User is sure - proceed
+                pass
+            else:
+                # User wrote something that wasn't "yes" or "no"
+                raise ValueError(
+                    "Unrecognised input. Expecting either yes or no."
+                )
+
         # script is running on travis, proceed with auth and helm setup
         if args.cluster == 'binder-ovh':
             setup_auth_ovh(args.release, args.cluster)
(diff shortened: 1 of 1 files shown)


deploy.py
```diff
@@ -96,7 +96,43 @@ def setup_auth_gcloud(release, cluster):
...
+        )
+    elif (client_version == "v2.11.0") and (client_version != server_version):
+        # The correct local version of helm is installed, but the server side
+        # has previously accidentally been upgraded. Perform a force-upgrade
+        # to bring the server side back to v2.11.0
+        print(
+            "Helm client and server versions do not match. Performing a force upgrade."
+        )
+        subprocess.check_call(["helm", "init", "--upgrade", "--force-upgrade"])
+    elif (client_version == "v2.11.0") and (client_version == server_version):
```
betatim: In this case we have helm locally and on the remote machine. Naively I'd expect `helm init --upgrade` to be a no-op in this case. Do we know if it is?

deploy.py
```diff
@@ -96,7 +96,43 @@ def setup_auth_gcloud(release, cluster):
...
+            "Please change your installation and try again.\n" +
+        )
+    elif (client_version == "v2.11.0") and (client_version != server_version):
+        # The correct local version of helm is installed, but the server side
+        # has previously accidentally been upgraded. Perform a force-upgrade
+        # to bring the server side back to v2.11.0
+        print(
+            "Helm client and server versions do not match. Performing a force upgrade."
+        )
+        subprocess.check_call(["helm", "init", "--upgrade", "--force-upgrade"])
```
betatim: Wondering if we should do this or print a message telling the user "Woah something weird must have happened, maybe you want to run ...blah... but mostly the universe is not in the state I was expecting so I am bailing out". Mostly I am worried that we will end up in a state we should be in and then automation starts doing more "unusual stuff" which auses more confusion.

betatim:
I like it, though reading the code made me wonder if trying to automate this is going to be something that we regret later. The "root cause" was that the script got executed locally and then we were in a weird state.

Maybe we should add a check to the script that stops you from running it locally or tells you to add some command-line flags that turn off things like auth and setup. I try to never run the script locally and when I do I first edit it to turn off setup and auth stuff as they'd screw with the local setup I already have.

WDYT?

sgibson91:
The downside of saying "don't run this script locally" is that, for example when I was setting it up the Turing cluster, it was helpful to know that this script worked without having to merge the PR (at least on my cluster). I don't know if you'd prefer lots of tiny merges, or knowing that when you do hit merge chances are high that it will "just work" (fateful last words).

It would be cool if we could find a way of running some tests on CI at the PR - maybe just to staging. My experience so far is that this requires a second CI vendor though.

deploy.py
sgibson91: I'm happy to make this a "This is likely to be the situation and this command is likely what will fix it. Please execute and re-run the script." message if that's easier.

deploy.py
sgibson91: Do you mean just `helm init` instead of `helm init --upgrade`? The `--upgrade` flag is definitely what's causing the problem of mismatched versions if it's run locally. I usually do `helm init --client-only` so my client is configured to match the server, but the server isn't reconfigured.

sgibson91:
Oh, I also modified the script to point to my second (correct) installation of helm - but I inevitably would forget to do this if I checked the script out so I wouldn't commit it. I'd rather have a big warning spring up saying "you forgot to switch helm versions and are about to break everything".

betatim:
What I was thinking is maybe more a "are you sure you want to be running this locally??11!! Did you make the changes you normally make??!?!" than preventing people from running it.

What about a check that stops the script from running locally unless you add `--local` to the command-line flags? If that flag is present the script would skip setup and auth steps and just run the deploy steps?

---

Tests as part of a PR before you merge it would be awesome. Super tricky to do, I think. At least I imagine you'd ned to setup a new cluster per PR or some such? Then how do you tear it down again. It would be super cool though.

Maybe some tests that don't need deploying to a full cluster would already be helpful?

sgibson91:
Yeah, I was thinking about a `--local` flag. Maybe we can do some detection for the travis directory that automatically enables auth and helm stuff and have the default run to be assume it's running locally and skip those steps?

---

So I know that GitHub Actions has some cool integrations with Azure to automate deployments and stuff. We could piggy back my BinderHub-deploy repo to automatically spin up a cluster, run tests and teardown. We could package it as a container and have the GitHub Action deploy the container. The problem is that GitHub Actions are still really new and documentation is scarce. cc/ @trallard as this is her new hobby area! 😂

sgibson91:
One thing I've done is set up an Azure pipeline to lint and validate my helm chart in PRs and on master: https://github.com/alan-turing-institute/hub23-deploy/blob/master/lint-pipeline.yml

deploy.py
sgibson91: Addressed in commit 2d9d5f6

deploy.py
sgibson91: Addressed in commit 2c0d897

deploy.py
sgibson91: Addressed in commit 2d9d5f6

deploy.py
```diff
@@ -96,7 +96,45 @@ def setup_auth_gcloud(release, cluster):
...
+            "You are not running helm v2.11.0 which is the version our continuous deployment system uses.\n" +
+            "Please change your installation and try again.\n"
+        )
+    elif (client_version == "v2.11.0") and (client_version != server_version):
+        # The correct local version of helm is installed, but the server side
+        # has previously accidentally been upgraded. Perform a force-upgrade
+        # to bring the server side back to v2.11.0
+        raise Exception(
+            "Helm client and server versions do not match. Performing a force upgrade often resolves this issue." +
+            "Please run the following command and re-execute this script.\n\n" +
```
sgibson91:
```suggestion
            "Please run the following command and re-execute this script.\n\n\t" +
```

deploy.py
```diff
@@ -96,7 +96,45 @@ def setup_auth_gcloud(release, cluster):
...
+            "\thelm init --upgrade --force-upgrade"
```
sgibson91:
```suggestion
            "helm init --upgrade --force-upgrade"
```

sgibson91:
Ok, I have implemented the following changes:

- Raise an error suggesting the user perform a force upgrade rather than performing it automatically
- Use `helm init --client-only` flag instead of `--upgrade` to configure the helm config
- Apply a `--local` flag that will skip the auth and helm setup functions when used
- Check whether the script is running on Travis or not. Catch the case where the script is _not_ running on travis and the user has _not_ set the `--local` flag. Ask them if they're really sure that that's what they want to do.

deploy.py
sgibson91: FYI, I think `helm init` by itself tries to install Tiller.

trallard:
>The problem is that GitHub Actions are still really new and documentation is scarce. cc/ @trallard as this is her new hobby area! 😂 

😂 precisely - if there is any unclear docs or parts missing do let me know and I will ping the team directly. 
At the moment we have a limited collection of Azure actions in this repo <https://github.com/Azure/actions#deploy-to-kubernetes> but I am sure I /we/the team can add more or expand the existing ones to support a better deployment of all things Jupyter/Binder

deploy.py
minrk: My only feedback here is to make the helm version a variable so it's reused, and possibly default to using a $HELM_VERSION environment variable, which we can set and use in .travis.yml.

That way updating helm shouldn't need more than one place to update

betatim:
Sorry for getting distracted and losing track of this. LGTM.

deploy.py
sgibson91: PR to address this point #1346

PR #1320, opened 2020-01-20
Actually enabling HTTPS on prometheus and static Turing A records

sgibson91:


config/turing.yaml
```diff
@@ -161,6 +162,7 @@ nginx-ingress:
 static:
   ingress:
     annotations:
+      cert-manager.io/cluster-issuer: letsencrypt-prod
       kubernetes.io/ingress.class: nginx
       kubernetes.io/tls-acme: 'false'
```
sgibson91:
```suggestion
      kubernetes.io/tls-acme: 'true'
```

PR #1327, opened 2020-01-28
Merge pull request #1327 from betatim/user-survey-jan-2020

betatim:
This PR links to the same user survey we used in June 2019. I think it would be nice to run it again to collect feedback on what people are thinking. Last time a big message was "make it faster". Let's find out what people are thinking now.

The editable form is https://docs.google.com/forms/d/1mY4rE6PgvpbLlw9bK61Bp4-XdPNqAIBsT_Ly55GYDvE/edit (if you can't access it tell me your Google account and I'll add you).

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -74,7 +74,7 @@ binderhub:
       per_repo_quota_higher: 200
 
       banner_message: |
-        <div style="text-align: center;">Thanks to <a href="https://cloud.google.com/">Google Cloud</a>, <a href="https://www.ovh.com/">OVH</a>, <a href="https://notebooks.gesis.org">GESIS Notebooks</a> and the <a href="https://turing.ac.uk">Turing Institute</a> for supporting us! 🎉</div>
+        <div style="text-align: center;">Give us feedback on how mybinder.org is doing and what to improve: <a href="https://docs.google.com/forms/d/e/1FAIpQLSd3fiLCMuQsc48_ga2q_FJFqgFcVUkie7RBex4DtzzOyyNWHg/viewform" target="_blank">the mybinder.org user survey</a> (it is only three questions!). Thank you!</div>
       about_message: |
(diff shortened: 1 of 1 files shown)


choldgraf:
I think this is a good idea - one quick thought: what do folks think about adding one extra question along the lines of: "For what purpose do you use mybinder.org (e.g., reproducible publishing, teaching)"

betatim:
I'd be ok with swapping a question but I wouldn't add one more.

betatim:
For example the question about which repo people launched last. I didn't get anything interesting from the responses to that and the use case question seems more interesting.

choldgraf:
That makes sense to me - I think that we can also infer "which repositories are being launched" by our data archive more easily than we can infer "why are these repositories being used", so I feel like that's a better use of our "asking a human" space. WDYT?

betatim:
I added a "what do you use this for" question with some topics that came to my mind. What should we add or remove and which options are unclear?

choldgraf:
I changed one option about teaching to be a bit easier to understand (IMO) and added a question about collaborating with a team. I think it looks good to me

betatim:
Nods, the school vs university teaching labels weren't very good. Do people outside the US know what K-12 teaching is? I have only ever seen it used by people in the US and asking my UK based friends they didn't really know what it means. Pretty sure if you aren't a native English speaker you won't know (I also don't really know what it means though I can guess from the context here).

How about using secondary and tertiary education as labels instead?

choldgraf:
That's a great point. Or just "university" vs secondary?

betatim:
Done! Avoiding latin (I think) is a good idea :)

Merge?

consideRatio:
Nice this LGTM!

betatim:
As people started saying "LGTM" but didn't press merge I pressed it :D

consideRatio:
I had not spent enough time on this to dare be the presser ;p

PR #1329, opened 2020-01-28
Manual update of binderhub helm chart

betatim:
Henchbot is out of action at the moment so this is a manual upgrade of the binderhub helm chart version.

### DIFF
--- a/mybinder/requirements.yaml
+++ b/mybinder/requirements.yaml
@@ -19,5 +19,5 @@ dependencies:
    tags:
      - certmanager
  - name: binderhub
-   version: 0.2.0-072.544c0b1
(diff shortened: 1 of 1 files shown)


betatim:
https://github.com/jupyterhub/binderhub/pull/1043
https://github.com/jupyterhub/binderhub/pull/1037

are the two PRs that have happened since we last updated I think.

consideRatio:
If hencbot went down, i suspect the chartpress upgrade. :(

I have written lots of stuff in chartpress' README file about how it behaves now, in complience with helm3 assuming tags are major.minor.patch only at least...

Yikes so much trouble caused by this :/

PR #1336, opened 2020-02-06
preserve host weights when using rendezvous routing

minrk:
rendezvous routing was ignoring our host weights and routing repos uniformly across members.

This algorithm is still routing each repo to one host 100% of the time, so if a repo gets a traffic spike, only one member of the federation will get hit.

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -83,20 +83,22 @@ def blake2b_hash_as_int(b):
     return int.from_bytes(blake2b(b, digest_size=8).digest(), "big")
 
 
-def rendezvous_rank(buckets, key):
+def rendezvous_rank(buckets, weights, key):
     """Rank the buckets for a given key using Rendez-vous hashing
 
     Each bucket is scored for the specified key. The return value is a list of
     all buckets, sorted in decreasing order (highest ranked first).
     """
     ranking = []
-    for bucket in buckets:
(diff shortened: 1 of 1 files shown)


consideRatio:
A function signature is updated to include weights, but how it is called hasn't been updated yet.

https://github.com/jupyterhub/mybinder.org-deploy/blob/c972335b606b0c69b93ac4eb9fe884404f701d04/images/federation-redirect/app.py#L190

To clarify as I was a bit uncertain what "host" referred to, even though I think it refers to for example ovh.mybinder.org or gke.mybinder.org. Is the following a correct summary of the situation?
mybinder.org spreads traffic to various members of the federation, and with rendezvous routing it is now ignoring weights on how to distribute this traffic to the federation members. This PR aims to fix this.

minrk:
@consideRatio good catch, fixed.

betatim:
What problem do we fix by using the weights?

It will change what fraction of repositories are assigned to a cluster. Right now a repository has an equal chance of being assigned to any of the clusters we have available.

In the random assignment case the weights influence the chances of a launch request to be assigned to a particular repo. With rendezvous routing every launch request will be assigned to the same cluster, weights or no weights.

minrk:
Reading about weighted-rendezvous-hashing (WRH), this doesn't actually apply the weights correctly. Need to cast to int and divide by log, etc.

Closing here, and moving to more general discussion in #1066

PR #1337, opened 2020-02-06
use github access token instead of client id and secret

### LINKED ISSUE #1335: Using `client_id` and `client_secret` as query parameters (deprecated API)
I just received the following email from GitHub. Does this ring a bell with anyone? I am surprised to only get one for staging and not also for prod. I had a quick grep through the BinderHub code and couldn't find a totally obvious location where this is happening. Thoughts and pointers would be welcome

---

Hello there!

On February 5th, 2020 at 17:27 (UTC) your application (MyBinder Staging) used its `client_id` and `client_secret` (with the User-Agent BinderHub) as part of a set of query parameters to access an endpoint through the GitHub API:

https://api.github.com/repositories/98230295/commits/fa84f12

Please use Basic Authentication instead as using OAuth credentials in query parameters has been deprecated.

Depending on your API usage, we'll be sending you this email reminder at most once every 3 days.
Just one URL that was accessed with a User-Agent combination will be listed in the email reminder, not all.

Visit https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters for more information.

Thanks,
The GitHub Team

---

minrk:
access with client id + secret is deprecated, apparently

closes #1335


consideRatio:
Reviewing binary changes on github :) Looks hard to me!

PR #1339, opened 2020-02-06
enable event log on turing

minrk:
required two manual out-of-band steps:

1. create service account in google cloud console with Logs Writer role,
2. create a kubernetes secret containing the key for that service account

This is copying what is in the configuration for other non-GKE clusters

In the future, I think we should try to:

1. fully document the out-of-band steps needed for fully adding a cluster to the federation
1. store the service-account credentials in the encrypted secrets in this repo
1. deliver them to the cluster as secrets from this repo, so that the updates can be fully automatically with PRs here

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -8,8 +8,6 @@ letsencrypt:
   contactEmail: drsarahlgibson@gmail.com
 
 binderhub:
-  extraConfig:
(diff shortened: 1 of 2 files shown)


sgibson91:
Definitely love the idea and documenting this process!

Issue #1344, opened 2020-02-13
CORS issue causing event stream error on gesis


minrk:
Builds on gesis are failing with "Failed to connect to event stream" due to a CORS issue. The request to open the event stream is failing with:

> [Error] Failed to load resource: Cross-origin redirection to https://notebooks.gesis.org/binder/build/gh/TomAugspurger/idp-results/master?binder_launch_host=https%3A%2F%2Fmybinder.org%2F denied by Cross-Origin Resource Sharing policy: Cannot use wildcard in Access-Control-Allow-Origin when credentials flag is true. (master, line 0)
> [Error] EventSource cannot load https://gesis.mybinder.org/build/gh/TomAugspurger/idp-results/master?binder_launch_host=https%3A%2F%2Fmybinder.org%2F due to access control checks.
> [Error] Failed to construct event stream – Event {isTrusted: true, type: "error", target: EventSource, …}
Event {isTrusted: true, type: "error", target: EventSource, currentTarget: EventSource, eventPhase: 2, …}Event
	(anonymous function) (bundle.js:21152)

I'm not sure what setting is different on gesis that causes this. We should probably disable gesis until this works out, since all builds will appear to fail (the builds and launches do not appear to actually be affected, only the event stream)

bitnik:
hi @minrk , can you help me here? I am not sure if I really understand the error. How can I produce it? So I can investigate what is wrong in our cluster.

minrk:
I'm trying to understand the root cause myself. Binder deployments have `Access-Control-Allow-Origin: *` to allow anyone to connect. However, there is a conflict between this option and `Access-Control-Allow-Credentials: true` and/or sending requests `withCredentials`.

It is 
There should be tw

minrk:
What I don't understand is what is different about gesis that is triggering this, why the credentials flag would be different.

I think the best way to test this would be to add gesis to the *staging* federation so that you can try builds there. It's hard to reproduce when gesis is temporarily out of rotation.

manics:
@bitnik Did anything change with your authenticated binderhub that might have inadvertently affected the public one?

minrk:
You can reproduce this in the js console in a browser with:

```javascript
evt = new EventSource("https://mybinder.org/build/gh/TomAugspurger/idp-results/master", {withCredentials: true})
evt.close()
```

[...]

PR #1345, opened 2020-02-13
remove gesis from the cluster

minrk:
while we figure out a fix for #1344


minrk:
already done in #1347

PR #1346, opened 2020-02-13
[MRG] Replace hard-coded helm version with environment variable in deploy.py

sgibson91:
This PR will replace the hard-coded helm version in deploy.py with an environment variable that can be set within .travis.yml. This will mean that when the helm version is bumped, it only needs to be updated in one place.

TODOS:

- [x] Update deploy.py to include a HELM_VERSION environment variable
- [x] Update .travis.yml to extract the helm version and parse to deploy.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,6 @@ before_deploy:
 - |
   # Stage 2, Step 2: Set up helm!
   helm init --client-only
-  echo ${TRAVIS_HELM_VERSION}
   helm repo add jupyterhub https://jupyterhub.github.io/helm-chart
   helm repo add jetstack https://charts.jetstack.io
   helm repo update
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,7 @@ before_deploy:
   # Doing this here, since git crypt doesn't like unclean checkouts
 - |
   # Stage 4, Step 1: Deploy to staging
-  python ./deploy.py staging staging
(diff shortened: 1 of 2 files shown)


deploy.py
```diff
@@ -15,6 +15,10 @@
 HERE = os.path.dirname(__file__)
 ABSOLUTE_HERE = os.path.dirname(os.path.realpath(__file__))
 
+# Get helm version environment variable
+HELM_VERSION = os.get_env("HELM_VERSION", None)
+if HELM_VERSION is None:
+    raise Exception("HELM_VERSION environment version must be set")
```
sgibson91:
```suggestion
    raise Exception("HELM_VERSION environment variable must be set")
```

.travis.yml
```diff
@@ -61,6 +61,8 @@ before_deploy:
 - |
   # Stage 2, Step 2: Set up helm!
   helm init --client-only
+  TRAVIS_HELM_VERSION=`helm version -c --short | sed 's/.*v\([0-9\.]*\).*/\1/'`
+  echo ${TRAVIS_HELM_VERSION}
```
minrk: Can we use $HELM_VERSION to set the version in the the "Install helm" step? on line 45 above? That way we don't need to query it here.

sgibson91: Done in 0356912 ✨

minrk: Thanks!

PR #1362, opened 2020-02-20
[WIP] Deploy cert-manager on staging GKE cluster

sgibson91:
This PR is editing the staging config to use cert-manager instead of kube-lego to request HTTPS certificates from Let's Encrypt.

Contact email is being switched to binder-team@googlegroups.com

addresses #1148


config/staging.yaml
```diff
@@ -25,6 +36,8 @@ binderhub:
     ingress:
       hosts:
         - hub.gke.staging.mybinder.org
+      annotations:
+        kubernetes.io/tls-acme: "true"
       tls:
         - secretName: kubelego-tls-jupyterhub-staging
```
consideRatio: A systematic naming of the secrets, currently there is both kubelego and certmanager in the name, staging and not staging, and tls before and after the service name ;)

consideRatio:
I don't think I have access to the GCP project for mybinder staging - could I get an invite?

config/staging.yaml
sgibson91: I know, but I didn't know what effect renaming the already existing secrets would have 😕

minrk:
@consideRatio added you to binder-staging

config/staging.yaml
sgibson91: Done in c996eb4

betatim:
I took a look at this over the weekend and I think the staging cluster contains leftovers from attempts to install cert-manager via manifests, helm, into a different namespace, into the same namespace, as well as using different versions of cert-manager.

Trying to remove all the namespace and cluster wide things that have been installed seems to be nearly impossible. At least I gave up after a few hours of trying to find things.

The steps required to get cert-manager (v0.12, we aren't using v0.13 which is the default in the cert-manager docs) deployed should be:
```
# run the following commands from your local machine and authenticate as your personal account
# select the staging namespace on the staging cluster
# install the cert-manager CRDs
$ kubectl apply --validate=false -f https://raw.githubusercontent.com/jetstack/cert-manager/release-0.12/deploy/manifests/00-crds.yaml
# enable the cert-manager chart as a dependency of mybinder via the tag
# deploy via travis
```
however this fails with errors about certain objects not being found. I suspect this is because they were installed previously with different credentials??

To use the shim or not, what exact annotations we need, the indentation change in this PR, etc are all things we can tackle once we have a working install of cert-manager.

Two weeks ago I setup (on a GKE cluster) a BinderHub with nginx-ingress, cert-manager v0.12, and helm v2 which "just worked". So I am confident that this recipe should work and we don't need to make additional changes/configuration updates etc.

I just realised that there is one difference between the setup I did two weeks ago and ours: I think the initial `helm upgrade --install` for the other setup was run on my laptop while I was auth'ed with my personal account. Not via CI which is auth'ed as "travis deployer". I don't think this should make a difference as tiller does all the work ... so any difference between my rights and travis-deployer shouldn't matter. Can we confirm/deny this?

At this point it would be good if we could reduce the "let's just try stuff" attempts as the cluster is already in a weird state with remnants of previous install attempts.

Unless we have an idea how to clean up our current staging cluster I would propose that we have to setup a new one from scratch, configure it to use kubelego (we won't get good certs but it is the closest we can get to the current state of prod), then attempt to install cert-manager via the steps above. This is a bunch of work but we need to have a testbed for figuring out if we can successfully transition our prod cluster from kubelego to cert-manager or not.

sgibson91:
> I don't think this should make a difference as tiller does all the work ... so any difference between my rights and travis-deployer shouldn't matter. Can we confirm/deny this?

I'm pretty certain installing the CRDs fails when you're auth'd as travis-deployer

betatim:
> I'm pretty certain installing the CRDs fails when you're auth'd as travis-deployer. So we wouldn't be able to have the install of CRDs in CI

Nods. I was thinking we'd have to always install CCRDs manually. The open question is if `helm upgrade ...` as betatim or travis-deployer has different permissions. My gut feeling says no but unsure.

sgibson91:
> ... the indentation change in this PR...

I'm happy for those particular changes to be reverted, it was mostly me figuring out why the helm templating was failing before I realised chartpress needed to be run. I pushed them because I wanted to pull them into my gcloud shell where I was actually interacting with the cluster.

How far to a clean cluster does `kubectl delete crds --all` get us?

betatim:
I removed the cert-manager CRDs by hand. Other CRDs are installed as well on the cluster that we need so we can't wipe all of them. It seems like the problem is various other "stuff" that gets created :-/

betatim:
Closing this as we now have cert-manager on staging and prod. OVH is a special case that we need to look at (at least I think we did something special).

#1415

PR #1363, opened 2020-02-20
make all redirections to gesis in staging federation

bitnik:
for testing https://github.com/jupyterhub/mybinder.org-deploy/issues/1344

### DIFF
--- a/config/staging.yaml
+++ b/config/staging.yaml
@@ -3,6 +3,7 @@ projectName: binder-staging
 binderhub:
   config:
     BinderHub:
+      pod_quota: 0
       hub_url: https://hub.gke.staging.mybinder.org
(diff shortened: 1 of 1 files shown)


betatim:
The size of the weight doesn't do anything (as long as it is above zero) when we are using rendezvous routing.

bitnik:
thats true, sorry. what about making all redirections to gesis?

betatim:
The goal of this PR is to send all traffic from the staging federation to Gesis to make debugging the CORS issue easier.

One thing we need to keep in mind is that if all launches fail while we are debugging this then we won't be able to deploy to production. I think the risk is small because the testing we do during deploying doesn't care about CORS.

PR #1367, opened 2020-02-22
[RVW] Migrate Turing cert-manager to using ingressShim config instead of annotations

sgibson91:
This change is at the advice of @consideRatio, see https://github.com/jupyterhub/mybinder.org-deploy/issues/1148#issuecomment-588651294

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -36,7 +36,6 @@ binderhub:
       - binder.mybinder.turing.ac.uk
       - turing.mybinder.org
     annotations:
-      cert-manager.io/cluster-issuer: letsencrypt-prod
       kubernetes.io/ingress.class: nginx
       kubernetes.io/tls-acme: "true"
       https:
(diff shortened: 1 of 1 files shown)


config/turing.yaml
```diff
@@ -212,3 +207,11 @@ gcsProxy:
   buckets:
     - name: mybinder-staging-events-archive
       host: archive-analytics-staging-mybinder.turing.10.0.0.1.xip.io
+
+cert-manager:
+  webhook:
+    enabled: false
```
betatim: Why did we turn this off?

betatim: This is the comment that lead to it being disabled https://github.com/jupyterhub/mybinder.org-deploy/issues/1148#issuecomment-590047510

consideRatio: https://github.com/jupyterhub/mybinder.org-deploy/issues/1148#issuecomment-590047510

sgibson91: Yes sorry, I left this info in the commit message rather than the PR

PR #1380, opened 2020-03-04
Update doc requirements.txt file

betatim:
Update the versions referenced in the requirements.txt to match the environment.yml
and pin to a specific sphinx theme version.


choldgraf:
we'll need to remember to start using the pypi version of the theme whenever it releases for the first time

betatim:
Nods.

Sad times, I was hoping that this would also fix the failing builds on RTD but no dice https://readthedocs.org/projects/mybinder-sre/builds/ ;(

choldgraf:
huh...that is a weird one, I've never seen a memory consumption error like that

PR #1401, opened 2020-04-08
Migrate Turing to Docker Hub from Azure Container Registry for image storage

sgibson91:
This PR will switch Turing from using an Azure Container Registry to Docker Hub for storing images. This will reduce  maintenance complexity and running cost.


betatim:
I made one edit to the secret based on what is in the OVH config. Right now trying to figure out how to push that change and then deploying this.

PR #1407, opened 2020-04-12
Switch to prod Let's Encrypt instance

betatim:
It seems like cert-manager is working and staging.mybinder.org now has a certificate from "staging" let's encrypt. This PR switches us to the prod LE.


choldgraf:
cc @consideRatio who may start working a bit on security stuff in BinderHub in case he isn't pinged by this

PR #1416, opened 2020-04-19
[MRG] Tell helm to clean up on failed deploys

### LINKED ISSUE #1409: Add `--cleanup-on-fail` to our `deploy.py`
From @consideRatio at the meeting today:

* [name=Erik] General notes on helm2 and helm3 upgrades:
        * Always use `--cleanup-on-fail`.
        * Don't use `--wait`, or if you do, ensure to not abort!

We should enable this for our `deploy.py` in this repository.

betatim:
This fixes a large class of problems where deploying (again) fails. The
problem is that a previous deploy failed and left tiller thinking that
the cluster is in a different state than it really is. Adding the
cleanup-on-fail flag tells helm to attempt to clean up. It is not
perfect but better than nothing.

closes #1409

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -208,6 +208,7 @@ def deploy(release):
         'mybinder',
         '--force',
         '--wait',
+        '--cleanup-on-fail',
         '--timeout', '600',
         '-f', os.path.join('config', release + '.yaml'),
         '-f', os.path.join('secrets', 'config', 'common.yaml'),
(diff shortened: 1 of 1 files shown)


consideRatio:
It is okay in this CI environment I think, but I recommend against it when used by humans. Hmm but perhaps it would be better not used here as well, because this is a CI job operating on the external GKE clusters? Then an aborted CI job can act as a terminal interrupt by a human, causing resources to be left unmanaged and then later cause conflicts.

Also, with wait, you will experience a difference between helm2 and helm3. Helm 3 requires the timeout to be specified with 5m0s instead of 300 for example.

PR #1418, opened 2020-04-19
Remove --wait and move --timeout from helm to kubectl

### LINKED ISSUE #1417: Stop using --wait in helm upgrade
We still use `--wait` in our CI env, and I think it poses a unwanted risk that makes a cancelled CI job potentially lead to a unwanted state in the GKE cluster with stranded resources that will cause conflicts in future upgrades.

I think a good practice would instead be to not ask helm to wait for various deployments etc to become ready, but do it manually using kubectl after `helm upgrade`, which is something I think we may already be doing.

## The `helm upgrade` call
https://github.com/jupyterhub/mybinder.org-deploy/blob/20fa5612a3c9af9277ee5f6c63180a464b1a9b2f/deploy.py#L204-L216

## The subsequent manual wait
https://github.com/jupyterhub/mybinder.org-deploy/blob/20fa5612a3c9af9277ee5f6c63180a464b1a9b2f/deploy.py#L237-L242

## Related issue
Also associated, there is a difference between helm2 and helm3 with the `--timeout` flag associated with the `--wait` flag. Helm 3 require the timeout to be specified like `5m0s` instead of `300` and will crash if not specified correctly.

## Suggested action
- Remove the `--wait` flag of `helm upgrade and rely on the explicit kubectl wait below instead.
- Remove the `--timeout` flag of `helm upgrade` and rely on its default value of 300 / 5m0s, as it make us avoid manage this specifically based on what helm version we use, as 300 is only valid for helm 2, and 5m0s is only valid for helm 3.
  The reduction of time is fine, we will now only wait for the helm upgrade hooks as compared to those and then for all deployments to get ready.
- Add a `--timeout` flag to the `kubectl rollout status` command, specifying for example 5 minutes for the deployments and daemonsets to get ready. We should make it a time that allow images to get pulled, containers to start and get running and considered ready by readiness probes. I think 5 minutes is plenty. The default value is to wait for eternity though, so let's not do that.

consideRatio:
Using --wait with helm is risky, because any cancellation during the
wait can let us fail to do the --cleanup-on-fail, which is the failsafe
to avoid getting stranded resources that are unmanaged according to helm
which can cause subsequent upgrade failures because "<...> already
exists."

Closes #1417 which describes the issue further.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -218,21 +218,14 @@ def deploy(release):
 
     # Explicitly wait for all deployments and daemonsets to be fully rolled out
     print(BOLD + GREEN + f"Waiting for all deployments and daemonsets in {release} to be ready" + NC, flush=True)
-    deployments = subprocess.check_output([
+    deployments_and_daemonsets = subprocess.check_output([
         'kubectl',
         '--namespace', release,
-        'get', 'deployments',
+        'get', 'deployments,daemonsets',
         '-o', 'name'
     ]).decode().strip().split('\n')
 
-    daemonsets = subprocess.check_output([
-        'kubectl',
-        '--namespace', release,
-        'get', 'daemonsets',
-        '-o', 'name'
-    ]).decode().strip().split('\n')
-
-    for d in deployments + daemonsets:
+    for d in deployments_and_daemonsets:
         subprocess.check_call([
             'kubectl', 'rollout', 'status',
             '--namespace', release,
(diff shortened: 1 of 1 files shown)


sgibson91:
LGTM!

betatim:
Thanks for infrastructuring :) Let's see how this goes

consideRatio:
@betatim @sgibson91 this failed because kubectl was older in the CI system. I used kubectl 1.18.0 when I concluded this flag existed and were useful.

PR #1419, opened 2020-04-19
CI: updating kubectl so expected flags exists

consideRatio:
#1418 failed when merged into staging because kubectl was outdated and didn't contain the `--timeout` flag for the `kubectl rollout status` command in the version we currently use.

I think it should work fine to use a very modern client no matter what the k8s clusters we work against.


consideRatio:
Thanks for your quick review/merge @sgibson91!

It seems to have worked ;D

PR #1426, opened 2020-05-04
[MRG] Move the turing cluster to the staging federation

### LINKED ISSUE #1425: Turing cluster: Scheduled maintenance 2020-05-04
Hi all,

Azure informed me that there will be some scheduled maintenance affecting the Turing federation cluster on **4th May 2020**. The plan is to move Turing to staging for that day to experiment with the federation redirector and if it's clever enough to work out when the cluster is down :-)

betatim:
The Turing cluster will be maintained/unavailable so we are moving it
to the staging federation. We want to see if the federation redirect/health checks are good enough that the cluster will be automatically removed from the set of healthy hosts during the maintenance window or not.

closes #1425

Leaving this for @sgibson91 to merge. I think this is what needs changing, at least it is what I'd change. If you spot something else let me know and we can add it.

### DIFF
--- a/config/staging.yaml
+++ b/config/staging.yaml
@@ -134,8 +134,11 @@ federationRedirect:
       weight: 1
       health: https://gke2.staging.mybinder.org/health
       versions: https://gke2.staging.mybinder.org/versions
-    # unset the gesis and turing entries
-    gesis:
-      weight: 0
     turing:
+      url: https://turing.mybinder.org
+      weight: 1
+      health: https://turing.mybinder.org/health
+      versions: https://turing.mybinder.org/versions
+    # unset the gesis entry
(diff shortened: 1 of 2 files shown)


sgibson91:
There is a replica failing to terminate on production that is causing CI to fail: https://travis-ci.org/github/jupyterhub/mybinder.org-deploy/builds/682843917

betatim:
This turned out to be a mistake in the config. See the follow up PR for more details/discussion.

sgibson91:
Thank you for the quick fix

choldgraf:
I felt a great disturbance in the force, like a million kubernetes pods cried out, and in an instant were silenced...

PR #1429, opened 2020-05-07
[MRG] Switch to cert-manager from kubelego on OVH

betatim:
Follow up to #1415 that installs cert-manager on OVH.

There were no cRDs on the cluster so we didn't have to delete any.

I installed cert-manager v0.12 with `kubectl apply --validate=false -f https://raw.githubusercontent.com/jetstack/cert-manager/release-0.12/deploy/manifests/00-crds.yaml` and then merged this PR.

### DIFF
--- a/config/ovh.yaml
+++ b/config/ovh.yaml
@@ -1,5 +1,9 @@
 projectName: ovh
 
+tags:
+  kubelego: false
+  certmanager: true
+
 binderhub:
   config:
     BinderHub:
@@ -201,3 +205,12 @@ gcsProxy:
   buckets:
     - name: mybinder-staging-events-archive
       host: archive-analytics-staging-mybinder.mybinder.ovh
+
+
+cert-manager:
+  webhook:
+    enabled: false
+  ingressShim:
+    defaultIssuerName: "letsencrypt-prod"
+    defaultIssuerKind: "ClusterIssuer"
+    defaultACMEChallengeType: "http01"


betatim:
The challenge solvers that cert-manager installs were not able to have traffic routed to them. For example a URL like  http://ovh.mybinder.org/.well-known/acme-challenge/12341234-AZoRNPpkJu0t4rosEaiuHMUcUo12341234 would get routed to BinderHub, which responded with a status 404.

I listed the Ingress objects and found that `kube-lego-nginx             binder.mybinder.ovh,ovh.mybinder.org                                                             80        350d` still existed. Deleted it. This didn't immediately resolve the problem.

betatim:
One thing that isn't right is that requests to "http" (for the HTTP01 challenges) are redirected to "httpS". This then fails because we don't have a valid TLS certificate. I am not sure why/what is causing the redirect.

betatim:
Reading https://letsencrypt.org/docs/challenge-types/#http-01-challenge the fact that we have invalid certs involved when redirecting to httpS shouldn't be a problem.

betatim:
Have to move on to something else now. In summary: CM is deployed and trying to get certs but for some reason traffic isn't being routed to the challenge solvers. Instead requests to `/.well-known/acme-challenge/...` are sent to the binderhub pod which responds with a 404. I don't (yet) understand why this happens, especially because it seems to also happen if I manually delete all the Ingress objects that point to the binderhub pod.

manics:
Is the public facing Nginx independently managed, e.g. on a gateway server which proxies to the ingress on an internal K8s cluster?

betatim:
I think the answer is "no". But I am not super sure I understand exactly what you mean.

We have a `nginx-ingress` controller running on the OVH cluster like on GKE. The Service object for nginx is of type LoadBalancer which (as far as I know) triggers the deployment of a cloud load balancer outside of k8s as it does on GKE.

PR #1436, opened 2020-05-15
[MRG] Stop deploying to OVH

betatim:
The OVH cluster doesn't work right now (certs and etcd storage). The failing
deploy to it means we never run the tests on GKE or Turing. Instead we are
getting used to the deploy being red. As a result this removes the deploy and test
of OVH until we can get it fixed up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,18 +91,18 @@ before_deploy:
 - |
   # Stage 5, Step 2: Deploy to production
   python ./deploy.py prod prod-a
-- |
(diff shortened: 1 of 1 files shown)


choldgraf:
ah I feel bad - I noticed this yesterday and started writing an issue to discuss and then got distracted and forgot about it :-(  thanks for catching it

betatim:
This PR was up for like 60s Chris! :D

choldgraf:
this is me all the time 

image

PR #1443, opened 2020-05-21
[MRG] Add a Jitsi link to our page template

betatim:
This gives every Binder a unique Jitsi room that people can join to talk
about that particular repository.

When people click the link to the jitsi room they arrive on a "entrance" page, so people aren't immediately "live" in the conference call. They also join audio muted. I think this could be super useful ... or super weird. I think we should run it for a bit and see what happens.

Original idea and inspiration via @yuvipanda https://twitter.com/yuvipanda/status/1263355644184158208

### DIFF
--- a/mybinder/files/etc/jupyter/jupyter_notebook_config.py
+++ b/mybinder/files/etc/jupyter/jupyter_notebook_config.py
@@ -20,11 +20,12 @@ binder_persistent_request = os.environ.get('BINDER_PERSISTENT_REQUEST', '')
 
 repo_url = os.environ.get('BINDER_REPO_URL', '')
 
-if repo_url:
+# Roll this out only for binder-example repos to gain experience
+if repo_url and "binder-example" in repo_url:
     # Need a simple way to generate a stable but unique value for every repo
     # that we can then use to create a Jitsi meet URL
     unique = hashlib.md5(repo_url.encode())[:10]
-    jitsi_url = 'https://meet.jit.si/mybinder.org-%s#config.startWithAudioMuted=true&config.prejoinPageEnabled=true' % unique
+    jitsi_url = 'https://meet.jit.si/mybinder.org-%s#config.startWithVideoMuted=true&config.startWithAudioMuted=true&config.prejoinPageEnabled=true' % unique
 else:
     jitsi_url = ''
 
--- a/mybinder/files/etc/jupyter/jupyter_notebook_config.py
+++ b/mybinder/files/etc/jupyter/jupyter_notebook_config.py
@@ -24,7 +24,7 @@ repo_url = os.environ.get('BINDER_REPO_URL', '')
 if repo_url and "binder-example" in repo_url:
     # Need a simple way to generate a stable but unique value for every repo
     # that we can then use to create a Jitsi meet URL
-    unique = hashlib.md5(repo_url.encode())[:10]
+    unique = hashlib.md5(repo_url.encode())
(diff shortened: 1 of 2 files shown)


yuvipanda:
We should set video to be off by default too, and with that YESSS LET US DO THIS. We should maybe also disable recording? We should do what we can to prevent https://en.wikipedia.org/wiki/Zoombombing

betatim:
https://github.com/jitsi/jitsi-meet/blob/master/config.js is the list of config stuff we can set (I think). Not super sure if we can disable recording and if we can what stops people from using a screenrecorder (like Kap) to record? I tried to use the "maintenance message" property to show people a message but couldn't make it work. Something to investigate more I reckon as we could use it to display a short message informing people about what is about to happen.

Do you think 10 hex digits make it hard enough to guess a valid room name? I picked ten because it is 2 more than 8 which we use to keep our docker image tags unique.

betatim:
Now starts with video muted and should only appear for repos that have `binder-example` in their repo URL (so basically https://github.com/binder-examples). Probably good to test with a small set of repos before we break the HTML for everyone :D

mybinder/files/etc/jupyter/jupyter_notebook_config.py
```diff
@@ -17,9 +18,20 @@
 binder_request = os.environ.get('BINDER_REQUEST', '')
 binder_persistent_request = os.environ.get('BINDER_PERSISTENT_REQUEST', '')
 
+repo_url = os.environ.get('BINDER_REPO_URL', '')
+
+if repo_url:
+    # Need a simple way to generate a stable but unique value for every repo
+    # that we can then use to create a Jitsi meet URL
+    unique = hashlib.md5(repo_url.encode())[:10]
```
yuvipanda, line 27: Let's just use the full md5? Is there a character limit to doing that?

mybinder/files/etc/jupyter/templates/page.html
```diff
@@ -5,6 +5,11 @@
 {% block login_widget %}
 {% endblock %}
 
+{% if jitsi_url %}
+    <a id="visit-repo-link" href="{{ jitsi_url }}" class="btn btn-default btn-sm navbar-btn" target="_blank"
+       style="margin-right: 2px; margin-left: 4px;">Start Video Chat</a></span>
```
yuvipanda, line 10:
```suggestion
       style="margin-right: 2px; margin-left: 4px;">Join this repo's Video Chat</a></span>
```

yuvipanda:
Right re: Kap. Just a way to increase the bar and reduce casual trolling. 

woop woop otherwise though. And +1 for staggered rollout!

mybinder/files/etc/jupyter/jupyter_notebook_config.py
betatim: I don't think so. At least the auto generated ones are pretty long and the docs suggest to use long but easy to remember room names.

betatim:
Thanks for the text tweak!

Let's do it.

yuvipanda:
@betatim YESS LET US.

betatim:
https://meet.jit.si/mybinder.org-123124124gdfdfgdgf

choldgraf:
This is really interesting! I think I have two major concerns:

1. Will users be confused by this? E.g. are we gonna get a bunch of "I clicked the Jitsi button but nobody was in the room" questions?
2. Is it a potential vector for abuse? Just spitballing here - but are we making ourselves in any way legally liable for conversations that happen inside of the Jitsi rooms we provide? Or are we opening ourselves to the same "zoom-bombing" problems others have experienced?

PR #1453, opened 2020-05-27
[MRG] Make jitsi button available in all repos

betatim:
In #1443 we added a "jitsi button" to a few repos. This PR makes it available for all repos.

We limited the repos to see if there is any negative/positive feedback on this. So far we haven't heard anything, either way.


choldgraf:
Cool - is this something we want to advertise in any way? Or keep it as a shadow feature for the 1337 binder users?

betatim:
I'd keep it quiet for a while to see what happens. "a few repos" -> "all repos but now publicity" -> "all repos with publicity" is probably a good set of steps up. It should be fairly obvious because of the big button that appears in the UI.

choldgraf:
OK, let's give it a whirl

one quick thing I noticed: the jupyter notebook CSS makes the button slightly taller than the others:

image

maybe not needed to be fixed though, since Tim is working on the bottom-right hover button...

Issue #1445, opened 2020-05-21
Figure out OVH ingress issue


minrk:
### Bug description

OVH is currently failing to deploy due to running out of etcd storage, which in turn appears to be related to cert-manager creating lots of ingresses for allocating certificates.

We have removed it temporarily from the deploy process while we work this out: #1436 

I ran into what I believe is the same issue when nbviewer was deployed on this cluster in the first week of March, when I wrote:

> In attempting to create ingresses for the new services, it didn't work, the ingresses wouldn't show up (no matter what namespace I created them in). While debugging the ingress controller (it's loading a stale list of ingresses and I can't get it to notice new ones or forget old ones that don't exist anymore)

I or someone will need to work on a more specific reproduction process.

#### Expected behaviour

Creating ingresses should affect traffic routing.

#### Actual behaviour

Ingresses appear to have no effect.

### How to reproduce

Still need to figure this one out, since it's been a while since I was dealing with this. Creating ingresses and sending traffic that should be handled by them appears to be the first step.

/cc @mael-le-gal @jagwar

mael-le-gal:
Hello, I'm on it.

Here is my advancement : 

* On Wednesday I cleaned the ETCD because it was running out of memory.
* After that I choose to upgrade the k8s version because 1.14 was not supported anymore by our k8s team.
* Problem -> I went a little bit to far and I realize that `mybinder.org-deploy` is only compatible with k8s <= `1.15`
* So I choose to respawn a cluster in `1.15` version and reinstall `binder` on it

What is still in `TODO` :

* One problem cames never alone and we also have a certificate problem on the `*.mybinder.ovh` => Once it would be fixed binder will be up again
* So that we can completely delete the previous cluster I also need to transfert the `nbviewer` namespace on the new cluster

choldgraf:
@mael-le-gal thanks for the update!

Do we know why mybinder.org is only supported by k8s <=1.15?

consideRatio:
Perhaps some of these deprecations of k8s 1.16? https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

I found two files with issues:
- https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/templates/redirector/deployment.yaml
- https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/templates/proxy-patches/deployment.yaml

[...]

PR #1455, opened 2020-05-29
[OVH] Fixing and re-enabling deployment

mael-le-gal:
OVH cluster is back on track : https://mybinder-sre.readthedocs.io/en/latest/operation_guide/federation.html

For a summary of what have been done : 
https://github.com/jupyterhub/mybinder.org-deploy/issues/1445

Summary of the summary : 
* New OVH K8S cluster for binder in version `v1.15.11` (new kubeconfig encrypted and included in PR)
* Inside OVH cluster we should'nt use the `cert-manager` except for the `ovh.mybinder.org` ingress => all other ingress with `*.mybinder.ovh` will fail because we have rates problem on `*.ovh` domains with LetsEncrypt.


betatim:
Let's see if it works

betatim:
Looks like it is all working again now :)

Thanks for taking the time to fix things up!

PR #1484, opened 2020-07-06
Remove Turing from the Federation

sgibson91:
Setting Turing's pod quota to zero til I figure out why the h*ll the chart won't upgrade

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -10,7 +10,7 @@ letsencrypt:
 binderhub:
   config:
     BinderHub:
-      pod_quota: 80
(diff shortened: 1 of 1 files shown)


manics:
Thanks, maybe we prioritise https://github.com/jupyterhub/mybinder.org-deploy/pull/1469 after scipy?

sgibson91:
> Thanks, maybe we prioritise #1469 after scipy?

Yes, well if I pin the actions of the third-party commits, then it's ready to merge. I can do the Docker env work in a second PR.

Issue #1485, opened 2020-07-07
Turing cluster un-upgradeable


sgibson91:
I think this may be due to the upgrade to k8s v1.16.9

Error message:

```
Error: UPGRADE FAILED: Upgrade --force successfully deleted the previous release, but encountered 5 error(s) and cannot continue: unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"; unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"; unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"; unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"; unable to recognize "": no matches for kind "DaemonSet" in version "extensions/v1beta1"
```

We can see that `extensions/v1beta1` has been deprecated [here](https://github.com/kubernetes/kubernetes/blob/9d1179360cfc9fb6e905252cccd482b987d45393/CHANGELOG/CHANGELOG-1.6.md#deprecations).

Does anyone have any suggestions to try before I do `helm delete turing --purge`?

Related: #1482 #1484

manics:
Sounds like the BinderHub was already fixed to work with BinderHub in November 2019:
- https://github.com/jupyterhub/binderhub/issues/1011
- https://github.com/jupyterhub/binderhub/pull/1012

I guess there's no harm in trying a delete.

consideRatio:
Not sure. I would do a manual inspection to delete k8s resources that may be registered with the k8s api-server and try cleanup. They may not be found any more without specifying them with the full name. Instead of kubectl get deployments, you may need kubectl get deployments.extensions.v1beta or something like that (i dont remember this syntax, but resource names are abbreviations for their kind followed by apigroup etc, which has changed whats supported etc)

betatim:
My hunch is that the problem is that resources of type `extensions/v1beta1` were created when the cluster was still on a version where this type was still allowed. Then the version of k8s got upgraded to one where `extensions/v1beta1` is no longer supported. Now we are trying to deploy a helm chart that doesn't contain any resources of type `extensions/v1beta1` any more but the current objects are of that type. So now during the deploy process helm is trying to find objects of that type, and failing.

One thing that I find helpful is to look at the log of the tiller pod while the deploy is going. It often contains more useful information than what `helm` prints on the console.

sgibson91:
[...]

PR #1493, opened 2020-07-10
bump prometheus chart to 11

minrk:
this appears to eliminate the use of extensions/v1beta1, which is causing deployments to turing to fail (#1485)

unclear why other deployments are not seeing a similar failure, since GKE/OVH are on 1.15 and GESIS is on 1.18 while turing is on 1.16. Maybe it's a change in 1.16 and GESIS doesn't deploy this prometheus chart version?

### DIFF
--- a/mybinder/requirements.yaml
+++ b/mybinder/requirements.yaml
@@ -3,7 +3,7 @@ dependencies:
    version: 0.20.0
    repository: https://kubernetes-charts.storage.googleapis.com
  - name: prometheus
-   version: 9.1.2
+   version: 11.0.2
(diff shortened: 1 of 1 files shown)


bitnik:
> Maybe it's a change in 1.16 and GESIS doesn't deploy this prometheus chart version?

we had prometheus 9.7.4 deployed and I just upgraded it to 11.0.2 as well. I dont know the details of (https://github.com/jupyterhub/mybinder.org-deploy/issues/1485), but what I remember is when we upgraded to 1.18, we didnt upgrade the cluster but setup a fresh cluster and deployed everything new, maybe that is why we didnt have the issue that turing is having.

Issue #1500, opened 2020-07-12
📈 2020-07-13 Request to bump resources for Ganga Tutorial at PyHEP 2020


matthewfeickert:
> 📅 Ideally open an issue two weeks before an event, though we will still consider requests up to 48 hours before the event 📅

This request is being put in less than 48 hours in advance. We fully understand this request might not be possible to fulfill and can drop Binder support for this section. We apologize for this late request.

### 1. Who you are

@matthewfeickert (PyHEP 2020 workshop organizer contact) on behalf of @egede (presenter).

### 2. A few details about the event/course/context

This is for a tutorial at the [PyHEP 2020 workshop](https://indico.cern.ch/event/882824/) 2020-07-13 through 2020-07-17.

### 3. If money is changing hands (attendees pay a fee, speaker is getting paid, free, etc)

No, all speakers and organizers are volunteers. The workshop is free of charge and all resources will be made publicly available both on the [workshop website/Indico page](https://indico.cern.ch/event/882824/) and on the [High Energy Physics Software Foundation YouTube channel](https://www.youtube.com/channel/UCv4hukXGkCYvBClQMKzypnQ).

### 4. A link to the repository you are using

https://github.com/ganga-devs/ganga/

### 5. What type of event it is:
- [ ] talk
- [x] workshop
- [ ] lecture course
- [ ] training session

(This is a tutorial at a workshop, so I chose "workshop". I will be submitting more of these, so please let me know if I should change this.)

### 6. How many people you expect to attend the event

There are 1000 attendees registered, but we would anticipate that an upper _estimate_ would be maybe 600-700 would connect at this time.

### 7. The exact times and dates of the event(s) (with timezone information)

[Ganga: flexible use of of virtualisation for user based large scale computations (TUTORIAL)](https://indico.cern.ch/event/882824/timetable/#37-ganga-flexible-use-of-of-vi) is occurring at 17:50 - 18:45 UTC-6 on 2020-07-13.

### 8. How do you plan to give mybinder.org a shout out at the event

We will make an announcement at the start of the workshop about Binder, the service, and how the resources are being provided. If there is specific language the Binder team would like us to use we are happy to integrate this in.

matthewfeickert:
@betatim Seems that the reported number of attendees for the "Pacific" timezone sessions is very much ff from reality as we peaked at 100 people in the call and at most had 38 Binder sessions running all at once.

[...]

Issue #1499, opened 2020-07-12
📈 2020-07-13 Request to bump resources for Uproot & Awkward Arrays Tutorial at PyHEP 2020


matthewfeickert:
> 📅 Ideally open an issue two weeks before an event, though we will still consider requests up to 48 hours before the event 📅

This request is being put in less than 48 hours in advance. We fully understand this request might not be possible to fulfill and can drop Binder support for this section. We apologize for this late request.

### 1. Who you are

@matthewfeickert (PyHEP 2020 workshop organizer contact) on behalf of @jpivarski (presenter).

### 2. A few details about the event/course/context

This is for a tutorial at the [PyHEP 2020 workshop](https://indico.cern.ch/event/882824/) 2020-07-13 through 2020-07-17.

### 3. If money is changing hands (attendees pay a fee, speaker is getting paid, free, etc)

No, all speakers and organizers are volunteers. The workshop is free of charge and all resources will be made publicly available both on the [workshop website/Indico page](https://indico.cern.ch/event/882824/) and on the [High Energy Physics Software Foundation YouTube channel](https://www.youtube.com/channel/UCv4hukXGkCYvBClQMKzypnQ).

### 4. A link to the repository you are using

https://github.com/jpivarski/2020-07-13-pyhep2020-tutorial

### 5. What type of event it is:
- [ ] talk
- [x] workshop
- [ ] lecture course
- [ ] training session

(This is a tutorial at a workshop, so I chose "workshop". I will be submitting more of these, so please let me know if I should change this.)

### 6. How many people you expect to attend the event

There are 1000 attendees registered, but we would anticipate that an upper _estimate_ would be maybe 600-700 would connect at this time.

### 7. The exact times and dates of the event(s) (with timezone information)

[Uproot & Awkward Arrays (TUTORIAL)](https://indico.cern.ch/event/882824/timetable/#29-uproot-awkward-arrays-tutor) is occurring at 8:00 - 9:00 UTC-6 on 2020-07-13.

### 8. How do you plan to give mybinder.org a shout out at the event

We will make an announcement at the start of the workshop about Binder, the service, and how the resources are being provided. If there is specific language the Binder team would like us to use we are happy to integrate this in.

jpivarski:
> ### 4. A link to the repository you are using
> https://github.com/jpivarski/2020-07-13-pyhep2020-tutorial

An exact Binder link is now available: we want 600‒700 people to be able to run

[...]

PR #1508, opened 2020-07-12
Increase limits for PyHEP conference day 1

betatim:
xref #1499 and #1500 for details

Dealing with the repos for the first day of the conference only. The limit is increased to 300 ("ganga tutorial") and 400 ("awkward array tutorial"). These are both below the requested number off ~600. Let's discuss tomorrow (Monday) morning if we want to increase the limit even further. For a live stream with about 600-700 viewers in the stream 200 was a good limit that corresponded to basically the number of concurrent instances for that repo.

It would be good to gain some experience with what the scale factor is between registered number of people and concurrent binder pods.

Another thing to ponder is that currently all of mybinder.org peaks at around 700-800 concurrent pods. Nearly doubling this will certainly lead to "interesting" things happening.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -83,6 +83,14 @@ binderhub:
         - pattern: ^dask/dask-tutorial.*
           config:
             quota: 300
+        # https://github.com/jupyterhub/mybinder.org-deploy/issues/1500
+        - pattern: ^ganga-devs/ganga.*
+          config:
+            quota: 300
+        # https://github.com/jupyterhub/mybinder.org-deploy/issues/1499
+        - pattern: ^jpivarski/2020-07-13-pyhep2020-tutorial.*
+          config:
+            quota: 400
 
     GitRepoProvider:
       banned_specs:


matthewfeickert:
Awesome. Thanks @betatim. I'll tag @rochaporto and @lukasheinrich here so it is easier to discuss what level CERN's Binder Hub can spin up for helping out as well. I'm off time zone, but I'll do my best to try and be available to make any changes or answer any questions.

eduardo-rodrigues:
Greetings and thanks very much  @betatim !

jpivarski:
I'd like to look up the actual usage statistics for all the repos that were offered during the PyHEP workshop. Is this the best source of information about how to do that, or is there API documentation somewhere?

https://github.com/choldgraf/binder-stats

Thanks!

betatim:
https://archive.analytics.mybinder.org/ contains an entry for every launch so this is the best place to go to compute stats.

jpivarski:
Fantastic, thank you!

image

jpivarski:
This is the impact that PyHEP had on Binder:

Screenshot 2020-07-28 at 12 24 45 PM

In retrospect, we needed an increase above 200, but not above 300, on the first day. On subsequent days, the regular limit of 200 would have sufficed. This is for 1000 registrants, so the factor you're looking for is 0.1‒0.2.

PR #1322, opened 2020-01-22
[Example] Add Some basic CI tests for PRs

sgibson91:
This PR uses GitHub Actions to add some basic CI tests to provide feedback in PRs (not just on merge to master). These tests **do not** block merging of PRs on failure.

Added actions:

- Check linting of Python scripts with Flake8
- Check linting of YAML with yamllint
- Lint and Validate script for helm charts from [zero-to-jupyterhub-k8s repo](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/master/tools/templates)
- Check linting and validate the charts of all deployments which are configured from this repo

### Future work

I'd like to work towards CI that can help us verify a deployment before we hit merge. This would most likely involve spinning up a new cluster. However, I think this would be possible given the work going on in integrating k8s and Actions happening at the minute (e.g. https://github.com/Azure/actions#deploy-to-kubernetes this is probably Azure specific). Failing that, we could create our own action that would run a docker container that does the deployment, runs tests and tears down following [this example](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-a-docker-container-action) and piggy-backing on the code in https://github.com/alan-turing-institute/binderhub-deploy (or https://github.com/alan-turing-institute/binderhub-deploy-gke if I ever get back around to it).

Related discussion: https://github.com/jupyterhub/mybinder.org-deploy/pull/1317#issuecomment-576564422 https://github.com/jupyterhub/mybinder.org-deploy/pull/1317#issuecomment-576586207 https://github.com/jupyterhub/mybinder.org-deploy/pull/1317#issuecomment-576609134 https://github.com/jupyterhub/mybinder.org-deploy/pull/1317#issuecomment-577165080

PR #1535, opened 2020-07-15
[MRG] Add GitHub Action workflows to lint helm charts

### LINKED ISSUE #1324: CI/GitHub Actions tests specifically for PRs
I've opened PR #1322 with some basic linting tests for the repo and a script that I borrowed from the zero-to-jupyterhub-k8s repo to validate the helm charts (that action currently doesn't work because I haven't uploaded the git-crypt secret). I think in particular, the validation of the helm charts is a good one as it will (hopefully) flag up issues before merge. The PR doesn't have to be merged, but it's there if we decide this sort of thing will be useful.

I think it's be cool to work to something that tests the deployment in the PR before going live. This will probably involve spinning up/tearing down a temporary cluster, but it's probably possible to hack something together from the scripts in https://github.com/alan-turing-institute/binderhub-deploy / https://github.com/alan-turing-institute/binderhub-deploy-gke and using a [docker action](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-a-docker-container-action) and some of the [k8s actions](https://github.com/Azure/actions#deploy-to-kubernetes).

sgibson91:
I'm using a relaxed config to lint the helm charts with (i.e. ignoring warnings, not checking for new-line-at-file-end, trailing-whitespace) since the point of this is to check for things that will cause a deployment to fail. If we do want to enforce proper linting, I'd recommend implementing that on the [jupyterhub/binderhub](https://github.com/jupyterhub/binderhub) repo.

closes #1324


scripts/yamllint-config.yaml
```diff
@@ -0,0 +1,19 @@
+# Use a more tolerant config
+# Indentation, line length, extra spaces, blank lines are raised as warnings - not errors
+#
+# https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration
+extends: relaxed
+
+# Because we can't be responsible for linting
+# the JupyterHub helm chart too
+ignore: |
+  **/jupyterhub/**
```
minrk, line 10: I think the `templates` directory also needs to be ignored, to exclude the not-yet-yaml templates in the chart itself.

sgibson91: So the workflow is:

* Run `helm template` and save to output directory
* Run `yamllint` on output dir

So I think the templates directory shouldn't be a problem? Unless we want to ignore it for folks who are running it manually?

sgibson91:
> This is great! Quick question: what happens if we have a lint error on an area containing a secret? Would it be revealed?

Good question, I don't know the answer to that. We could always template in some fake secrets to avoid that, though I'd have 2 follow-up questions:

1) Would that still catch problems with the secrets?
2) Would we still need to run the test for each cluster?

scripts/yamllint-config.yaml
minrk: I was basing that on [this error message](https://github.com/jupyterhub/mybinder.org-deploy/pull/1535/checks?check_run_id=877973877#step:10:13):

> [ERROR] templates/analytics-publisher/deployment.yaml: unable to parse YAML
>	error converting YAML to JSON: yaml: line 37: mapping values are not allowed in this context

but now I think I was misinterpreting that. I believe if you run:

    chartpress --reset

in the repo and commit the result, these tests will pass

sgibson91: Damn, I _always_ forget chartpress 😅

minrk:
> perhaps we can engineer that scenario to test?

Yes! Let's make each of the linters fail, just to see what's included in the error message. Because if they only show the location in the file, we don't need to worry about this. I ran some quick checks, and I think they don't print anything about the content that causes the errors.

> Would that still catch problems with the secrets?

maybe, but only if we replicate the structure of secret files exactly. I wrote a quick script in #1543 to allow copying a yaml file, replacing all scalar values that should contain secrets with placeholders, so the structure remains the same while hiding any actual secret values. But we can skip all that if the linter errors don't print context.

> Would we still need to run the test for each cluster?

I think so, because it's easy to get things wrong in fields we set everywhere, and templates can render incorrectly if fields are unspecified (as [this error](https://github.com/jupyterhub/mybinder.org-deploy/pull/1535/checks?check_run_id=877973877#step:10:13) reveals, where the chart is invalid if some image fields are not overridden.

scripts/yamllint-config.yaml
minrk: alternatively, running chartpress prior to lint should also fix it, and maybe we should do that anyway? It would also catch another class of error preventing image builds, etc.

minrk:
btw, should we close #1322? This looks like a second stab at the same effort.

sgibson91:
> btw, should we close #1322? This looks like a second stab at the same effort.

Yes, I just had it open while I was copying stuff across (I started afresh since it's so old). But I probably have all I need from it now.

scripts/yamllint-config.yaml
sgibson91: Running chartpress (just `chartpress`) fixed some of the issues with yamllint (think Turing has a genuine error), but it does break helm lint https://github.com/jupyterhub/mybinder.org-deploy/runs/881266943#step:11:12

minrk:
Looks like it caught a [real lint error](https://github.com/jupyterhub/mybinder.org-deploy/runs/881277384?check_suite_focus=true#step:13:8)!

> helm-turing/mybinder/charts/binderhub/templates/ingress.yaml
> 14:5      error    duplication of key "kubernetes.io/tls-acme" in mapping  (key-duplicates)

(tls-acme annotation in binderhub.ingress in config/turing.yaml is redundant)

sgibson91:
Whoop! Got it passing, now to break it see if secrets are revealed!

minrk:
I think the latest commit didn't trigger an error because we only run the linter on the final render, not directly on our input files. It might be worth running yamllint on `config` and `secrets/config` before running `helm template` to catch informative errors.

minrk:
**update:** I ran yamllint on secrets/config myself, and lacking quotes doesn't raise an error. Quotes aren't generally necessary on strings in yaml, so maybe something else should be tested?

sgibson91:
> **update:** I ran yamllint on secrets/config myself, and lacking quotes doesn't raise an error. Quotes aren't generally necessary on strings in yaml, so maybe something else should be tested?

Added a new job to test this. I made it a new job since it didn't need to be a matrix.

sgibson91:
> **update:** I ran yamllint on secrets/config myself, and lacking quotes doesn't raise an error. Quotes aren't generally necessary on strings in yaml, so maybe something else should be tested?

sgibson91:
Ha, this entire PR could be replaced with this 😂 https://github.com/marketplace/actions/yamllint-github-action

betatim:
Nice work!

All my knowledge about validation of helm charts comes from reading this PR and then looking at https://learnk8s.io/validating-kubernetes-yaml :D

PR #1543, opened 2020-07-17
add obfuscate-yaml.py script

minrk:
to allow testing linters, etc. with secret files, but without secret *values*

For possible inclusion in #1535 

Not sure if we need this, but if it's deemed useful to test with secret values, but we can use it if it turns out linter errors in secret files would leak secret values, we can use this to test with *structure* but without using the actual values themselves.


minrk:
this appears to be unnecessary after testing in #1535

PR #1571, opened 2020-08-19
[WIP] Edit lint-validate workflow so secrets aren't linted in PRs from forks

sgibson91:
We had problems that the linting GitHub Action fails on PRs from forks since they can't access the secrets. This PR edits the workflow file to only run jobs that require secrets in PRs that come from within the origin repo.


sgibson91:
I seem to have broken this PR beyond running the action so I'm closing and trying again 😅

PR #1572, opened 2020-08-19
[MRG] Edit linting workflow so secrets aren't required when PR comes from fork

sgibson91:
We had failure issues on PRs that come from forks because they can't access the repo's secrets. I'm trying to reconfigure the lint-validate.yml workflow file so that the steps that require secrets are only run when the source branch of a PR is owned by the host repository.

I've opened this PR from my fork to test this.

### DIFF
--- a/.github/workflows/lint-validate.yml
+++ b/.github/workflows/lint-validate.yml
@@ -1,10 +1,6 @@
 name: Lint and Validate helm chart
 
-on:
-  push:
-    paths:
-      - "**/*.yaml"
-      - "**/*.yml"
+on: [push]
 
 env:
   HELM_VERSION: "v2.16.3"
--- a/.github/workflows/lint-validate.yml
+++ b/.github/workflows/lint-validate.yml
@@ -10,16 +10,6 @@ env:
   HELM_VERSION: "v2.16.3"
 
 jobs:
-  set-env-var:
-    runs-on: ubuntu-16.04
-    steps:
-      - name: Set fork status as an environment variable
-        run: |
-          STATUS="$( echo ${PAYLOAD} | jq .pull_request.head.repo.fork )"
-          echo "::set-env name=fork::${STATUS}"
-        env:
-          PAYLOAD: ${{ toJSON(github.event) }}
-
   yamllint-raw:
     needs: set-env-var
     runs-on: ubuntu-16.04
@@ -46,19 +36,19 @@ jobs:
           yamllint -d scripts/yamllint-config.yaml --no-warnings config/
 
       - name: Unlock git-crypt secrets
-        if: env.fork == 'false'
+        if: github.event.pull_request.head.repo.fork == 'false'
         uses: sliteteam/github-action-git-crypt-unlock@a09ea5079c1b0e1887d4c8d7a4b20f00b5c2d06b
         env:
           GIT_CRYPT_KEY: ${{ secrets.GIT_CRYPT_KEY }}
 
       # We use --no-warnings in this step to reduce output to critical errors
       - name: Run yamllint on secrets/config
-        if: env.fork == 'false'
+        if: github.event.pull_request.head.repo.fork == 'false'
         run: |
           yamllint -d scripts/yamllint-config.yaml --no-warnings secrets/config/
 
   yamllint-templates:
-    needs: set-env-var
+    if: github.event.pull_request.head.repo.fork == 'false'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


sgibson91:
@hamelsmu I wonder if you can help me out on this? I'm just looking for something that tells me if the source branch in a PR is from a fork or not and it's proving to be trickier than I anticipated. Here are the things I've tried so far:

- **Conditionals.** `if: github.repository == 'jupyterhub/mybinder.org-deploy`

   This ended up not working I _think_ because the original trigger was `pull_request` so the context was running as a merge commit and so the above conditional always returned true. As you can see from the current state of this PR, I then tried the `push` trigger, but that doesn't seem to be providing feedback in the PR (though I can see that it is successfully running in my fork https://github.com/sgibson91/mybinder.org-deploy/runs/1002561676?check_suite_focus=true )

- **`pull_request_target` trigger.**

   From my understanding of [these docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target), my understanding that this would run similarly to `pull_request` except it would "see" the PR as coming from within the same repo and therefore could access secrets. But maybe I misunderstood that and what is actually happening is that it's only triggered when the PR _does_ come from the same repo?

What I really want is to do `if: github.event == 'pr_from_fork'` or something similar. Am I missing something?

hamelsmu:
@sgibson91 if you would like a conditional, I think [these variables](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions) will be useful.

For example 
`if:github.repository_owner != 'jupyterhub'`
 or  
you can inspect the `github.event` payload to get the field you want.  The way I suggest seeing this payload is to introspect it in an actual workflow [like this](https://github.com/fastai/fastai/blob/master/.github/workflows/main.yml#L9-L24):

example:
```yaml
name: CI
on: 
  workflow_dispatch: #allows repo admins to trigger this workflow from the Actions tab
  pull_request:
  push:
    branches: 
      - master

jobs:     
  debug:
    runs-on: ubuntu-latest
    steps:

    - name: see payload
      run: |
        echo "$PAYLOAD"
      env:
        PAYLOAD: ${{ toJSON(github.event) }}
    
    - name: store payload
      uses: actions/upload-artifact@v2
      with:
        name: payload.json
        path: ${{ github.event_path }}
 ```

You will likely find the information you want in the payload and can use [javascript dot notation](https://www.w3schools.com/js/js_json_objects.asp) to access the relevant field in the JSON payload.  

I hope that helps!  LMK how it goes!

**Comment by sgibson91** [comment]
> `if:github.repository_owner != 'jupyterhub'`

I just want to reiterate that `github.repository_owner` _always_ returns `jupyterhub` when running in a `pull_request` event (I think because it's happening in the merge commit). This doesn't feel like very useful behaviour to me when trying to distinguish between an "internal" PR and a forked PR. Obviously running in a `push` event will return `sgibson91` as expected, but that doesn't report here which also feels suboptimal.

You're right that I can get the info I want from the payload, thank you for that pointer. I'm still writing more code than I'd like to to get this to work though 😛

**Comment by sgibson91** [comment]
Unfortunately the following config is invalid:

```yaml
yamllint-templates:
  needs: set-env-var
  if: env.fork == 'false'  # where this is set by set-env-var
```

Therefore, I had to add `if: env.fork == 'false'` to every step in `yamllint-templates` which is really messy. However, we now have a workflow that only executes the parts that need secrets when the PR _doesn't_ come from a fork.

hamelsmu:
@sgibson91 I took a look at your yaml file -- you can set an if statement at the job level also, do you think that would work for you?  Also, you could have one job depend on another job with the `needs:`  keyword.

.github/workflows/lint-validate.yml
```diff
@@ -29,18 +40,25 @@ jobs:
...
       # We use --no-warnings in this step to reduce output to critical errors
-      - name: Run yamllint on config/ and secrets/config
+      - name: Run yamllint on secrets/config
+        if: env.fork == 'false'
         run: |
-          yamllint -d scripts/yamllint-config.yaml --no-warnings config/
           yamllint -d scripts/yamllint-config.yaml --no-warnings secrets/config/
 
   yamllint-templates:
+    needs: set-env-var
```
hamelsmu, line 61: Something like this here
```suggestion
    needs: set-env-var
    if: github.event.pull_request.head.repo.fork
```

.github/workflows/lint-validate.yml
```diff
@@ -10,7 +10,18 @@ env:
...
 jobs:
+  set-env-var:
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Set fork status as an environment variable
+        run: |
+          STATUS="$( echo ${PAYLOAD} | jq .pull_request.head.repo.fork )"
+          echo "::set-env name=fork::${STATUS}"
+        env:
+          PAYLOAD: ${{ toJSON(github.event) }}
```
hamelsmu, line 21: I am not confident that environment variables can be shared b/w different jobs?   Does this work, out of curiosity?  

I always thought you have to set global environment variables to have it visible by all jobs

.github/workflows/lint-validate.yml
sgibson91: Ah! With `if: github.event.pull_request.head.repo.fork == 'false'` I don't need the set-env-var job at all!

.github/workflows/lint-validate.yml
sgibson91: You can see from this workflow run that it worked as I expected https://github.com/jupyterhub/mybinder.org-deploy/runs/1004149313 Two steps in yamllint-raw that required secrets were skipped

sgibson91:
I think this has cracked it! Thank you for your help @hamelsmu!

.github/workflows/lint-validate.yml
hamelsmu: oh wow that's wild it works learned something new!

hamelsmu:
anytime @sgibson91 you are an expert now, 🙇 !!!

.github/workflows/lint-validate.yml
sgibson91: Here are the docs https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

betatim:
The bots are happy so gonna merge this. See what the bots have to say when other bots (henchbot) make PRs ;)

Issue #1560, opened 2020-08-07
📈 2020-08-30 Request to bump resources for PyCon JP 2020 Tutorial "Using Open Data to Analyze Data in Python"


kozo2:
### 1. Who you are
@kozo2

I am one of the tutors of [PyCon JP 2020 Tutorial](https://pycon.jp/2020/en/tutorial/) "Using Open Data to Analyze Data in Python"

### 2. A few details about the event/course/context
In the tutorial, we will share how to analyze Japanese open data using pandas, plotly, scikit-learn, and pydeck.

### 3. If money is changing hands (attendees pay a fee, speaker is getting paid, free, etc)
PyCon JP will pay us 50000 Yen.
There are 5 tutors including me (We will divide the 50000yen equally).
10000 Japanese Yen equals 94.73 United States Dollar.

### 4. A link to the repository you are using
https://github.com/hannari-python/tutorial

### 5. What type of event it is:
- [ ] talk
- [x] workshop
- [x] lecture course
- [x] training session

### 6. How many people you expect to attend the event
About 180~230.

At the moment, there are 193 registered participants.
(173[YouTube Live] + 20[Zoom] .)

Please refer to the following pages to check the number of registered participants (There are only Japanese pages).
- https://pyconjp.connpass.com/event/182390/
- https://pyconjp.connpass.com/event/181065/

### 7. The exact times and dates of the event(s) (with timezone information)
Japan timezone 2020/08/30 10:00 ~ 17:00

### 8. How do you plan to give mybinder.org a shout out at the event
At the beginning of the tutorial, let the audience know about the feature of **mybinder.org** and the fact that **mybinder team** provide the special support to the audience (for this tutorial).

minrk:
@kozo2 yes, we'll bump the resources on Friday ahead of the event. See #1574

kozo2:
@minrk Thank you for the reply. I am grateful for your support.

kozo2:
@minrk @betatim
Thank you for your support.
The tutorial was successfully completed without incident.
If there's anything that I can do (in return for this support), please let me know.

betatim:
Thanks a lot for the feedback and glad that things worked well for you.

I think a good ask in return would be: keep promoting mybinder.org and encourage people to help with the tools we use to build the service.

PR #1574, opened 2020-08-25
2020-08-28 bump resources for pycon-jp

minrk:
Event starts 2020-08-30 01:00Z, so we should merge this bump on the 28th (Friday)

ref: #1560

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -63,6 +63,11 @@ binderhub:
         - pattern: ^ipython/ipython-in-depth.*
           config:
             quota: 128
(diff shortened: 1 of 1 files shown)


minrk:
Edit: updated date to the 28th, since I realized 29th is Saturday

betatim:
I merged it now because I was looking at it and would have had to make a note to remind myself. It seemed like there is no big downside to merging early.

minrk:
@betatim yeah, no problem. As long as it's close-ish to the event. Better too early than too late!

PR #1576, opened 2020-08-25
apply memory limit to builds

### LINKED ISSUE #1529: Collect information about builds
In #1524, one of the things that occurred was a pretty large number of builds. We should investigate our current resource allocations to builds and model what happens when there are a lot of build requests, to see if there are any changes that should be made there, either in configuration here, or capabilities in BinderHub.

minrk:
- request 1G per build for scheduling purposes
- apply 2G limit on builds
- reduce dind pod memory request to account for builds now requesting resources while they are running

The result (for GKE prod) will be:

- less memory reserved while fewer than 7 builds are running on a node
- no change when 7 builds are running, other than per-build memory limit of 2G
- more memory reserved while more than 7 builds are running, increasing likelihood of builds triggering scale-up

closes #1529

requires https://github.com/jupyterhub/binderhub/pull/1144 for the build_memory_request to have its desired effect. We can either wait for that, or merge, in which case `build_memory_limit` would be used for the request as well, meaning builds would request 2G instead of 1G until the separate config option lands

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -27,7 +27,7 @@ binderhub:
     resources:
       requests:
         cpu: "2"
-        memory: 8Gi
+        memory: 1Gi
       limits:
         cpu: "5"
         memory: 12Gi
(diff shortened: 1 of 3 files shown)


betatim:
Let's wait for henchbot to update the chart and such for the BinderHub changes that just got merged. Then we can deploy this. My thinking is: lets not get in the way of the bot :D

minrk:
Makes sense! Bot has landed the binderhub bump in #1577

PR #1582, opened 2020-09-01
bump nginx-ingress, cert-manager

### LINKED ISSUE #1498: Upgrade nginx-ingress chart
We need to upgrade the nginx-ingress chart (#1494) in order to get turing back into the federation (#1485) because the very-old version we have now is too old for kubernetes 1.16.

However, upgrading the ingress controller is very disruptive, and can result in downtime (as I saw after merging #1494), and an upgrade this far might need some manual fixes. So we should try to do it when:

1. there isn't going to be a lot of traffic, and
2. folks have time to dedicate to figuring out the manual steps necessary, since a simple chart version bump is known to not work. In particular, when I ran it:

    - staging launched, but didn't load the right SSL certificates
    - prod launched, but default-backend never started, due to:

            Warning  FailedMount  14m (x23 over 45m)  kubelet, gke-prod-a-user-20200604-8fa7ad84-5l0x  MountVolume.SetUp failed for volume "prod-nginx-ingress-backend-token-67xbd" : secret "prod-nginx-ingress-backend-token-67xbd" not found
      and never became available.

As with many upgrade failures, it could be a matter of manually deleting the right objects before re-running an upgrade. Identifying what those are can take some time.

minrk:
Planning to do this Thursday, since it is likely to require manual intervention, as these are pretty long upgrades, which we should probably try to avoid.

In the meantime, I'm looking at upgrade docs of both to see if I can be well prepared for what comes.

Notes:

- upgrading cert-manager from 0.12 to 0.15 instead of 0.16, due to note [here](https://cert-manager.io/docs/installation/upgrading/upgrading-0.15-0.16/) suggesting that 0.16 requires kubernetes 1.16.14, while our GKE clusters are still on 1.15
- [ ] [0.14 upgrade docs](https://cert-manager.io/docs/installation/upgrading/upgrading-0.13-0.14/) suggest manually deleting all cert-manager deployments prior to upgrade
- [x] same doc notes that the webhook is now required, so our `webhook.enabled = false` config should be removed
- [x] looks like `ingressShim.defaultACMEChallengeType` doesn't exist anymore

nginx:

- [x] upgrading from 0.14 to 2.13
- [x] updating repo from deprecated `stable/nginx-ingress` to dedicated 
- [x] update config key to `ingress-nginx`
- [x] lots of config changes for the ingress
- [x] might need to add annotations for ingress controller to be used?
- [x] metrics config appears to have changed a lot
- [x] defaultBackend must be enabled explicitly

closes #1498

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -205,9 +205,3 @@ gcsProxy:
   buckets:
     - name: mybinder-staging-events-archive
       host: archive-analytics-staging-mybinder.turing.10.0.0.1.xip.io
-
-cert-manager:
-  ingressShim:
-    defaultIssuerName: "letsencrypt-prod"
-    defaultIssuerKind: "ClusterIssuer"
-    defaultACMEChallengeType: "http01"
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -1,6 +1,5 @@
 projectName: turing
 
-tags:
 
 letsencrypt:
   contactEmail: drsarahlgibson@gmail.com
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -1,8 +1,6 @@
 projectName: turing
 
 tags:
-  kubelego: false
-  certmanager: true
 
 letsencrypt:
   contactEmail: drsarahlgibson@gmail.com
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -158,12 +158,10 @@ prometheus:
           secretName: turing-prometheus-tls-crt
 
 
-nginx-ingress:
+ingress-nginx:
   controller:
     service:
       loadBalancerIP: 51.105.120.231
-    config:
-      proxy-body-size: 64m
 
 
 static:
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -207,8 +207,6 @@ gcsProxy:
       host: archive-analytics-staging-mybinder.turing.10.0.0.1.xip.io
 
 cert-manager:
-  webhook:
-    enabled: false
   ingressShim:
     defaultIssuerName: "letsencrypt-prod"
     defaultIssuerKind: "ClusterIssuer"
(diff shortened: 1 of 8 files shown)


minrk:
I've looked through the various config and updated everything I could identify.

Following the cert-manager upgrade docs, I will delete the cert-mananger deployments by hand on each cluster prior to upgrading. This shouldn't be disruptive, since cert-manager should only need to be running while new clusters are being issued.

Due to the large ingress chart update, I think there is a pretty good chance that we will lose the lease on our loadBalancerIP and need to update mybinder.org DNS for GKE clusters.

I'm also going to start with a trial run of #1583 so I can do this for staging only without touching prod.

minrk:
In anticipation of the possible DNS issue, I just updated the TTL on the GKE DNS records to 2 minutes, so it should hopefully propagate quickly if it does need updating tomorrow.

betatim:
I think we will keep the IP because we registered it as a "static IP", at least `gcloud compute addresses list` lists the one.

config/turing.yaml
```diff
@@ -1,8 +1,6 @@
 projectName: turing
 
 tags:
```
betatim, line 3:
```suggestion
```
missed line, i think

minrk:
@sgibson91 Trigger worked, great job! But we have one hiccup: staging deploy failed due to lack of access to git-crypt secrets on the pull_request event.

sgibson91:
Ah, that's because the PR came from a fork. I don't really know how to resolve this.

minrk:
Maybe we need to merge into a `staging` branch to trigger builds with credentials? Can we trigger that push from an Action with a tag event and then the build from the push event?

minrk:
For today, I'm going to try deploying staging manually with `deploy.py --local` so we don't have to wait too long. But This can be a job for another day.

minrk:
Local deploy to staging worked without a hitch, no dns update needed. The only manual intervention needed was deletion of the old nginx-ingress default backend. Going to go ahead and try prod now.

PR #1583, opened 2020-09-02
Deploy open PRs to staging only

sgibson91:

https://github.com/jupyterhub/binderhub/compare/<OLD-HASH>...<NEW-HASH>

https://github.com/jupyter/repo2docker/compare/<OLD-HASH>...<NEW-HASH>

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -7,11 +7,14 @@
 
 name: Continuous Deployment
 
-# Only trigger the workflow when pushing to master
+# Only trigger the workflow when pushing to master or a label is applied to a
+# Pull Request
 on:
   push:
     branches:
       - master
+  pull_request:
+    types: [labeled]
 
 # Global environment variables
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -26,6 +29,12 @@ env:
 jobs:
   # Job to upgrade staging
   staging-deploy:
+    # Only run the job if the 'test-staging' label is present OR if the event
+    # is a push to master
+    if: |
+      (github.event.label.name == 'test-staging') ||
```
minrk, line 35: IIUC, this means that the act of applying the label triggers the deploy? That means to deploy a new commit, we un-tag and re-tag the PR?

If so, do you want to add a bit in `docs/source/deployment/how.md` for how this is supposed to work?

minrk (commented):
Awesome! I'll give this a test tomorrow morning on #1582

.github/workflows/cd.yml
sgibson91: I believe so, yes. I can add it to the docs later on tonight 👍

sgibson91: Done! I had a bit of a tidy up while I was at it. Mentioned that henchbot takes care of _most_ of the procedure on that page and changed some travis references to GitHub Actions.

minrk: Awesome, thanks! Taking it for a spin!

minrk:
Exciting! I'm going to give it a go on #1582.

PR #1586, opened 2020-09-02
Tidying up lint-validate.yml workflow file

sgibson91:
In #1583, I noticed that our lint and validate GitHub Action wasn't triggering as I expected it to and this PR rectifies that. The problem was that the `if` statements were checking for the **string** `false` not the **bool** `false`. Also, the two jobs in this workflow file were running on different runners, so I made that consistent too.

### DIFF
--- a/.github/workflows/lint-validate.yml
+++ b/.github/workflows/lint-validate.yml
@@ -35,7 +35,7 @@ jobs:
           yamllint -d scripts/yamllint-config.yaml --no-warnings config/
 
       - name: Unlock git-crypt secrets
-        if: github.event.pull_request.head.repo.fork == false
+        if: !github.event.pull_request.head.repo.fork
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-validate.yml
```diff
@@ -35,19 +35,19 @@ jobs:
           yamllint -d scripts/yamllint-config.yaml --no-warnings config/
 
       - name: Unlock git-crypt secrets
-        if: github.event.pull_request.head.repo.fork == 'false'
+        if: github.event.pull_request.head.repo.fork == false
```
minrk, line 38: What about explicit boolean not with `!`?

```suggestion
        if: !github.event.pull_request.head.repo.fork
```

sgibson91: I tried it, but it turns out not to be valid in the GHA schema

minrk: Okay, then! The docs suggest it should be, so maybe a bug in the schema.

PR #1593, opened 2020-09-03
deploy cert-manager in its own chart

minrk:
This should hopefully finish the transition of cert-manager to its own helm release

I've already deployed this manually to staging via `deploy.py --local` and it has successfully issued the cert I started the request for in #1590 so hopefully this will work in prod as well

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -448,8 +448,3 @@ federationRedirect:
       weight: 1
       health: https://turing.mybinder.org/health
       versions: https://turing.mybinder.org/versions
-
-cert-manager:
-  ingressShim:
-    defaultIssuerName: letsencrypt-prod
-    defaultIssuerKind: ClusterIssuer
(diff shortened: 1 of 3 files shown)


mybinder/values.yaml
```diff
@@ -448,8 +448,3 @@ federationRedirect:
       weight: 1
       health: https://turing.mybinder.org/health
       versions: https://turing.mybinder.org/versions
-
-cert-manager:
```
minrk, line 452: this config has moved to config/cert-manager.yaml

minrk:
The important info here is that the publicly available cert-manager CRDs hardcode that cert-manager is in the cert-manager namespace, and various other things about service names, due to the webhook, which we used to have disabled but is no longer optional. That means that cert-manager can't be used as a dependency in another chart as we were doing, it must be its own chart and must be in the `cert-manager` namespace.

It does have an `installCRDs` option in the chart that is supposed to address these issues, but I think that only works correctly with very recent helm and kubernetes that aren't available to us yet (it requires a relatively old version of kubernetes, but it is still not old enough to be available GKE as of today).

PR #1600, opened 2020-09-09
update ingress apiVersion

minrk:
extensions/v1beta1 is removed in 1.16. The newer networking.k8s.io/v1beta1 is available from 1.14.

All our clusters are on at least 1.15, so we should be set. Put this in `kubeVersion` in the Chart.yaml to be safe.

this should help with #1485


sgibson91 (approved):
LGTM, let's try it!

PR #1602, opened 2020-09-09
helm doesn't evaluate kubernetesVersion prereleases correctly

minrk:
Workaround found [here](https://github.com/helm/helm/issues/6190#issuecomment-519712714)


minrk:
Noting that https://github.com/helm/helm/issues/6190 is closed as "wontfix" because they think 1.15.3-prerelease should be excluded by `>=1.13` which is simply an incorrect interpretation of how versions work.

PR #1603, opened 2020-09-09
specify kube version in kubeval

### LINKED ISSUE #1601: Update lint-validate CI to check k8s API versions for federation
This could help us avoid future issues like #1485 

Potentially useful tool https://github.com/instrumenta/helm-kubeval

minrk:
- lints all our yaml, not just config
- use helm-kubeval plugin as mentioned in #1601, skipping what should be a redundant yamllint on helm templates
- run helm-kubeval with the bare minimum secret config (in config/test-secrets.yaml) when secrets are not available, giving us a better chance of catching helm template issues prior to merge

closes #1601

### DIFF
--- a/.github/workflows/lint-validate.yml
+++ b/.github/workflows/lint-validate.yml
@@ -53,6 +53,7 @@ jobs:
       fail-fast: false
       matrix:
         release: ["staging", "prod", "ovh", "turing"]
+        kube_version: ["1.15", "1.18"]
 
     steps:
       - name: Checkout repo
@@ -113,6 +114,6 @@ jobs:
           yamllint -d scripts/yamllint-config.yaml --no-warnings helm-${{ matrix.release }}
 
       # Use --ignore-missing-schemas to avoid failure of CRDs
-      - name: Validate k8s resources for ${{ matrix.release}}
+      - name: Validate k8s resources for ${{ matrix.release }} kube-${{ matrix.kube_version }}
         run: |
-          kubeval --strict --ignore-missing-schemas -d helm-${{ matrix.release }}
+          kubeval --kubernetes-version ${{ matrix.kube_version }} --strict --ignore-missing-schemas -d helm-${{ matrix.release }}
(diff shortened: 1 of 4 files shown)


sgibson91 (approved):
LGTM! Thanks for this 🤖

minrk:
Got it to run and pass, but marking as draft again because it appears to be not validating against any kube schemas. 🤔

.github/workflows/lint-validate.yml
```diff
@@ -29,30 +29,50 @@ jobs:
...
+        - release: staging
+          kube_version: "1.15.7"
+        - release: prod
+          kube_version: "1.15.7"
+        - release: prod
+          kube_version: "1.18.1"
+        - release: ovh
+          kube_version: "1.15.7"
+        - release: turing
+          kube_version: "1.16.4"
```
sgibson91, line 75:
```suggestion
          kube_version: "1.16.10"
```

This is Turing's current `kubectl` version

minrk: I added a comment explaining the reason - this can't actually be the full version because not all version have published schemas. It has to be listed here: https://github.com/instrumenta/kubernetes-json-schema

sgibson91: Whoops, missed that! 😄

minrk:
ok, this actually works and validates against schemas now. We could skip `--ignore-missing-schemas` which would give us better errors (specifying an unavailable version of kubernetes will pass as it is now!) if we moved the ClusterIssuer CRD out of our helm chart and into our cert-manager deployment

sgibson91:
Sounds like a sensible idea because at least then there won't be any (cert-manager) CRDs in our namespace to toy with deployment - but I do realise cert-manager is difficult to work with.

minrk:
It shouldn't be too bad, but we have to be careful about the order. It seems to work best when we do:

1. remove from helm, deploy
2. add manual, deploy again

because if we add it then remove it, things could get confused.

That can be another PR, though.

betatim:
Nice work!

Issue #1596, opened 2020-09-08
Setting up `gke2.mybinder.org` (migrate gke.mybinder.org to new project)


betatim:
Wanted to start this issue to discuss the migration from our current GCP projects (binder-prod, binder-staging, binder-sandbox) to the new GCP project we have.

Is there anything to do for binder-sandbox? I've not used it in forever and don't know off things that live there. I assume all of them can be turned off/deleted though. It is a sandbox after all. I would also not re-create a sandbox for now in the new project.

For binder-staging we need to create a new GKE cluster with all the trimmings. A new thing is that we need to pick a prefix for the image registry that doesn't interfere with binder-prod. Right now staging and prod have their own registries. If we can create multiple registries in one project we should consider that. I'd propose we try to set it up as `staging2.mybinder.org`, add it to our `deploy.py` and when we have confidence that it works turn off `staging.mybinder.org`.

For binder-prod we need to do everything we need to do for binder-staging and more:
* setup logging (we have some filters/excludes to reduce ingested log volume to keep costs down)
* setup launch analytics (with support for receiving events from other clusters like OVH and Gesis)
  * transfer the existing events/data from the old analytics instance
* setup matomo
  * do people actively use it? I still use GA on the rare occasion that I look at analytics. If no one loves using it maybe we can save some effort and not migrate it?
  * needs a SQL database and proxy
  * transfer existing data
* transfer the grafana boards/config (we probably don't need the data)
* transfer the federation redirector

I'd propose that we setup the new "prod" instance BinderHub as `gke2.mybinder.org` and add it to `deploy.py` and the federation redirect as soon as we can. Basically we pretend it is just another cluster joining the federation. Then we can start working on setting up/transferring the "support infrastructure" (logging, analytics, etc) one by one. At some point we will have everything running on the new cluster (hopefully).

xref https://github.com/jupyterhub/mybinder.org-deploy/issues/1595 which is about setting up terraform recipes(?) to help setup infrastructure.

What do people think?

---

please edit this post's lists if you want to help flesh them out/correct things/add missing things

manics:
How often do the Grafana dashboards change? For other projects I've uploaded the dashboards to https://grafana.com/grafana/dashboards, the Grafana Helm chart can automatically pull dashboards from there (you pass it the online dashboard id): https://github.com/grafana/helm-charts/tree/main/charts/grafana#import-dashboards

The main downside is it's more work to modify a dashboard.

[...]

Issue #1595, opened 2020-09-04
terraform for deploying clusters


minrk:
Provisioning of cloud resources themselves is not yet automated, which is coming up now as we discuss some possible shifts in deployments which could mean redeploying our GKE clusters. We have manually deployed things (SQL server for matomo, buckets for analytics, clusters) so far, but it would make transitions like this easier if we could use terraform to automate this. If nothing else, it would serve as documentation for what we've actually deployed!

On gitter, folks have been discussing using terraform for this step, and several folks have had positive experiences with this in the last year.

yuvipanda:
Trying to list the bits of cloud infrastructure we need that aren't handled by helm / deploy.py.

1. GKE Cluster itself
2. NodePools for the Cluster
3. Container Registry
4. Stackdriver config for analytics (and buckets too? I'm not sure)
5. SQL Server for Matomo
6. Reserve the load balancer IPs so we don't lose them during cluster re-creates
7. Service Accounts + keys for various things

manics:
I've used it for a test project and liked it's usability. The main downside from a team perspective is it requires a persistent tfstate file that has to be saved after every run and shared amongst anyone who wants to update the deployment. There's several ways to automatically share the tfstate file but I've never used them, this is one of the main reasons I'm not using it in production.

Terraform won't incorporate resources created outside the tfstate file (at least it didn't last time I used it), in contrast to some other tools such as Ansible which take your existing state into account. OTOH if it's designed for "one-off" use where it's only used to create the infrastructure but not update it this could be quite a nice way to get started with a new cluster.

yuvipanda:
> There's several ways to automatically share the tfstate file but I've never used them, this is one of the main reasons I'm not using it in production.

I've used https://www.terraform.io/docs/backends/types/gcs.html for this. Was fairly trivial to set up, and worked pretty flawlessly after!

minrk:
> Terraform won't incorporate resources created outside the tfstate file (at least it didn't last time I used it)

It can with [terraform import](https://www.terraform.io/docs/import/index.html) I used this a couple of times in the spring. It's not perfect and a bit tedious, but it can work.

minrk:
> I've used https://www.terraform.io/docs/backends/types/gcs.html for this

We also stored tfstate on azure storage as soon as we had two folks deploying, and it worked well, so I think we should go the GCS route from the start.

betatim:
[...]

PR #1604, opened 2020-09-09
start deploying staging to new gcp project with terraform

minrk:
so far:

- [x] store tf-state in gcs bucket (bucket itself had to be created manually, I've created buckets for staging and prod)
- [x] GKE cluster for staging
- [x] reserve public ip for staging cluster
- [x] SQL for matomo
- [x] service accounts, roles
- [x] service account keys
- [x] analytics buckets
- [x] stackdriver sink

see #1596, #1595


minrk:
BTW, I'm doing this as simple terraform "main.tf" scripts instead of organizing in modules. When we get to prod, there may be enough in common that we want to do that, but I'm not sure. We don't have that much to deploy.

deploy.py
```diff
@@ -209,36 +249,65 @@ def deploy(release):
...
 
     for d in deployments_and_daemonsets:
-        subprocess.check_call([
-            'kubectl', 'rollout', 'status',
-            '--namespace', release,
-            '--timeout', '5m',
-            '--watch', d
-        ])
+        subprocess.check_call(
+            [
```
minrk, line 299: These indentation changes on the lines I've edited are the result of [darker](https://pypi.org/project/darker/) which applies black formatting *only to changed parts of the code*, which is pretty nifty.

choldgraf:
I know this is almost done, but wanted to ping @salvis2 as I know he's [done some work deploying JupyterHub with terraform](https://medium.com/pangeo/terraform-jupyterhub-aws-34f2b725f4fd)...maybe he has thoughts on this particular deployment

PR #1609, opened 2020-09-12
Limit all jovian.ml for the moment

betatim:
There is no one repo that uses a lot of resources so instead this regex
tries to match every jovian.ml repository and applies a lower limit to
it.

It seems as if almost every launch from jovian.ml comes from a different repository/URL.

It would be better if we could apply a limit to the sum of all repos that match a pattern.

Does someone know who we can contact from the website to let them know before we do this or find a compromise that works for everyone?

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -61,6 +61,9 @@ binderhub:
         - pattern: ^ipython/ipython-in-depth.*
           config:
             quota: 128
+        - pattern: ^jupyter-api-2dgit.*
+          config:
+            quota: 20
 
     GitRepoProvider:
       banned_specs:
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -61,13 +61,11 @@ binderhub:
         - pattern: ^ipython/ipython-in-depth.*
           config:
             quota: 128
-        - pattern: ^jupyter-api-2dgit.*
-          config:
-            quota: 20
 
     GitRepoProvider:
       banned_specs:
         - ^https%3A%2F%2Fbitbucket.org%2Fnikiubel%2Fnikiubel.bitbucket.io.git/.*
+        - ^https%3A%2F%2Fjovian.ml%2Fapi%2Fgit%2F.*
     BinderHub:
       use_registry: true
       build_image: jupyter/repo2docker:0.11.0-141.gf0c6f65


manics:
There's a generic contact email on the website
`hello@jovian.ml`, and also a Discourse forum 
https://jovian.ml/forum/

Linkedin has a list of people (login required) https://www.linkedin.com/company/jovianml, I don't recognise any names.

mybinder/values.yaml
```diff
@@ -61,6 +61,9 @@ binderhub:
         - pattern: ^ipython/ipython-in-depth.*
           config:
             quota: 128
+        - pattern: ^jupyter-api-2dgit.*
```
minrk, line 64: where does this pattern come from? I don't think this will match jovyan.ml repos, which use GitRepoProvider, not GitHub, and start with `https://jovyan.ml/api/git/...` (escaped)

mybinder/values.yaml
```diff
@@ -61,6 +61,9 @@ binderhub:
...
+          config:
+            quota: 20
```
minrk, line 66: title says 30, but this is 20 :)

mybinder/values.yaml
betatim: Ah yes! This came from looking at the pod name but that is the wrong place to look. Just got used to it from github based pods because you can more or less copy and paste it. Will update

charitarthchugh:
Hello, 
I just happened to come across this PR. I think that you should contact Aakash NS.
@aakashns -  Aakash
CEO of JovianML.
https://github.com/aakashns

Thanks!

sidujjain:
Hi, Siddhant from Jovian.ml here. We have been seeing higher traffic from past few weeks which translates to more number of Mybinder runs. We are currently in the process of setting up hosted JupyterHub. That should ease up the load on Mybinder. 

If you guys want to go down the path of rate limiting our repos, that's understandable. As @betatim mentioned applying rate limiting to sum of repos might be a better approach here. If there's a metrics api that gives aggregate count of pods running, we can implement rate limiting on our side. Either way, happy to work together towards a solution.

betatim:
I think the jovian repos actually use the git repo provider so the previous content of this PR was just wrong.

Siddhant we don't really have a public API to query for how many copies of each repo (or group of repos) is running.

Currently there were about 70 instances of different repos from jovian.ml running as well as about 10 builds that had been going for several hours.

In addition we are seeing a much higher number of builds than usual
Screenshot 2020-10-01 at 21 46 56

Experimenting to see if this is the right regex to catch all repos from you guys.

minrk:
Technically, our [prometheus](https://prometheus.mybinder.org) is public and can be accessed, but I wouldn't exactly call it a stable api, though, and is per-federation member. One can see *completed* builds and launches, though:

```
sum(increase(binderhub_launch_count_total{provider="Git",repo=~"https://jovian.ml.*"}[60m]))
sum(increase(binderhub_build_count_total{provider="Git",repo=~"https://jovian.ml.*"}[60m]))
```

which is an okay proxy if builds are rare, however, we critically lack a metric for in- progress builds by repo without querying mangled pod names:

```
sum(kube_pod_status_phase{pod=~"build.*jovian.*"}) by (phase)
```

PR #1620, opened 2020-09-22
send events to new GCP project

minrk:
- adds credentials for events-archivers for each, stored in secrets/events
- uses them in turing, ovh, and old prod config
- fix ACL permissions on log sink buckets, which
- gesis creds file is here, but we don't deploy gesis (@bitnik)

Needs to wait until we can confirm the event-log sink is actually collecting events. Everything I can find in configuration says it should be collecting events, but it hasn't put anything into the buckets yet. There don't appear to be any mechanisms for debugging log sinks. Some manual fiddling gave an error that the earlier uniform permissions on the raw-events bucket was incompatible with log sinks, despite no error in creating the sink via the API. I've fixed the permissions and deleted and recreated the sinks. Hopefully they will collect something now, but the fact that it failed completely silently before doesn't give me a lot of confidence.


minrk:
I think I figured out the permission issue, but we can only know once every hour, when the sink does its thing, since sinks don't validate their own permissions and can't check or be tested in any way.

minrk:
ok, it worked! Sending new events from OVH, gke1, gke2, turing to the new events archive

PR #1629, opened 2020-09-23
decomission binder-staging

minrk:
replace with gke2 staging cluster

This is going to be a practice run for decomissioning the prod cluster.
Deploying this is done in concert with updating DNS for staging.mybinder.org to the new staging cluster.

When we do it for prod tomorrow, we won't *also* lose the separate config at the same time


.github/workflows/cd.yml
```diff
@@ -42,9 +42,6 @@ jobs:
       matrix:
         include:
           - federation_member: staging
-            binder_url: https://staging.mybinder.org
-            hub_url: https://hub.gke.mybinder.org
-          - federation_member: staging-gke2
```
minrk, line 47: Leaving the binder/hub-urls here for the initial deploy since we know DNS won't be transferred for a few minutes

minrk:
DNS has been updated, let's see how long this takes

PR #1641, opened 2020-09-24
[MRG] Change federation redirector to use weighted rendezvous

betatim:
This adds the ability to weight clusters in addition to limiting them by
quota.
 This is nice to spread the "random repos" across clusters.

I checked that it works with things like:
```
In [57]: Counter([rendezvous_rank([('a', 1.), ('b', 1.)], "key-%i" % i)[0] for i in range(100)])
Out[57]: Counter({'a': 48, 'b': 52})

In [60]: Counter([rendezvous_rank([('a', .2), ('b', .8)], "key-%i" % i)[0] for i in range(100)])
Out[60]: Counter({'a': 82, 'b': 18})
```
so I think it works.

If we want to start using weights we need to update the comments/docs that were added in #1639

### DIFF
--- a/images/federation-redirect/README.rst
+++ b/images/federation-redirect/README.rst
@@ -24,3 +24,12 @@ health they are automatically added back to the list of viable targets.
 The redirect response contains a short lived cookie that is used to remember
 the choice of target. This means that subsequent visits from the same user
 agent will be directed to the same target.
+
+
+-------------
+Running tests
+-------------
+
+To run the automated tests for the redirector change to this directory in
+your terminal and run ``pytest``. This should find the tests in
+``test_rendezvous.py``.
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import math
 import os
 import random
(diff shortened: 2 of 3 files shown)


minrk:
Nice! That does indeed look right. Would it be possible to add some tests? Even if you don't run them in the CI, having them there to check the code would be handy.

Before merging, we should note that the old prod cluster is not getting chartpress image updates anymore. That means it can't deploy a new federation-redirect image. This should be fine, as long as we set federationRedirect.enabled = false on the old prod cluster at the same time, since it should no longer be receiving any requests anyway now that mybinder.org is served from gke2.

betatim:
I went with testing the rendezvous rank function only. Is that what you had in mind or should we test more "end to end" (start up the tornado server and make requests to it to see that it does the right thing)?

minrk (approved):
💯 I think this will be super interesting! unittests seem like plenty, and what I was thinking of. I think e2e is not practical for this one.

betatim:
Should we deploy it to see what happens?

I think the weights in the chart right now corresponds to an equal weighting of all clusters which should be a "no change" change.

PR #1646, opened 2020-09-28
update grafana-api-key secret

minrk:
for new grafana deployment

our annotations weren't getting published to the new grafana


minrk:
Allocation of this key is a manual step. I don't see a way to automate its creation

PR #1648, opened 2020-10-01
log reason for failed health checks

minrk:
- log reason on each failure, rather than only logging reason when a member is removed from the federation
- don't retry successful checks with unhealthy status (identified with FailedCheck exception class)
- use `python:3.8-slim-buster` as base image for both images in this repo, reducing image sizes from 890-950MB to 120-150MB.

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 import os
 import random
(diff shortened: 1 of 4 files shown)


images/federation-redirect/app.py
```diff
@@ -208,7 +215,8 @@ def initialize(self, active_hosts):
         self.active_hosts = active_hosts
 
     async def get(self):
-        self.write({"active_hosts": self.active_hosts})
+        self.set_header("Content-Type", "application/json")
+        self.write(json.dumps({"active_hosts": self.active_hosts}))
```
betatim, line 219: Tornado should do the serialisation and content-type header automagically when you `write` a dictionary.

minrk: 👍 reverted

PR #1675, opened 2020-10-27
remove heritage from matchLabels for gcs-proxy

minrk:
I think this is causing current lack of availability for archive.analytics


minrk:
https://github.com/jupyterhub/mybinder.org-deploy/commit/8ca1454ef07e6b2d1ce4141523383b642679c7ca was incomplete - pod target removed the label, but service did not remove it from its selector

Issue #1672, opened 2020-10-27
Fix lint and validate workflow


sgibson91:
Our helm lint and validate workflow is having issues after some upgrades to the chart and migrating to helm3

minrk:
I think this may have had to do more with making a PR from a branch of this repo instead of a fork, since it looks to be passing on both new PRs and master. Maybe need to check our logic for what happens when a PR is issued from within the repo. Errors looked like it *expected* the secrets to be unlocked, but the unlock step didn't happen. So maybe our conditions aren't consistent between those two steps.

sgibson91:
I'm not sure. [This run](https://github.com/jupyterhub/mybinder.org-deploy/actions/runs/331113119) came from a branch within the repo so it was correct to unlock the secrets with git-crypt, there was just something wrong with the charts at that particular moment. And actually this `error calling required: Proxy token must be a 32 byte random string generated with 'openssl rand -hex 32'!` is the problem I was facing when migrating Hub23 to helm3 and ended up deleting and redeploying from scratch.

sgibson91:
For clarity: I encountered this exact same error at this stage of the migration even though I had a local, plaintext values file with unencrypted secrets in it.

PR #1682, opened 2020-10-30
use newer github style to set environment variables

minrk:
shouldn't change how anything works, but I noticed that our CI is getting warnings that set-env will be removed: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

### DIFF
--- a/.github/workflows/lint-validate.yml
+++ b/.github/workflows/lint-validate.yml
@@ -46,7 +46,7 @@ jobs:
             export YAMLLINT_CONFIG=scripts/yamllint-no-secrets.yaml
           fi
           # persist env for the next step
-          echo "::set-env name=YAMLLINT_CONFIG::$YAMLLINT_CONFIG"
+          echo "YAMLLINT_CONFIG=$YAMLLINT_CONFIG" >> "$GITHUB_ENV"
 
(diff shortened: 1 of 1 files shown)


manics (approved):

consideRatio:
I think this PR broke the lint jobs somehow, see: https://github.com/jupyterhub/mybinder.org-deploy/pull/1685/checks?check_run_id=1334528674

image

manics:
The checks were passing in this PR, what do you think causes them to fail in the other PR?

consideRatio:
@manics because this PR changed the workflow file, and perhaps the workflow used is that which is in the master branch rather than suggest in the PR? I'm not sure... But, it is very weird if a revertion PR that restores the state would fail, and this PR seem to fit the bill with regards to causing the observed failure. I'm not 100% though because I don't understand much about this.

:'( Yikes, I'm tech fatigued at the moment --- too much complexity

sgibson91:
I have a suspicion there is some flakiness going on with encoded variables in the chart as this is the exact same error I came across when upgrading Hub23 to use Helm3 https://github.com/jupyterhub/mybinder.org-deploy/issues/1672

PR #1689, opened 2020-11-02
Remove Turing from federation and CI

sgibson91:
This PR removes the Turing cluster from the federation while I upgrade it from helm v2 to v3. It also removes it from the CI matrix to avoid clashing with some concurrent debugging.


minrk:
can you also set weight: 0 for turing in mybinder/values.yaml? That will remove it from the federation-redirect health checks

sgibson91:
Done!

PR #1690, opened 2020-11-02
Update CI to install helm3 for Turing cluster

sgibson91:
This PR reintroduces the Turing cluster into CI (but not the federation or redirector) and handles installing helm v2 or v3 depending on the cluster matrix values.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -213,8 +213,15 @@ jobs:
           version: ${{ env.KUBECTL_VERSION }}
 
       - name: "Stage 1: Install helm ${{ env.HELM_VERSION }}"
+        if: matrix.helm_version == env.HELM_VERSION
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -209,17 +212,25 @@ jobs:
         with:
           version: ${{ env.KUBECTL_VERSION }}
 
-      - name: "Stage 1: Install helm"
+      - name: "Stage 1: Install helm ${{ env.HELM_VERSION }}"
```
minrk, line 215: would it work to do something like `${{ env.HELM_VERSION || matrix.helm_version}}` to avoid duplication of the helm steps?

Or is that not worth it, since the duplicates will just go away when you're done with the migration?

sgibson91: > Or is that not worth it, since the duplicates will just go away when you're done with the migration?

This, and the steps aren't strict duplications. For instance, the URL we download the binary from is slightly different for v3 than v2.

PR #1694, opened 2020-11-02
add health checks to deployment tests

minrk:
Do not consider a deployment failing health checks to be a success

I believe this would have caught the recent deployment failures and prevented them from deploying across the whole federation. My suspicion is that staging would still have succeeded while prod would have failed, since it appears to be a performance issue, but we would at least have been notified of the problem


sgibson91 (approved):
👩‍⚕️

PR #1695, opened 2020-11-02
Reflect new project name in SRE docs

sgibson91:
When we migrated Google projects, we now only have one, called
"binderhub", as opposed to two. The production cluster also has a
different name and is in a different zone to staging.

This PR attempts to update the SRE docs to catch changes in our
workflow to the references project ID and cluster name/zone.

I may not have caught all places where changes are pertinent, please feel free to add to this PR if you spot anything!

### DIFF
--- a/docs/source/operation_guide/command_snippets.md
+++ b/docs/source/operation_guide/command_snippets.md
@@ -36,8 +36,8 @@ It should take a couple of minutes.
 To upgrade the master version with `gcloud`:
 
 ```bash
-gcloud --project=binderhub container clusters upgrade staging --master --zone=us-central1-a
-gcloud --project=binderhub container clusters upgrade prod --master --zone=us-central1
+gcloud --project=binderhub-288415 container clusters upgrade staging --master --zone=us-central1-a
+gcloud --project=binderhub-288415 container clusters upgrade prod --master --zone=us-central1
 ```
 
 Now we can start the process of upgrading node versions, which takes more time.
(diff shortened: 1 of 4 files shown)


docs/source/operation_guide/command_snippets.md
```diff
@@ -36,8 +36,8 @@ It should take a couple of minutes.
 To upgrade the master version with `gcloud`:
 
 ```bash
-gcloud --project=binder-staging container clusters upgrade staging --master --zone=us-central1-a
-gcloud --project=binder-prod container clusters upgrade prod-a --master --zone=us-central1-a
+gcloud --project=binderhub container clusters upgrade staging --master --zone=us-central1-a
```
minrk, line 39: These gcloud `--project` args need the numerical id, which is `binderhub-288415`

Otherwise 👍

docs/source/operation_guide/command_snippets.md
```diff
@@ -130,24 +130,24 @@ First we'll create variables that point to our old and new node pools to make it
...
 ```
 
 Then we can create the new user pool:
 
 ```bash
 # create the new user pool
-gcloud beta --project=binder-prod container node-pools create $new_user_pool \
-    --cluster=prod-a \
-    --zone=us-central1-a \
+gcloud beta --project=binderhub container node-pools create $new_user_pool \
```
minrk, line 148: I need to rewrite this! It's going to be totally different now that we use terraform. We will no longer be managing node pools through gcloud CLI, instead we will be editing the terraform config and using `terraform apply`

docs/source/operation_guide/command_snippets.md
sgibson91: Done in 4daab5a

minrk:
Thanks!

PR #1698, opened 2020-11-03
Add GESIS team to list of people with secrets access

betatim:
@bitnik now has a copy of the key.

Once he has things working and can decrypt the secrets he should post a message here and we merge this.

### DIFF
--- a/docs/source/getting_started/production_environment.md
+++ b/docs/source/getting_started/production_environment.md
@@ -278,6 +278,7 @@ People who currently have the git-crypt secret include:
 - @choldgraf
 - @mael-le-gal
 - @sgibson91
+- @bitnik
 - *add yourself here if you have it*
 
 Contact one of them if you need access to the git-crypt key.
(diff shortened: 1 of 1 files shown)


choldgraf:
I'm not sure if @bitnik 's 👍 means "all is working" or "message received and I'll let you know" 😄

bitnik:
It was "message received and I'll let you know" :) and now I can confirm that I have the key, used it to decrypt the secrets and it worked.

-------

I have one question. Since we needed that key as GESIS Notebooks team, should we add GESIS Notebooks into the list? I would like to add the key into where we save secrets of GESIS Notebooks. What do you think?

bitnik:
Btw with GESIS Notebooks team I mean @arnim and @MridulS.

betatim:
I think changing the entry to reflect who has access is a good idea. Maybe we can list the individual usernames from the team? That way it is clear how to contact people. "GESIS team" is harder to turn into a email/GitHub notification.

bitnik:
This works for us. I just shared the key with Arnim and Mridul.

betatim:
Mergegd

Issue #1701, opened 2020-11-11
ESiWACE2 DSL Training for climate & weather model developers


BenWeber42:
Dear mybinder.org developers,

I'd like to request an increase in resources for our upcoming training course :)
### 1. Who you are

[MeteoSwiss](https://www.meteoswiss.ch)

### 2. A few details about the event/course/context

This is a training course targeting developers of climate & weather models. The goal is to show domain scientists how domain specific languages can be used to gain performance portability without sacrificing productivity. Performance portability allows to run bigger simulations on stronger machines which leads to improved weather forecasting and climate predictions.

The training course is offered by ESiWACE (Centre of Excellence in Simulation of Weather and Climate in Europe)

More specifically:
https://www.esiwace.eu/events/esiwace2-dsl-training/esiwace-2-domain-specific-language-dsl-training

### 3. If money is changing hands (attendees pay a fee, speaker is getting paid, free, etc)

Free/None

### 4. A link to the repository you are using

https://github.com/dawn-ico/dusk-dawn-notebook

Related Software:
dusk: https://github.com/dawn-ico/dusk
dawn: https://github.com/MeteoSwiss-APN/dawn

### 5. What type of event it is:

- [ ] talk
- [ ] workshop
- [ ] lecture course
- [x] training session

### 6. How many people you expect to attend the event

We will end up with roughly 90 - 110 registered participants. Of course, it's unlikely that all will turn up, but I think 60 - 80 people using the notebooks is realistic.

### 7. The exact times and dates of the event(s) (with timezone information)

Nov 25: 13:30 - 18:00
Nov 26: 09:00 - 18:00
Nov 27: 09:00 - 15:00
All times in Europe/Vienna / UTC100

### 8. How do you plan to give mybinder.org a shout out at the event

We will have a small slide set introducing how to use mybinder to do the exercises. One slide is dedicated to acknowledge and thank for your support. It currently contains the four logos of the binderhub federation (https://mybinder.readthedocs.io/en/latest/about/federation.html). Not sure if I'm missing important contributors. Guidance here is very much welcome :)

Thanks a lot for all the support so far and the great service!

choldgraf:
[...]

PR #1714, opened 2020-12-01
Remove resources for dawn-ico/dusk-dawn-notebook

BenWeber42:
The ESiWACE2 DSL workshop has concluded now. I don't know if you want to keep this entry for the future. I think for maintainability, it's easier to keep the increased resources _clean_. From our side, we would need such an increase probably at maximum once per year. And I think this time we didn't even come close to hitting the 100 concurrent users. :sweat_smile: 

Based on preliminary discussions and feedback, we can say that climate and meteorological scientists really enjoyed working with the jupyter labs and notebooks. Additionally, mybinder was great since it was so convenient for participants to get access to the exercise tools (instead of having to setup complicated dependencies themselves). We were very happy to be able to use mybinder :smile: (feel free to contact me if you want a proper _reference_)

Related #1701

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -73,9 +73,6 @@ binderhub:
         - pattern: ^michaelJwilson/DESI-HighSchool.*
           config:
             quota: 200
-        - pattern: ^dawn-ico/dusk-dawn-notebook.*
-          config:
-            quota: 128
 
     GitRepoProvider:
       banned_specs:


BenWeber42:
Tagging @betatim here as well :)

choldgraf:
thanks so much @BenWeber42 for following up, and we're super pleased to hear that folks enjoyed using Binder 🎉

Issue #1468, opened 2020-06-17
DNS resolution of *.mybinder.org on turing


minrk:
I was trying to demo [binderlyzer](https://gke.mybinder.org/v2/gh/betatim/binderlyzer/c956adfbabcf7cebc2b961da38cf47c342c148e5?filepath=binder-launches.ipynb) and was assigned to the turing cluster. The notebook fetches from https://archive.analytics.mybinder.org and it failed DNS resolution for mybinder.org. I tested a few mybinder.org subdomains, including the top-level mybinder.org, and it couldn't resolve any mybinder.org domain. Manually assigning the pod to gke or ovh (didn't try gesis) doesn't have the same problem, so it appears to be a cluster DNS issue with turing specifically.

Simple test:

    !wget https://mybinder.org

will hang at `Resolving mybinder.org (mybinder.org)... `

I'm not sure what's up there, maybe something with cert-manager?

consideRatio:
Hmmm perhaps kubectl exec into the hub pod and try ’nslookup mybinder.org 8.8.8.8’ and then ’nslookup mybinder.org’ and if the latter doesnt work inspect the k8s clusters DNS server and potentially network policies that could block/allow access to it. The DNS server can sometimes be found as the service kube-dns.kubesystem even though coredns is now used etc to be the kubernetes clusters dns server.

Kubectl get pods -n kube-system (look for kube-dns or coredns)
Kubectl svc -n kube-system (look for kube-dns or coredns etc)
Kubectl get netpol --all-namespaces (look for rules saying what outgoing traffic for the hub pod is allowed while testing dns from the hub pod)

betatim:
From the hub pod `dig mybinder.org @8.8.8.8` and `dig mybinder.org` both resolve to 35.202.202.188. The singleuser pods don't have `dig` or `nslookup` installed so I am not sure how to test on one of them.

betatim:
There are two coredns pods running in `kube-system`. They've both been up for about 30h.

The log of the coredns autoscaler pod:
```
I0616 07:59:25.009028       1 k8sclient.go:277] Cluster status: SchedulableNodes[8], SchedulableCores[32]
I0616 07:59:25.010202       1 k8sclient.go:278] Replicas are not as expected : updating replicas from 2 to 3
I0616 09:23:35.006851       1 k8sclient.go:277] Cluster status: SchedulableNodes[7], SchedulableCores[28]
[...]

PR #1715, opened 2020-12-01
Disable z2jh singleuser network policy

sgibson91:
related: #1468

cc: @manics @consideRatio

We've identified a conflict between the singleuser and binder-users network policies which is causing the DNS problem we see on the Turing cluster. This PR is suggesting that we disable the restrictive z2jh singleuser policy set in the binderhub helm chart for the time being for the Turing cluster **only**.

This means that singleuser pods will only have a single policy applied to them (binder-users) instead of two (binder-users and singleuser).

Further investigation needs to go into **why** this is happening? The network controller? The Turing cluster runs Azure's CNI, maybe this will be fixed by switching to calico/kubenet/something else? But that would require a complete redeployment of the infrastructure. That is to say, I don't think the CNI can be switched out in place. Docs: https://docs.microsoft.com/en-us/azure/aks/use-network-policies

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -70,6 +70,10 @@ binderhub:
         enabled: true
         replicas: 5
 
+    singleuser:
+      networkPolicy:
+        enabled: false
+
   imageCleaner:
     # Use 40GB as upper limit, size is given in bytes
     imageGCThresholdHigh: 40e9


consideRatio:
@sgibson91 @manics nice work isolating this!

I want to advertise that JupyterHub 0.10.X will have a different default singleuser network-policy, which will include DNS stuff I think.

manics:
That would allow DNS, but it wouldn't allow traffic to other ports that should be allowed.

betatim:
Nice work! I read the summary of your investigation in the issue. I think the z2jh policy we were applying here can be summarised as "allow no outgoing connections" and the binderhub policy explicitly allows connections to a list of ports at any IP address. My question: do we know if not applying the z2jh policy on turing will lead to (1) "all connections are allowed to any port at any IP" (because we don't explicitly block egress) or will we get (2) "everything not explicitly allowed will be blocked" or (3) do we not quite know and need to check once deployed?

I think either outcome is good for me, so this is more for my education than anything else.

consideRatio:
@betatim if a network policy declaring egress and/or ingress rules targets the a pod, it is by a functional NetworkPolicy enforcer thingy, supposed to isolate that pods egress and/or ingress entirely. It is my understanding that removing the Z2JH netpol will reduce permissions only with regards to egress because it was already targeted by the binder-users netpol also declaring egress rules. But, for ingress, I believe the singleuser pods are now perhaps no longer targeted by any ingress rule, and therefore incoming connections to the pods will be allowed.

So to conlcude, I think for ingress, (1) applies now because singleuser pods ingress are no longer targeted by any netpol. But, for egress, (2) applies, as the singleuser pods are still being targeted by the `binder-users` netpol.

minrk:
Update: the issue appears to have been that binder-users was not applied at all, so we have (1) for both ingress and egress

PR #1718, opened 2020-12-02
Change OVH's version of Helm in CI matrix job

sgibson91:
This PR is bumping the version of helm used during OVH's upgrade to helm 3

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -169,7 +169,7 @@ jobs:
           - federation_member: ovh
             binder_url: https://ovh.mybinder.org
             hub_url: https://hub-binder.mybinder.ovh
-            helm_version: "v2.16.10"
+            helm_version: "v3.3.4"
 
     steps:
       - name: "Stage 0: Checkout repo"


consideRatio:
Nice work keeping this updated @sgibson91 ❤️

betatim:
Looks like something broke when deploying this :-/

https://github.com/jupyterhub/mybinder.org-deploy/runs/1489332291

sgibson91:
Yeah, is it the pvcs? That always happens. We tend to manually delete them and re-run the job

betatim:
I don't think it is related to PVCs. This is the error message:
```
Error: UPGRADE FAILED: cannot patch "proxy-patches" with kind Deployment: Deployment.apps "proxy-patches" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"app":"proxy-patches", "component":"proxy-patches", "heritage":"Helm", "hub.jupyter.org/network-access-proxy-api":"true", "release":"ovh"}: `selector` does not match template `labels`
```

sgibson91:
Yes, but the labels it can't patch have been on the pvcs previously

sgibson91:
Hmmm, this is confusing me (now that I've sat down at an actual computer instead of squinting at my phone)

sgibson91:
I _think_ it might be related to this PR thoughhttps://github.com/jupyterhub/mybinder.org-deploy/pull/1675

sgibson91:
Managed to get a new error

```
Error: UPGRADE FAILED: cannot patch "proxy-patches" with kind Deployment: Deployment.apps "proxy-patches" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"proxy-patches", "component":"proxy-patches", "heritage":"Helm", "hub.jupyter.org/network-access-proxy-api":"true", "release":"ovh"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

Managed this by adding

```
heritage: {{ .Release.Service }}
hub.jupyter.org/network-access-proxy-api: "true"
```

to this block:

https://github.com/jupyterhub/mybinder.org-deploy/blob/444ef2d37447feb81866018628e4b67796b933f1/mybinder/templates/proxy-patches/deployment.yaml#L14-L16

consideRatio:
@sgibson91 the selector.matchLabels should not contain `hub.jupyter.org/network-access-proxy-api: "true"`.

Any change to the matchLabels will require a recreation of the deployment resource, and the labels in there should be limited to app/component/release which are enough to pinpoint the specific pod.

The #1675 PR would certainly have caused the same kind of issue because it changes the selector.matchLabel content of a deployment resource. The need for that PR was because the heritage label were changing between versions of helm2 and helm3, and should not have been part of the selector.matchLabels field in the first place according to Helm's best practices.

To resolve this issue, I suggest to manually delete the deployment resource, and then deploy again. If using Helm2, I think `--force` would have done the trick, but it won't do the trick in Helm3 which doesn't provide a sensible option to replace the `helm2 --force`.

sgibson91:
`kubectl get Deployment/proxy-patches -o yaml -n ovh | sed "s/Tiller/Helm/g" | kubectl replace --force=true -f -` also worked

Now I'm at the pvc error that I was expecting

sgibson91:
@consideRatio and I fixed this by restarting some deployments to force them the create new labels which would be immutable while the deployments still existed. Specifically, grafana and prometheus deployments. There is ongoing work in #1720 and jupyterhub/binderhub to get our CI green again involving some test logic and a regression in chartpress meaning that the image containing a patch wasn't available to be pulled onto the cluster.

PR #1722, opened 2020-12-03
Bump chartpress to 1.0.*

consideRatio:
This bump is not needed for anything specific, it's just maintenance. Since I know its a safe version upgrade I figure it makes sense for me to suggest it so others don't have to read the changelog etc.


sgibson91 (approved):
🎉

PR #1712, opened 2020-11-23
binderhub: 0.2.0-n412.h1f1f0c4...0.2.0-n423.ha36348e

henchbot:
This is a binderhub version bump. See the link below for a diff of new changes:

https://github.com/jupyterhub/binderhub/compare/1f1f0c4...a36348e 

Associated PRs:
- [MRG] Use slim-buster as base image for binderhub and image-cleaner [#1158](https://github.com/jupyterhub/binderhub/pull/1158)

PR #1724, opened 2020-12-03
Revert "Bump chartpress to 1.0.*"

sgibson91:
Reverts jupyterhub/mybinder.org-deploy#1722

With chartpress==1.0.*, we see the SemVer error when building the chart. Reverting back to 0.7 for the time being.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,4 @@ pytest-timeout
 pytest-xdist
 requests
 kubernetes
-chartpress==1.0.*
+chartpress<1.0.0


consideRatio:
@sgibson91 give me a second to spot why this is the case and we go with fixing it instead.

consideRatio:
Let's revert this for the time being. I'm opening a new issue in this repo with a suggested action to fix the error.

betatim:
A few days ago there was an issue with chartpress 1.0 which is why it got pinned https://github.com/jupyterhub/mybinder.org-deploy/pull/1712 might be relevant for understanding what happened here.

consideRatio:
@betatim I see from the logs it was the same issue.

> ValueError: The version in Chart.yaml 'prod-n2671.h28b9db9' isn't SemVer2 compliant as required by Helm 3!

I think it's mostly correct by chartpress to say this, as the mybinder.org-deploy helm chart would fail be packaged and published by helm3, but it isn't, so in a way chartpress is a bit strict failing even before that is done.

Let's bring the discussion to #1725

PR #1726, opened 2020-12-03
ci: use non-deprecated repo for setup-gcloud action

consideRatio:
I noticed warnings about setup-gcloud action being deprecated, and that they recommend a new version. This PR does that.

image


sgibson91 (approved):
💖

PR #1730, opened 2020-12-04
[MRG] Expose nodeSelector config

### LINKED ISSUE #1729: Deploy reported failure because pods scheduled on user nodes got their image pulling queued
## Issue summary

A recent deployment to gke-prod failed because both got pods described below got stuck pulling images and failing to become ready. This in turn caused `helm upgrade --wait` deem the upgrade to have failed even though it succeeded in the end after a while longer wait.

### analytics-publisher

```
Events:
  Type    Reason     Age    From                                         Message
  ----    ------     ----   ----                                         -------
  Normal  Scheduled  7m35s  default-scheduler                            Successfully assigned prod/analytics-publisher-77cbc7c5c6-22vjv to gke-prod-user-202009-b9c03ca0-v8jc
  Normal  Pulling    7m32s  kubelet, gke-prod-user-202009-b9c03ca0-v8jc  Pulling image "gcr.io/binderhub-288415/mybinder-tool-analytics-publisher:2020.12.4"
  Normal  Pulled     27s    kubelet, gke-prod-user-202009-b9c03ca0-v8jc  Successfully pulled image "gcr.io/binderhub-288415/mybinder-tool-analytics-publisher:2020.12.4"
  Normal  Created    26s    kubelet, gke-prod-user-202009-b9c03ca0-v8jc  Created container analytics-publisher
  Normal  Started    26s    kubelet, gke-prod-user-202009-b9c03ca0-v8jc  Started container analytics-publisher
```

### federation-redirect

```
Events:
  Type    Reason     Age    From                                         Message
  ----    ------     ----   ----                                         -------
  Normal  Scheduled  9m1s   default-scheduler                            Successfully assigned prod/federation-redirect-57d8f4767b-qh8gh to gke-prod-user-202009-b9c03ca0-v8jc
  Normal  Pulling    8m58s  kubelet, gke-prod-user-202009-b9c03ca0-v8jc  Pulling image "gcr.io/binderhub-288415/mybinder-tool-federation-redirect:2020.12.4"
  Normal  Pulled     82s    kubelet, gke-prod-user-202009-b9c03ca0-v8jc  Successfully pulled image "gcr.io/binderhub-288415/mybinder-tool-federation-redirect:2020.12.4"
  Normal  Created    81s    kubelet, gke-prod-user-202009-b9c03ca0-v8jc  Created container federation-redirect
  Normal  Started    80s    kubelet, gke-prod-user-202009-b9c03ca0-v8jc  Started container federation-redirect
```

## Issue comprehension

Why did these images take so long time to pull? It is because a node will pull docker images one at the time for the pods that's starting up on it. I assume there were pods starting up on this node slightly earlier than these pods were scheduled on the nodes, which caused them to need to wait even though their image sizes are small (~50 MB each).

There are two potential fixes, and we can apply either or both as I see it.
- [ ] Schedule federation-redirect and analytics-publisher pods on non-user nodes (#1730)
- [x] Increase the timeout for `helm --wait` to at least 10 minutes as these images took ~9 minutes to pull (#1731)

consideRatio:
This PR makes nodeSelector configurable for: analytics-publisher, federation-redirect, matomo's mysqld-exporter.

It does not firmly close #1729 as initially intended as we opted to not make them schedule on the core node pool, but with #1731 we should be fine anyhow with regards to that.

---

I inspected these pods resource requests in k8s and how much free resources were around on the core nodes. They didn't request anything which means in k8s cluster they are unlimited I think. The nodes seem to have some room for these pods overall.

Assuming k8s node upgrades are done by adding a node and then removing a node using _node surge upgrades_, there should be no issues then either I think.

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -183,6 +183,7 @@ matomo:
       memory: 1Gi
 
 analyticsPublisher:
+  nodeSelector: *coreNodeSelector
   project: binderhub-288415
   destinationBucket: binder-events-archive
   events:
@@ -191,6 +192,7 @@ analyticsPublisher:
     sourceBucket: binder-billing-archive
 
 gcsProxy:
+  nodeSelector: *coreNodeSelector
   buckets:
     - name: binder-events-archive
       host: archive.analytics.mybinder.org
@@ -198,4 +200,5 @@ gcsProxy:
       host: archive.analytics.gke2.mybinder.org
 
 federationRedirect:
+  nodeSelector: *coreNodeSelector
   enabled: true
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -183,7 +183,6 @@ matomo:
       memory: 1Gi
 
 analyticsPublisher:
-  nodeSelector: *coreNodeSelector
   project: binderhub-288415
   destinationBucket: binder-events-archive
   events:
@@ -192,7 +191,6 @@ analyticsPublisher:
     sourceBucket: binder-billing-archive
 
 gcsProxy:
-  nodeSelector: *coreNodeSelector
   buckets:
     - name: binder-events-archive
       host: archive.analytics.mybinder.org
@@ -200,5 +198,4 @@ gcsProxy:
       host: archive.analytics.gke2.mybinder.org
 
 federationRedirect:
-  nodeSelector: *coreNodeSelector
   enabled: true
--- a/mybinder/templates/analytics-publisher/deployment.yaml
+++ b/mybinder/templates/analytics-publisher/deployment.yaml
@@ -23,6 +23,7 @@ spec:
       annotations:
         checksum/configmap: {{ include (print $.Template.BasePath "/analytics-publisher/configmap.yaml") . | sha256sum }}
     spec:
+      nodeSelector: {{ .Values.analyticsPublisher.nodeSelector | toJson }}
       volumes:
       - name: secrets
         secret:
--- a/mybinder/templates/federation-redirect/deployment.yaml
+++ b/mybinder/templates/federation-redirect/deployment.yaml
@@ -21,6 +21,7 @@ spec:
       annotations:
         checksum/configmap: {{ include (print $.Template.BasePath "/federation-redirect/configmap.yaml") . | sha256sum }}
     spec:
+      nodeSelector: {{ .Values.federationRedirect.nodeSelector | toJson }}
       affinity:
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
--- a/mybinder/templates/matomo/mysqld-exporter/deployment.yaml
+++ b/mybinder/templates/matomo/mysqld-exporter/deployment.yaml
@@ -28,6 +28,7 @@ spec:
         component: mysqld-exporter
     spec:
       automountServiceAccountToken: false
+      nodeSelector: {{ .Values.matomo.nodeSelector | toJson }}
       volumes:
       - name: cloudsql-instance-credentials
         secret:
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -452,6 +452,7 @@ analyticsPublisher:
     sourceBucket: binder-billing
     fileName: cloud-costs.jsonl
     kind: csv
+  nodeSelector: {}
 
 # this is defined in secrets/ for the OVH cluster
 eventsArchiver:
@@ -493,3 +494,4 @@ federationRedirect:
       weight: 1
       health: https://turing.mybinder.org/health
       versions: https://turing.mybinder.org/versions
+  nodeSelector: {}


betatim:
Why limit the pods to core nodes instead of allowing them to run on any of the nodes?

We upgrade node pools according to https://mybinder-sre.readthedocs.io/en/latest/operation_guide/command_snippets.html#upgrading-kubernetes which involves cordoning nodes after adding the new node pools.

consideRatio:
I saw this comment first where I discussed why I think a node pool pin is preferred: https://github.com/jupyterhub/mybinder.org-deploy/issues/1729#issuecomment-738879729

consideRatio:
To conclude, I learn from the analysis of their configured replicas and PDBs, that they are not blocking downscaling which would be unacceptable if they were allowed to schedule on user pods in my mind.

## Replicas / PDB analysis - OK to schedule on all nodes?

Without having the right replica/PDB settings, we would potentially block the nodes downscaling so I wanted to make sure we don't.

I analyzed the following deployments replica count and PDB configuration. Autoscaling would be blocked with a single 1 replica and 0 minAvailable in the PDB. Something I championed we avoid by default for Z2JH in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1934.

- analytics-publisher, staging/prod only, 1 replicas, 0 minAvailable --- OK
- federation-redirect, staging/prod only, 2 replicas, 1 minAvailable --- OK
- gcsProxy, staging/prod only, 1 replicas, 0 minAvailable --- OK

This is not an issue, good!

## Pro/con

With regards to compute resources, there are no objective pro/con in my mind other than not risking to cause #1729 again.

### Pros of no nodeSelector
- Not sure

### Cons of no nodeselector / ## Pros of nodeSelector
- Will risk causing / won't cause issue #1729 again.

### Cons of nodeSelector
- Not sure

## Action points
- [ ] Merge
- [ ] Request I update the PR to not pin federation-redirector/gcs-proxy/analytics-publisher on core nodes
- [ ] Request input from someone else

betatim:
Overall I'd veto a change that increases the reliance on core nodes and their "special status" in the cluster. I'd welcome work that reduces the need for having core nodes.

Moving more pods to the core nodes will not significantly reduce the number of times a deploy run is incorrectly marked as failed. It will increase the specialness of the core nodes which has caused problems in the past which were harder to debug than "restart the CI job".

Overall I think this is a case of https://how.complexsystems.fail/#14

---

At the very beginning of mybinder.org we had a core pool and a user pool for nodes. We thought this was a smart idea. Over time we learnt that it wasn't as smart as we thought. It works against the principle of "cattle, not pets" because there are pods which can only run on certain nodes, which are different from most other nodes. This has lead us to transition to a "no special nodes" policy. There are still core nodes but pods that could be moved/someone has moved were moved to run "anywhere". 

Some problems we have had in the past include that the core node pool runs out of resources leaving pods unschedulable during upgrades, the core node pool running with a lot of "slack" because a new service got added that causes a scale up of the core node pool resulting in several nodes running at 50% utilisation. Or we ended up with just one node which got in the way of anti-affinity rules for pods with several replicas. As well as other things caused by the fact that having two node pools is more complex than having one, having pods that can only run on one node pool is more complex than having all pods run on all nodes.

Overall the current setup has served us well and we have had few operational issues. Neither operators causing outages by making a mistake nor node failures causing outages. We also don't observe any issues with the auto-scaling behaviour of the cluster. We don't observe very high (overloaded) or very low (underloaded) CPU/memory loads on our user nodes. This had happened before when the auto-scaling wasn't working as well as it does now.

The fact that one deploy failed to report the correct status at the end of the CI run (but didn't cause any problems for users) is at most annoying from an operations point of view. The circumstances that lead to the timeout in the deploy are rare and addressed by increasing the timeout. The fix and default action is to re-run the CI job. While it is annoying to have to restart the job, there are many other reasons we have to do this, so it is a routine practice that operators are familiar with.

betatim:
I just looked at what we are running on the core node (there is only one right now 🤭) and I think most of it could be running somewhere else. Some of the things which are running there right now (e.g. federation redirector) are using the node not because they have to but because they can. The only two things we'd have to think about are the JupyterHub pod and the two binderhub pods.

---

(really a new issue but I'm lazy)

This means removing the need for core nodes is achievable 🎉. I think the hub pod will always be special because it can only have one replica. The bhub deployment already runs with 2 replicas. I think if we modify binderhub so that it does not exit until all (reasonable) HTTP requests are completed we could move it to the user pool as well. Streaming off the build logs is about the only long running request we have. This is what we should ponder a bit before making any changes.

consideRatio:
@betatim I understand your concerns. I believe its fine to not pin these pods as I suggested. I have ensured for myself that we won't have issues associated for dynamic downscaling of user nodes due to them given our current settings. To me, perhaps the key benefit was to not need to ensure this by pinning them to the core node pool, or risk getting such issues by a change to the replicas/pdb.

I have updated the PR to no longer pin but simply add missing configuration ability to be in parity with the other deployments etc configuration options now. This PR is now harmless.

consideRatio:
@betatim I'm strongly against removing the node pool due to the complexity of ensuring that everything installed will scale down reliably without manual interventions, and because movement of our single point of failure pods like proxy/hub will cause disruptions when moved around or block downscaling if not moved around.

consideRatio:
Triaging my own PRs, this is a really safe change as it does nothing except expose configuration so I'll go for it.

consideRatio:
```
Events:
  Type    Reason     Age    From                                         Message
  ----    ------     ----   ----                                         -------
  Normal  Scheduled  18m    default-scheduler                            Successfully assigned prod/analytics-publisher-6784548cb6-r67rr to gke-prod-user-202009-b9c03ca0-tg2p
  Normal  Pulling    18m    kubelet, gke-prod-user-202009-b9c03ca0-tg2p  Pulling image "gcr.io/binderhub-288415/mybinder-tool-analytics-publisher:2020.12.4"
  Normal  Pulled     4m42s  kubelet, gke-prod-user-202009-b9c03ca0-tg2p  Successfully pulled image "gcr.io/binderhub-288415/mybinder-tool-analytics-publisher:2020.12.4"
  Normal  Created    4m42s  kubelet, gke-prod-user-202009-b9c03ca0-tg2p  Created container analytics-publisher
  Normal  Started    4m40s  kubelet, gke-prod-user-202009-b9c03ca0-tg2p  Started container analytics-publisher
```

Deployment failed according to CI, but worked. The reason for the failure was that waiting 10 minutes didn't do the trick as there was too long time wait for the image to pull on the user node it was scheduled. In other words, #1731 didn't cut it.

This was the original reason for me to open this PR in the first place =/ Let me know if you withdraw your veto at some point @betatim.

betatim:
> movement of our single point of failure pods like proxy/hub will cause disruptions when moved around or block downscaling if not moved around.

I think this is why we had PDBs protecting them but apparently that is unusual and got removed from z2jh :p

My point was that it looks like with some work it is possible to achieve a "no core nodes" setup. I am not super worried about scale down because (1) we should protect the hub and proxy pods with a PDB so they don't get moved and (2) the mybinder.org cluster never scales down below 3-4 nodes. So while a hub pod might block scaling down of a node, it isn't a big deal because the minimum size of the cluster is larger than the number of hub pods. (This is of course not true for all binderhub deployments but here we only care about mybinder.org)

consideRatio:
>>movement of our single point of failure pods like proxy/hub will cause disruptions when moved around or block downscaling if not moved around.

>I think this is why we had PDBs protecting them but apparently that is unusual and got removed from z2jh :p

@betatim mybinder.org-deploy has deviated from the z2jh defaults already. The defaults i updated in z2jh reflects the kind of settings mybinder.org-deploy have been using that does not always force manual deletion of hub/proxy pods. The reason for them not having downtime relates to the node pool is reliable and doesn't evict them.

> My point was that it looks like with some work it is possible to achieve a "no core nodes" setup. I am not super worried about scale down because (1) we should protect the hub and proxy pods with a PDB so they don't get moved ...

If you protect the single point failure pods with a PDB that makes them block all scale down, then you wont reliably scale down as a compromise. And, since the pods can schedule on all nodes without using the user-scheduler, they will spread out, which means that you are likely to end up with many nodes not scaling down in the end which will either cost money and require manual interventions.

>  and (2) the mybinder.org cluster never scales down below 3-4 nodes. So while a hub pod might block scaling down of a node, it isn't a big deal because the minimum size of the cluster is larger than the number of hub pods. (This is of course not true for all binderhub deployments but here we only care about mybinder.org)

I don't understand what you mean. My conclusion is that if you peak at 6 nodes for example, and we made a mybinder.org-deploy update that made a pod that we have started protecting to block scale down from a cluster autoscaler, then we will end up staying at up to 6 nodes. Due to that, I think there is a big issue.

---

In general, I find ourselves not really understanding each other so well. Perhaps we could have a call about this topic?

PR #1732, opened 2020-12-04
Run tests from the top level tests directory only

betatim:
The tests in the federation redirector sub-directory don't run in our CI
environment and are not meant to run there either. This selects only the
tests in the top level `tests/` directory.

In #1641 I added a test in the federation redirector sub directory. Hadn't realised that pytest would search all directories in the repo to look for tests.


consideRatio:
GitHub stuck for me, but go for merge whenever it works!

image

PR #1740, opened 2020-12-06
Various updates to the GitHub workflows

consideRatio:
A set of fixes to the workflow files following Helm2 is no longer part of our CI/CD system.

Edit: its a general refactoring.


.github/workflows/cd.yml
```diff
@@ -7,19 +7,17 @@
...
 
-# Only trigger the workflow when pushing to master or a label is applied to a
-# Pull Request
+# Only trigger the workflow when pushing to master
 on:
   push:
     branches:
       - master
-  pull_request:
-    types: [labeled]
```
betatim, line 17: Is this related to the helm 2 -> 3 transition? Why was it here/now being removed :) ?

.github/workflows/cd.yml
```diff
@@ -29,23 +27,16 @@ env:
...
+
     strategy:
       fail-fast: false  # Do not cancel all jobs if one fails
       matrix:
         include:
           - federation_member: staging
-            binder_url: https://gke2.staging.mybinder.org
-            hub_url: https://hub.gke2.staging.mybinder.org
+            binder_url: https://staging.mybinder.org
+            hub_url: https://hub.staging.mybinder.org
```
betatim, line 38: Please don't change this. Unrelated to going from helm 2 -> 3.

https://staging.mybinder.org is the whole "staging" federation and https://gke2.staging.mybinder.org is one member of the staging federation. I assume this step wants to talk to an individual member

.github/workflows/cd.yml
consideRatio: It was @sgibson91 attempting for a "test-staging" label trigger of PRs that didnt work. But, i realize now, that it can work for prs on branches in the repo though so probably shouldnt be removed but just documented. I tried it out by finding it only to conclude it ran into issues accessing github secrets from my fork-based label-triggered PR

.github/workflows/cd.yml
```diff
@@ -104,18 +95,9 @@ jobs:
...
 
-      - name: "Stage 2: Setup helm"
-        working-directory: mybinder
+      - name: "Stage 3: Run chartpress to build/push images and updates values.yaml"
         run: |
-          helm repo add stable https://charts.helm.sh/stable
-          helm repo add jupyterhub https://jupyterhub.github.io/helm-chart
-          helm repo add jetstack https://charts.jetstack.io
-          helm repo update
-          helm dep up
```
betatim, line 114: Do we not need to explicitly add repos any more with helm 3? We still run `helm dep up` a few lines earlier, so I think helm is now smart enough to work out which repos need adding from the chart itself?

.github/workflows/cd.yml
```diff
@@ -104,18 +95,9 @@ jobs:
...
-
-      - name: "Stage 3: Build images (if required)"
-        run: |
-          chartpress --push ${{ matrix.chartpress_args }}
```
betatim, line 118: Why are we removing the ability to set extra chartpress args?

.github/workflows/cd.yml
```diff
@@ -212,46 +187,19 @@ jobs:
...
 
-      - name: "Stage 1: Install helm ${{ env.HELM_VERSION }}"
-        if: matrix.helm_version == env.HELM_VERSION
+      - name: "Stage 1: Install and setup helm"
         run: |
-          curl -L https://storage.googleapis.com/kubernetes-helm/helm-${{ env.HELM_VERSION }}-linux-amd64.tar.gz | tar --directory ${HOME} --extract --gzip --file -
-          sudo mv ${HOME}/linux-amd64/helm /usr/local/bin/helm
-
-      - name: "Stage 1: Install helm ${{ matrix.helm_version }}"
-        if: matrix.helm_version != env.HELM_VERSION
```
betatim, line 222: Does this remove the ability to use a different version of helm for different jobs? It would be good to keep this ability for future upgrades so that we can upgrade staging first, then one/some of the federation members and then all of them. Instead of forcing our hand and having to upgrade all of them in one go.

.github/workflows/lint-validate.yml
```diff
@@ -7,22 +7,22 @@ on:
       - "**/*.yml"
 
 env:
-  HELM_VERSION: "v2.16.10"
+  HELM_VERSION: "v3.4.1"
 
 jobs:
   lint-config:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
betatim, line 14: Is this change needed/how is it related to cleaning up from the helm transition?

It is weird that `ubuntu-latest` corresponds to 18.04 and not 20.04. I think it is fine to move to 20.04 but surprised to see it in this PR so asking.

.github/workflows/lint-validate.yml
```diff
@@ -80,25 +80,20 @@ jobs:
...
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7
+          python-version: 3.8
 
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install "chartpress<1.0.0"
+          python -m pip install "chartpress==1.0.*"
```
betatim, line 91: In other parts we asked for a version `>=1.0.2` (I think), why not here?

As far as I remember there were a few bugs in the first few 1.x chartpress versions that we are trying to excclude

.github/workflows/lint-validate.yml
```diff
@@ -109,7 +104,7 @@ jobs:
 
       - name: Run chartpress
         run: |
-          chartpress
+          chartpress --skip-build
```
betatim, line 107: Why add this flag?

I'm not familiar with all the chartpress argument but `--skip-build` sounds like it will skip building images which isn't something we should be doing in the tests unless somewhere else in the tests (ideally in one that runs before merging a PR) we check that the images build.

.github/workflows/cd.yml
sgibson91: Yes, I haven't figured out a good way to make secrets available to forks yet - but this triggers works fine for PRs coming from branches of the same repo

.github/workflows/cd.yml
```diff
@@ -85,10 +76,10 @@ jobs:
         with:
           version: ${{ env.KUBECTL_VERSION }}
 
-      - name: "Stage 1: Install helm ${{ env.HELM_VERSION }}"
+      - name: "Stage 1: Install and setup helm"
```
sgibson91, line 79: I personally think it's nice to leave the echo of the version in the name of the step as this makes it much quicker for humans to verify that the correct version is being parsed in the logs. I.e. if it's parsed correctly in the step name, it's likely correctly parsed in the action itself.

.github/workflows/cd.yml
sgibson91: Yes, this would remove that functionality.

.github/workflows/cd.yml
consideRatio: requirements.yaml describe everything needed for the helm upgrade command

.github/workflows/cd.yml
consideRatio: It was unused and made the code comprehension go down by remaining and is is trivial to add back.

.github/workflows/cd.yml
consideRatio: A federation member can have its env HELM_VERSION updated and it will work again if needed. I believe we only need to do this for a Helm3->Helm4 transition in the future because helm minor/patch differences have been unproblematic in the past.

.github/workflows/cd.yml
consideRatio: Im adding this back with some inline comments.

.github/workflows/cd.yml
consideRatio: They are a remnant of a transition period i believe. We only have one staging cluster now but had two during a transition i think. gke2.staging and staging are the same according to a nslookup through a CNAME. @minrk can perhaps confirm this and suggest what makes sense at this point.

.github/workflows/lint-validate.yml
consideRatio: We can add that, but its a ci system without cache, so it will never be needed.

.github/workflows/lint-validate.yml
consideRatio: Because it doesnt help linting to build the docker images. This step will makr values.yaml be updated with image tags etc which is required for template rendering still though

.github/workflows/lint-validate.yml
consideRatio: I should have excluded the helm2 mention at all in this PR title, it is a general rework. It is not related to the helm2 removal and I've updated the title.

.github/workflows/cd.yml
minrk: Having an "extra command args" variable that can be empty is pretty a standard pattern in shell scripting. Maybe it would be clearer to have it defined and empty in the top-level of the matrix? We've demonstrated that this is useful to have, since we needed it during the federation transition. I don't really see the benefit of making it more work to do similar things in the future by deleting it, but it is true that we don't need it at the moment and probably won't very often.

.github/workflows/cd.yml
consideRatio: The need to do helm repo add, is if you do `helm install reponame/chartname`, which we don't anywhere, but we do `helm install ./local-folder` with a requirements.yaml file.

.github/workflows/cd.yml
consideRatio: I've added back the version!

.github/workflows/cd.yml
sgibson91: I think some `helm repo add`s and a `helm dep up` might get run in `deploy.py` too. This might be a good opportunity to reduce unnecessary duplication.

.github/workflows/cd.yml
consideRatio: I reverted this change as its only a partial change anyhow as both gke2.staging and staging remain valid domain names with valid TLS certificates etc in our deployment logic.

minrk: nslookup results aren't really meaningful here - the same is true of `gke.mybinder.org` and `mybinder.org` (and archive.analytics.mybinder.org, etc.), but ingresses differentiate them. These should point to a `binderhub` ingress, not `federation-redirect`, but we can use `gke.staging` instead of `gke2`.

```bash
$ kubectl get ingress
...
binderhub                   gke.staging.mybinder.org,gke2.staging.mybinder.org                                   35.222.35.25   80, 443   82d
federation-redirect         staging.mybinder.org                                                                 35.222.35.25   80, 443   82d
...
```

.github/workflows/lint-validate.yml
minrk: I think being explicit about versions instead of using '-latest' is a good idea in general.

.github/workflows/cd.yml
consideRatio: It was used in the past [here](https://github.com/jupyterhub/mybinder.org-deploy/pull/1604/files#diff-ea3ea8c9932adc7ba8161ceda844fedd43b006848ef1140c050cbd7ea0788a18R40-R52) to set an image prefix. I'll add it back with a comment on how it can be used.

consideRatio: It is now added back with a comment also explaining some details about gke2.staging vs staging

.github/workflows/cd.yml
consideRatio: I feel it was a bit messy to keep this functionality while refactoring away helm2 logic in this workflow without knowing how the deploy.py script will be refactored.

After deploy.py is refactored, I can add it back as a preparation for a helm3->helm4 transition.

consideRatio: I've added it back now for future use!

consideRatio:
I have addressed the review feedback thoroughly at this point I think, with the last piece being the support of `helm_version` matrix arguments.

@betatim I know you don't like self merges so I want to explain why I'm opting to self-merge this PR at this point. It is for me to be able to baby-sit this through now that I have time to do so. By doing this now, I can also prepare and validate I got #1753 right so it can be ready for review without failing tests.

I'd appreciate if you let me know if this feels bad in any way!

PR #1743, opened 2020-12-07
Cleanup deploy.py Helm logic

sgibson91:
**Changes in to `deploy.py` in this PR include:**

- Removal of logic to check if version 2 or 3 of Helm is available
- Removal of the `setup_helm` function. It is not required now that the whole federation is running Helm3.
- Add an assertion that Helm3 is available.

**Question:** Do we want to retain the `--local` flag? With the helm setup removed, all it skips is authenticating against the cluster.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -286,7 +286,7 @@ def main():
     argparser.add_argument(
         '--local',
         action='store_true',
-        help="If the script is running locally, skip auth and helm steps."
+        help="If the script is running locally, skip auth step"
     )
 
     args = argparser.parse_args()
--- a/deploy.py
+++ b/deploy.py
@@ -127,7 +127,16 @@ def setup_auth_gcloud(release, cluster=None):
 def check_helm_major_version():
     """Gets the major version of helm and checks it's what we're expecting
     """
-    client_helm_cmd = ["helm", "version", "--short"]
+    try:
+        client_helm_cmd = ["helm", "version", "--short"]
+    except subprocess.CalledProcessError as err:
+        raise Exception(
+            "\n".join[
+                "Failed to get correct Helm version. Expected version >=3.*.*.",
+                err,
+            ]
+        )
+
     client_version = (
         subprocess.check_output(client_helm_cmd).decode("utf-8")
     )
--- a/deploy.py
+++ b/deploy.py
@@ -20,12 +20,6 @@ else:
 HERE = os.path.dirname(__file__)
 ABSOLUTE_HERE = os.path.dirname(os.path.realpath(__file__))
 
-# Get helm version environment variable
-HELM_VERSION = os.getenv("HELM_VERSION", None)
-if HELM_VERSION is None:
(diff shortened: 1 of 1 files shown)


deploy.py
```diff
@@ -130,13 +124,10 @@ def setup_auth_gcloud(release, cluster=None):
...
-def get_helm_major_version():
-    """Get the major version of helm
-
-    Returns:
-        helm_major_version (str): Either "v2" or "v3"
+def check_helm_major_version():
+    """Gets the major version of helm and checks it's what we're expecting
     """
-    client_helm_cmd = ["helm", "version", "-c", "--short"]
+    client_helm_cmd = ["helm", "version", "--short"]
```
minrk, line 130: FWIW, if helm 3 is expected and helm 2 is present, the result is:

```
Error: could not find tiller
  File "./deploy.py", line 337, in <module>
    main()
  File "./deploy.py", line 294, in main
    check_helm_major_version()
  File "./deploy.py", line 132, in check_helm_major_version
    subprocess.check_output(client_helm_cmd).decode("utf-8")
  File "/Users/minrk/conda/lib/python3.8/subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/Users/minrk/conda/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['helm', 'version', '--short']' returned non-zero exit status 1.
```

maybe catch CalledProcessError and show our own  "Failed to get helm version, expected ..." error?

minrk (approved):
> Do we want to retain the --local flag?

I still think this is useful because the auth step can make a mess when run locally. It confuses my kubectl/gcloud/az configuration and I prefer to be able to do this step myself. It also takes a substantial amount of time, which is frustrating during those times I have to do manual deploys due to some weird failure. The downside, however, is that doing it myself does allow me to point a deployment of one configuration to the wrong cluster!

deploy.py
sgibson91: Such as 7bfc288 ?

sgibson91:
> I still think this is useful because the auth step can make a mess when run locally. It confuses my kubectl/gcloud/az configuration and I prefer to be able to do this step myself. It also takes a substantial amount of time, which is frustrating during those times I have to do manual deploys due to some weird failure. The downside, however, is that doing it myself does allow me to point a deployment of one configuration to the wrong cluster!

I could expand the user input/feedback a little such that...?

```
if args.local:
    response = input("Have you correctly authenticated against {cluster_name} cluster? [yes/no] ")
    ...
else:
    # Do the CI check we currently run
    ...
```

deploy.py
minrk: Yup, perfect!

minrk:
I think as it is now is probably best. While that issue is theoretical, local deploys are pretty rare and it's worked well as it is, at least in my experience. Another PR to add confirmation to local deploys might still be a good idea!

PR #1744, opened 2020-12-08
Use matchLabels for network policies

minrk:
avoid matchExpressions in network policies

matchExpressions don't work as advertised on AKS ([cross-ref](https://github.com/jupyterhub/mybinder.org-deploy/issues/1468#issuecomment-738797895)).

Instead, produce the same intended result by rendering a separate policy with `matchLabels` for each label (there are only two).

restores default singleuser network policy enforcement on turing, reverting #1715

**EDIT:** my pr tool made a mess!

### DIFF
--- a/mybinder/templates/netpol.yaml
+++ b/mybinder/templates/netpol.yaml
@@ -1,24 +1,29 @@
 {{- if .Values.binderhub.networkPolicy.enabled -}}
+{{- $root := . }}
+{{- range $component := tuple "singleuser-server" "dind" -}}
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: binder-users
+  name: binder-users-{{ $component }}
   labels:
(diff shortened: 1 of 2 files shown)


mybinder/templates/netpol.yaml
```diff
@@ -1,24 +1,29 @@
 {{- if .Values.binderhub.networkPolicy.enabled -}}
+{{- $root := . }}
```
consideRatio, line 2: I'm not sure when it changed, but at least in Helm3, you can avoid needing to stash away the root context like this and instead just rely on `$` pointing to it.

So for example...

```diff
-{{- $root := . }}
 # within a range loop context where we can no longer use .
-release: {{ $root.Release.Name }}
+release: {{ $.Release.Name }}
```

minrk:
closing due to turing cluster redeploy means this workaround is not necessary

PR #1745, opened 2020-12-08
restore network policy on turing

minrk:
reverts #1715 

with singleuser policy enabled, the result is more restrictive than it should be. With it disabled, it is *less* restrictive than it should be.

Plus, this will make it easier to verify that the next deploy of the turing cluster fixes the issue.

#1744 includes a workaround for AKS, so let's not merge that one if we are still trying to figure out the root of #1468

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -70,10 +70,6 @@ binderhub:
         enabled: true
         replicas: 5
 
-    singleuser:
-      networkPolicy:
-        enabled: false
-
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):
Thanks! I'll reenable this, spin up a little cluster with a different netpol controller and test if that solves the issue. If it does, I'll tear down the big Turing cluster and redeploy.

Issue #1717, opened 2020-12-02
Please bump resources for https://gitlab.kwant-project.org/kwant/minicourse.git


akhmerov:
### 1. Who you are

Developer team of an open source quantum transport package https://kwant-project.org/

Christoph Groth (CEA Grenoble)
Xavier Waintal (CEA Grenoble)
Michael Wimmer (TU Delft)
Anton Akhmerov (TU Delft)

### 2. A few details about the event/course/context

A 4 hour hands-on tutorial on quantum transport using binder for exercises, full description here: https://virtualscienceforum.org/#/quantum-transport-workshop

### 3. If money is changing hands (attendees pay a fee, speaker is getting paid, free, etc)

No money is changing hands.

### 4. A link to the repository you are using

http://gitlab.kwant-project.org/kwant/minicourse.git (mostly empty so far)

### 5. What type of event it is:
- [ ] talk
- [x] workshop
- [ ] lecture course
- [ ] training session

### 6. How many people you expect to attend the event

500–600

We have 400 registrations 2.5 weeks ahead of time. This number will still grow, but I expecft only around 60% of the registrants to show up. The hard limit of the zoom session is 500, and there may be people joining via live stream.

### 7. The exact times and dates of the event(s) (with timezone information)

Start: 2020-12-17 14:00:00 UTC
End: 2020-12-17 18:30 UTC

### 8. How do you plan to give mybinder.org a shout out at the event

- Using the acknowledgment is on the event page
- Verbally, during the opening of the workshop (we'll also need to explain how to use binder).

akhmerov:
An additional question: is it possible to increase the container/kernel culling times?

betatim:
We can't easily change culling times for just one repository. A reason to keep it short is that people will get "disconnected" from the session for various reasons like accidentally closing the tab, starting a second one, etc. Right now they go away quite quickly instead of blocking resources/quota.

A pattern that seems to have worked well for others is to announce the link people should click to start their binder close to when they will use it. Not at the start of the call, followed by 20min of introductory material (this means most binders have timed out again). You can also tell people to run `time.sleep(10 * 60)` in a notebook if they plan to go away for a bit.

What is the proposed quota? In my experience if a video call has N participants then about 0.666 * N binder instances will be needed. With the zoom call limited to 500 I think a quota of around 350 would work.

This event will lead to close to double the typical number of pods running. This means we are pushing into new territory, which is exciting but also means we could encounter dragons (problems).

[...]

PR #1750, opened 2020-12-09
bump resources for kwant/minicourse

akhmerov:
refx: #1717

Does this look correct? I've pretty much copied stuff from comparing to what else is around.


betatim:
Looks good to me. I'll merge it already now so it doesn't get forgotten.

PR #1753, opened 2020-12-12
[MRG] Merge requirements.yaml into Chart.yaml as supported in Helm3

consideRatio:
Having them separated is a legacy of Helm2. By merging them, we keep updated with Helm3 practices.

This will make the Helm chart require deployment by Helm3, which should be fine at this point as all our federated members use Helm 3 already.


mybinder/Chart.yaml
```diff
@@ -1,5 +1,48 @@
...
+  - name: grafana
+    version: 5.0.26
+    repository: https://grafana.github.io/helm-charts
+
+  # BinderHub
+  # Source code:    https://github.com/jupyterhub/binderhub/tree/master/helm-chart
+  # App changelog:  https://github.com/jupyterhub/binderhub/blob/master/CHANGES.md
+  - name: binderhub
+    version: 0.2.0-n454.h97fb8c3
+    repository: https://jupyterhub.github.io/helm-chart
```
consideRatio, line 48: There is no change in versions etc, just a migration of the requirements.yaml file to be embedded into this Chart.yaml file as Chart.yaml apiVersion: 2 wants to have it.

consideRatio:
@henchc will this PR break the henchbot, or does it work with dependencies stored in Chart.yaml ?

henchc:
Yes, I think, all that needs to change is the filename though, it looks like you're maintaining the structure?

https://github.com/henchbot/mybinder.org-upgrades/blob/master/src/mybinder-upgrades/henchbot.py#L302
https://github.com/henchbot/mybinder.org-upgrades/blob/master/src/mybinder-upgrades/henchbot.py#L158
https://github.com/henchbot/mybinder.org-upgrades/blob/master/src/mybinder-upgrades/henchbot.py#L170

Does that sound right to you? I'm happy to make the PR if that sounds right.

consideRatio:
Yes that sounds correct! The structure remains as in requirements.yaml but is moved to be contained in Chart.yaml. This is the new Helm3 structure used.

👍🥳❤️ for such update!

henchc:
Here's the PR: https://github.com/henchbot/mybinder.org-upgrades/pull/5

@consideRatio I've invited you as a collaborator to give you write access. So whoever/whenever this is merged please merge this one as well 😄

PR #1764, opened 2021-01-08
some more mining pools IP

MridulS:
found them mining on GESIS notebooks


choldgraf:
many thanks @MridulS 👍

PR #1766, opened 2021-01-09
cert-manager: request some CPU/memory to ensure some capacity

### LINKED ISSUE #1765: cert-manager instabilities
cert-manager made me run into [this failure](https://github.com/jupyterhub/mybinder.org-deploy/runs/1672618503?check_suite_focus=true) on OVH recently.

```
Error: UPGRADE FAILED: cannot patch "letsencrypt-staging" with kind ClusterIssuer: Timeout: request did not complete within requested timeout 30s && cannot patch "letsencrypt-prod" with kind ClusterIssuer: Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=30s: net/http: TLS handshake timeout
```

What's going on is that the k8s-apiserver receives an request to update a ClusterIssuer k8s resource, which cert-manager has registered interested in having a say about. cert-manager has "a registered webhook" that the k8s api-server will use before accepting the suggested edit to the ClusterIssuer. But, when the k8s api-server tries to speak with this webserver responding to the webrequest being made to it by the k8s-apiserver fail to respond, we experience this crash.

In past versions of cert-manager, it was possible to disable that webhook. Now it doesn't seem possible any more.

So, why didn't it respond? Well, probably because it was a single replica that were starved of CPU resources by being scheduled on a user node starved of resources due to a bitcoin mining attack or similarly. In this case, there was such attack going on.

## Related
- [cert-manager docs](https://cert-manager.io/v0.15-docs/concepts/webhook/)

## Suggested action plan
- We explicitly allocate CPU or similarly, that is probably enough to ensure it can respond to the infrequent web requests it will get.
- We schedule this pod on core node pool for a reduced risk of issues from mischievous users.

It seems like they have a suggested cpu/memory from their default values.yaml file (commented out), of 10m / 32Mi. I'll go ahead and apply that with 50% margin. This is an improvement from the 2m which is the absolute minimum for all pods.

The current usage can be seen here by `kubctl top pod -n cert-manager`.

```
NAME                                       CPU(cores)   MEMORY(bytes)   
cert-manager-784f598689-79vvc              4m           30Mi            
cert-manager-cainjector-67b7c65dff-nxggk   5m           26Mi            
cert-manager-webhook-7d5d8f856b-4wbvz      2m           6Mi 
```

I think the right call is to add a nodeSelector to isolate our sensitive workloads from the turbulence on the user nodes in general but I know @betatim disagrees.

consideRatio:
Closes #1765 hopefully, not setting a nodeSelector though.

### DIFF
--- a/config/cert-manager.yaml
+++ b/config/cert-manager.yaml
@@ -1,4 +1,26 @@
+# References:
+# Note that 0.15 is hardcoded in these links as it was the version currently in use.
+# - Cert-manager docs: https://cert-manager.io/v0.15-docs/concepts/webhook/
+# - Default values: https://github.com/jetstack/cert-manager/blob/release-0.15/deploy/charts/cert-manager/values.yaml
+#
 ingressShim:
   # our default ClusterIssuer is in mybinder/templates/cluster-issuer.yaml
   defaultIssuerName: letsencrypt-prod
   defaultIssuerKind: ClusterIssuer
+
+# Added to protect against being starved of CPU and failing to respond, which
+# can cause helm upgrade failures.
+resources:
+  requests:
+    cpu: 15m
+    memory: 48Mi
+webhook:
+  resources:
+    requests:
+      cpu: 15m
+      memory: 48Mi
+cainjector:
+  resources:
+  requests:
+    cpu: 15m
+    memory: 48Mi


betatim:
Why do you think this change will fix the problem?

consideRatio:
The linked issue contains the issue outlines everything.

PR #1771, opened 2021-01-10
Add port 16286 to egress whitelist

okofish:
TCP port 16286 is used by the [Wolfram Engine software](https://www.wolfram.com/engine/) for product licensing functionality, as noted [here](https://support.wolfram.com/12409). When egress on this port is blocked on mybinder.org, repos/images based on the [Wolfram Language kernel](https://github.com/WolframResearch/WolframLanguageForJupyter) can't be run in on-demand licensing mode.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -48,6 +48,7 @@ binderhub:
         - 873 # rsync
         - 1094 # xroot
         - 1095 # xroot
+        - 16286 # Wolfram Engine on-demand licensing
       cidr: 0.0.0.0/0
 
   config:


choldgraf:
Can you explain a bit further what this use-case is? Is it when people want to run a proprietary Wolfram kernel that requires a license to authenticate, but they want to run it on mybinder.org?

okofish:
Yes, that is correct - I'm assisting a user who's trying to use such a setup in a classroom setting and we ran into the port-blocking issue. For reference, their test repo is [here](https://github.com/gvarnavi/binder-wl-test).

Normally the Wolfram Engine is licensed in a node-locked manner, which pretty much doesn't work in all in clustered environments like Kubernetes where the end-user can't control which machine a job gets scheduled to. A new metered floating license mechanism was introduced recently, but it needs to talk to port 16286 on a Wolfram server in order to function.

I'm happy to answer any questions as needed.

manics:
@okofish Could you explain how the licensing process works? Presumably there needs to be some sort of authentication step, but since mybinder.org is public this can't be done in the image, so how are you managing it?

Also note this FAQ: https://mybinder.readthedocs.io/en/latest/about/about.html#can-i-push-data-from-my-binder-session-back-to-my-repository
> We cannot guarantee the security of data moving through mybinder.org, and your password or any sensitive data may be compromised. You shouldn’t do anything on mybinder.org that you wouldn’t mind sharing with the world!

okofish:
@manics A form of license key (called an "entitlement ID") does have to be embedded in the image as an environment variable or otherwise fetched in by the image's initialization code. I entirely agree that this is unfortunate from a security perspective, but it's the best solution that we have at present.

There are options provided to users to place restrictions on their license keys to limit the potential of abuse, such as setting an arbitrary expiration date and limiting the maximum number of concurrent kernel processes that can be run using a key. In the case of the user I mentioned with the classroom use case, they might set the expiration date to be the end of their course and the process limit to be the number of students in their course. A key can be manually revoked by its owner at any time.

okofish:
I should mention that another option is for a repo developer to include machinery that prompts the end-user for a license key when the image is started instead of embedding the key in the image. There isn't yet any native mechanism in the Wolfram Language kernel for this, but it's something I'm looking into. It could be done via a shell script included in a repo and that an end-user launches from the Jupyter terminal.

manics:
Have you thought about running your own JupyterHub (if there are only a small number of Docker images) or BinderHub (if you need the flexibility to create new images) instead? You can configure authentication to have full control over who can access it which would avoid any security issues with exposing license keys.

gvarnavi:
@okofish Thanks for opening this.
@choldgraf A few more specifics on the particular use-case, although of-course this will be useful for all binder-wolfram language integration. I'll use a form similar to the resources bump request form (although default resources should be fine):

#### 1. Who you are
My name is George Varnavides, and I am a graduate researcher at the Massachusetts Institute of Technology.
#### 2. A few details about the event/course/context
We're teaching a [generative art workshop](http://student.mit.edu/searchiap/iap-9289af8f5f00a363015f0d7edae8003a.html), during MIT's Independent Activities Period [(IAP)](http://student.mit.edu/searchiap/iap-9289af8f5f00a363015f0d7edae8003a.html). The content is distributed as python, julia, and wolfram language notebooks. We were hoping to share a single executable environment for all three languages.  
Re: licensing issued raised above, while unfortunate security-wise the model proposed by @okofish serves our needs (the entitlement id is set to expire soon after the workshop, and I'm monitoring its usage for signs of abuse)
#### 3. If money is changing hands (attendees pay a fee, speaker is getting paid, free, etc)
Free event. Typically limited to students in Boston but anyone can join this year (virtual instruction)
#### 4. A link to the repository you are using
Workshop repository with content: https://github.com/gvarnavi/generative-art-iap
Test repository with wolfram-language integration: https://github.com/gvarnavi/binder-wl-test
#### 5. What type of event it is:
- [ ] talk
- [x] workshop
- [ ] lecture course
- [ ] training session
#### 6. How many people you expect to attend the event
120 students are currently signed up. The workshop will be mostly synchronous but we will provide lecture recordings, so I suspect the actual instances to be under 100 at any given time (due to time-difference etc) 
#### 7. The exact times and dates of the event(s) (with timezone information)
01/19 - 01/22: 10am-1pm ET
#### 8. How do you plan to give mybinder.org a shout out at the event
Already on repository's README (happy to beef it up), and will highlight during first session.

betatim:
We tend to be very black and white when it comes to the "do not enter or store in a binder information you wouldn't tweet to the world". This is mostly because we need a message that is easy to understand. A more nuanced "it depends ..." is hard for users to evaluate.

We also treat all repositories the same in terms of resources allocated to them and the like.

However, for resources we allow people to ask for a time limited exception. Via the form you found and posted in the previous comment.

What do people think of treating this as a request for "extra resources for a limited time"? Like we do when increasing the number of instances?

How long would you need the port to be open for?

---

Personal though: making the life of those people who are remote teaching and are being remote taught easier in the current climate is something that is part of our mission. Especially if we put a time limit on it.

However I am also a big proponent of policies that work for everyone and do not require exceptions/not making exceptions. As this usually leads to "all the exceptions all the time" :-/

gvarnavi:
Looking forward to hearing what people think about the "extra resources for a limited time" idea. Thanks again for considering, it's really appreciated!

> How long would you need the port to be open for?

The four-day workshop will be held from Jan 19th - 22nd. Ideally, we'd have it open a bit before then, to make sure everything works smoothly, and conceivably a couple of days after, for people watching the recorded content asynchronously to play around with the binder. Would it be possible to ask for earliest convenience until Sunday Jan 24th (or even Jan 31st)?

betatim:
I'm (obviously) in favour. @manics what do you think?

choldgraf:
This sounds good to me as well, especially if it is time-limited.

manics:
Sounds good to me! Let's see how it goes, if it all goes to plan we could perhaps formalise this as an option for workshops in our docs?

betatim:
Thanks everyone for the positive attitudes :)

The change has been deployed, should be live in a few minutes. Let us know how it goes and if you hit any weirdness.

gvarnavi:
Great, thanks all for the flexible and prompt response!
Seems to work properly:
```Mathematica
$Version
$MachineName
```
>12.2.0 for Linux x86 (64-bit) (December 12, 2020)
jupyter-gvarnavi-2dbinder-2dwl-2dtest-2dl7gdi2pz

I'll let you know how the workshop goes and any binder-specific feedback!

choldgraf:
thanks @gvarni - I hope that it goes well for you!

okofish:
Thanks all!

I'm also going to look into options to prompt end users for licensing info after a session is started through some form of UI - this may prove a more sustainable solution for the future.

Issue #1663, opened 2020-10-21
Use credentials when pulling from dockerhub


betatim:
Dockerhub will start rate limiting pulls from its registry. The rate limit will be applied based on the IP of the docker client. So all pulls from one cluster will be under the same "account".

There are a few scenarios in which this will touch mybinder.org.

1. repos that use a custom `Dockerfile` with a base image that we don't already have
1. repos that use the github action to pre-build images that we then pull (once) on launch

In addition some of the clusters use dockerhub as their internal registry. This means every launch (with the image not already present on the node) will need to pull, contributing to the rate limit.

These are a few places I think we should go and investigate:
1. configure k8s so that it uses credentials when pulling from docker hub
	- this is probably relevant https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
1. configure repo2docker to use credentials when pulling from docker hub

betatim:
Triggered by https://discourse.jupyter.org/t/dockerhub-pull-limits-mybinder-org/6381/4

minrk:
Might also be useful to look at [image pull secrets for a service account](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account) which should have the effect of a 'default' imagePullSecret without having to specify it for every pod separately.

arnim:
@MridulS we now have the Pro account and use the corresponding credentials for pulling at GESIS - is this correct?

MridulS:
Yes, and as we don't use DIND to build/pull, we just use authenticated docker (`$ docker login`) on our host server . cc @bitnik

betatim:
Looks like the OVH cluster is out of action because it is hitting the docker hub rate limit. I will take a look at adding a image pull secret by hand and if that works figure out how/where to add it in the BinderHub chart.

Right now thinking it should be separate from the "docker registry" credentials we have because you might have a private registry and also want to configure docker hub credentials for pulling (base) images from there. Something I don't know yet is how to ask helm to patch the default service account. Or if it is even sensible to try and do that.

PR #1784, opened 2021-01-19
reinstate ovh at low weight

minrk:
- rendezvous load-balancer now implements weighting (#1641)
- dockerhub credentials have been added (#1711) which was the reason it was removed (#1663) but ovh was not restored to the federation


betatim:
Excited to see the weighted rendezvous in action

Issue #1786, opened 2021-01-20
OVH default imagePullPolicy is Always


minrk:
When we create resources without an imagePullPolicy, the default policy [should be IfNotPresent](https://kubernetes.io/docs/concepts/configuration/overview/#container-images) if the tag is

/cc @mael-le-gal how can we fix the default pull policy on OVH? Do you know where the source of the bug is? Can we fix it?

minrk:
Turns out this is a deeper issue and #1787 does not fix it. `imagePullPolicy` does not appear to be respected at any level, as even setting `imagePullPolicy: IfNotPresent` in the hub deployment results in creating the hub pod with `imagePullPolicy: Always`. This means we can't bring OVH back until this is fixed at the cluster-level. This needs support from OVH folks.

consideRatio:
Hmmm, I guessed there was a mutating webhook setup doing this, but there wasn't any except cert-managers stuff also in the other clusters (`kubectl get MutatingWebhookConfiguration`).

consideRatio:
I googled on the docker pull rule and found...

> The rules are: Anonymous users can pull 100 images in six hours. Authenticated users can pull 200 images in six hours. Paid users are not limited.

Does it mean that if we have multiple IPs, it will be perceived as different anonymous users, and therefore we would get 100 images in six hours per machine, or, are we currently pulling docker images as the same authenticated user? Then I think we would get less capacity.

minrk:
There is no mutating webhook.

We do have a docker credential for single-user images, at least. I don't know if it's being used for both init contianers and regular containers. I assume it's both. I'm not sure if each VM would have a different egress IP or not. All I know is that we were hitting the rate limit with very low traffic after turning OVH back on yesterday.

consideRatio:
> I'm not sure if each VM would have a different egress IP or not.

On GKE, you have private and public k8s clusters, where public versions get an ip per node for egress traffic, while private will need to have a Cloud NAT setup. With a NAT to provide public IP(s) one can make it become one or more different IPs. But, this is GCP specific details.

betatim:
I think we should wait and see what Mael has to say. It seems very likely that this behaviour is something that can be configured at the level of the cluster. So I wouldn't investigate workarounds until we know that it can't be changed.

mael-le-gal:
Indeed I reproduced the problem easily.
[...]

PR #1787, opened 2021-01-20
override imagePullPolicy on ovh

minrk:
OVH cluster has the wrong default imagePullPolicy (#1786), so we must specify IfNotPresent a bunch to avoid rate limiting

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -291,6 +291,7 @@ binderhub:
       initContainers:
         - name: tc-init
           image: minrk/tc-init:0.0.4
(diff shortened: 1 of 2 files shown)


mybinder/values.yaml
```diff
@@ -291,6 +291,7 @@ binderhub:
       initContainers:
         - name: tc-init
           image: minrk/tc-init:0.0.4
+          imagePullPolicy: IfNotPresent
```
minrk, line 294: This one and the singleuser.image.pullPolicy are the main ones that matter, since singleuser pods start a lot, while other pods start rarely.

PR #1798, opened 2021-01-27
set maintenance window to weekends

minrk:
to avoid node auto-upgrades during high traffic periods

from late Friday night UTC-10 to early Monday morning UTC+11

also bump kube version, which has already been auto-upgraded

Already deployed to staging to verify, will deploy to prod after merge


sgibson91 (approved):
LGTM! 🎉

PR #1801, opened 2021-01-29
Increase resources for LTSQueens/Let-s-Talk-Astrophysics

LTSQueens:
Requesting increase of resources for LTSQueens/Let-s-Talk-Astrophysics, to bump to 200 users.


choldgraf:
🎉

PR #1803, opened 2021-02-01
Reinstate Turing into the federation

sgibson91:
This PR reverts #1800 and reinstates the Turing

After local running of deploy.py, Turing can't pull the minesweeper pod images, I wonder if this will be resolved when run through CI which runs chartpress?


minrk:
>  I wonder if this will be resolved when run through CI which runs chartpress?

Yes, it should. minesweeper images are public on docker hub.

Local deploy should work after:

    chartpress --skip-build

minrk (approved):

sgibson91:
Great, thanks!

PR #1805, opened 2021-02-01
build tc-init, proxy-patches images with chartpress

minrk:
These two images came from my personal repos during a time of more experimentation in the binder deployment. We never got around to moving them to a more central location. Now that we have chartpress to build images for all federation members, we can add these to our existing pipeline to make updates easier, centralize control, etc.

Additionally, this should make it easier for these images to come from somewhere else, which may be required for OVH (#1786).

The images are imported via `git-subtree` and updated with more recent dependencies

- tc-init: used as an init container to enforce egress bandwidth. Kubernetes *claims* to have this as a feature applied by annotations, but testing suggested it is ~never done in practice. I haven't done tests in a while, so this may be something we can remove. Source repo: https://github.com/minrk/tc-init

- proxy-patches: used to add routes to the proxy for nicer 404 handling of culled containers  https://github.com/minrk/proxy-help


images/tc-init/README.md
```diff
@@ -0,0 +1,46 @@
...
+e.g `10mbit` for ten megabits per second,
+or `5mbps` for five mega*bytes* per second.
+
+For example:
+
+```yaml
+spec:
+  initContainers:
+  - name: tc-init
+    image: minrk/tc-init
```
consideRatio, line 38:
```suggestion
    image: jupyterhub/mybinder.org-tc-init
```

Given that we build images with chartpress, we publish them as well in jupyterhub I figure.

images/tc-init/example.yml
```diff
@@ -0,0 +1,22 @@
+# Copy of pod.yaml without file extension for test
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-pod
+spec:
+  initContainers:
+  - name: tc-init
+    image: minrk/tc-init
```
consideRatio, line 9:
```suggestion
    image: jupyterhub/mybinder.org-tc-init
```

consideRatio:
> * Kubernetes _claims_ to have this as a feature applied by annotations, but testing suggested it is ~never done in practice.

It seems to be [documented as an experimental feature](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-traffic-shaping). I presume it isn't enabled by default and can vary depending on CNI.

consideRatio:
@minrk can you add a README for proxy-patches also where you provide a sentence or two about its purpose and function?

- [ ] Readme added

minrk:
Those repos were created specifically for use here, because we had no other way to build images for this repo until a couple weeks ago. There's no reason to keep them in sync or upstream anything. The original repos will likely never be updated.

minrk:
@consideRatio good call. Done!

consideRatio (approved):
Thanks for documenting this @minrk, this LGMT!

I'll let you self merge this so you can monitor the deployment.

minrk:
Thanks! Merging. This *shouldn't* change anything other than where the images come from, but since the bases are bumped, there's always a chance.

PR #1814, opened 2021-02-04
kickout GESIS, the network is under maintenance

MridulS:


sgibson91:
May also be worth removing GESIS from the [health check](https://github.com/jupyterhub/mybinder.org-deploy/blob/58dc7927d4802516ae01f6152e02559232e185ab/mybinder/values.yaml#L503) too since we know it's unreachable :)

MridulS:
The health check link takes me to the line I changed in the PR itself. I think for gesis this is the only thing that needs to be changed when its down.

sgibson91:
Oh yes, sorry - this is me trying to be helpful whilst I'm supposed to be in a meeting 😅

PR #1818, opened 2021-02-05
disable privilege escalation

minrk:
disallows sudo, etc.

*builds* can create images with sudo access, etc. but we don't need (and shouldn't) allow containers to do this.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -218,6 +218,8 @@ binderhub:
       extraConfig:
         neverRestart: |
           c.KubeSpawner.extra_pod_config.update({'restart_policy': 'Never'})
+        noPrivilegeEscalation: |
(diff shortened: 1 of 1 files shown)


minrk:
Looks like we need kubespawner 0.15 for https://github.com/jupyterhub/kubespawner/pull/450 which means updating binderhub, which means:

- we need jupyterhub chart 0.11 ([ref](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1946/commits/56934713fecf22503787a4a9dae15ebed1f645f7))
- which means we need a binderhub bump after: https://github.com/jupyterhub/binderhub/pull/1258

consideRatio (approved):
LGTM, can be merged now but to reduce complexity of a potential rollback of z2jh 0.11 I figure we wait until we verify it lands correctly.

PR #1815, opened 2021-02-04
Document which files should be edited, and where, to remove a cluster from the federation

sgibson91:
This is a super quick PR to add guidance on how to remove a cluster from the federation (e.g. for maintenance). Updates, additions, rewrites are welcome!

PR #1825, opened 2021-02-08
let ovh back into the federation at low capacity

minrk:
start small, since it has a new local registry which will populate slowly:

- weight 10 (out of 300 + 10)
- pod quota: 10

following docs in #1815 (thanks @sgibson91)

cf #1786

### DIFF
--- a/config/ovh.yaml
+++ b/config/ovh.yaml
@@ -3,7 +3,7 @@ projectName: ovh
 binderhub:
   config:
     BinderHub:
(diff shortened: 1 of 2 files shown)


minrk:
Since it's getting toward the end of my day, I plan to merge this first thing tomorrow (Tuesday) AM/Europe. But if anyone else wants to merge and watch OVH for issues in the meantime, feel free!

PR #1829, opened 2021-02-09
collect datacenter ips and block traffic from AWS, GCP, Azure

minrk:
- scripts/datacenters.py collects ips from cloud providers,
  using public registries of data. This is done on every deploy and PR check.
- cidrs are collapsed to a minimal set using [ipaddress.collapse_addresses](https://docs.python.org/3/library/ipaddress.html#ipaddress.collapse_addresses).
  This reduces the total set from ~40k to < 3k since cloud providers (especially Azure) report lots of overlapping ip blocks
  with more specific information than we need.
- cidrs for each datacenter are dumped to `BinderHub.ban_networks` config in `config/common/datacenter-$name.yaml`
- Updated a few places so that any yaml files placed in `config/common` will be loaded
  without update elsewhere
- manual ip bans are moved to ban_networks from ingress, with a note that ingress bans can still be useful for crawlers, DoS, etc.
- <del>blocking each datacenter can be toggled with standard `datacenters.$name.enabled`</del>
- <del>only AWS is blocked at the moment, see #1828 for why we need to do more before blocking cloud providers that include our own federation members.</del>


minrk:
This now uses BinderHub.ban_networks from https://github.com/jupyterhub/binderhub/pull/1262 instead of blocking traffic at the ingress level

minrk:
With  jupyterhub/binderhub#1262 merged and deployed, I'd like to land early this next week to see if it has any adverse effects.

deploy.py
```diff
@@ -174,14 +171,18 @@ def deploy(release, name=None):
...
-        "-f",
-        os.path.join("config", release + ".yaml"),
-        "-f",
-        os.path.join("secrets", "config", "common.yaml"),
-        "-f",
-        os.path.join("secrets", "config", release + ".yaml"),
     ]
 
+    # common config files
+    config_files = glob.glob(os.path.join("config", "common", "*.yaml"))
```
manics, line 177:
```suggestion
    config_files = sorted(glob.glob(os.path.join("config", "common", "*.yaml")))
```
IIRC glob doesn't guarantee ordering. Doesn't matter at the moment, but might do if in future other configs are added to common.

manics (approved):
LGTM, one minor suggestion

scripts/datacenters.py
```diff
@@ -0,0 +1,142 @@
...
+import requests
+import yaml
+
+
+def cidrs_aws():
+    """AWS datacenters"""
+    url = "https://ip-ranges.amazonaws.com/ip-ranges.json"
+    r = requests.get(url)
+    r.raise_for_status()
+    #
```
willingc, line 25:
```suggestion
 
```

scripts/datacenters.py
```diff
@@ -0,0 +1,142 @@
...
+                if attr.lower() == "href" and value.startswith(
+                    "https://download.microsoft.com"
+                ):
+                    if value not in self.links:
+                        self.links.append(value)
+
+
+def cidrs_azure():
+    """Azure data centers"""
+    # leave it to microsoft to make this require a human confirmation dialog
```
willingc, line 61: lol

willingc (approved):
@minrk Looks good to me. Thanks for doing this.

deploy.py
```diff
@@ -148,18 +148,15 @@ def deploy(release, name=None):
...
-            "python3",
-            "secrets/ban.py",
-        ])
+
+    print(BOLD + GREEN + f"Updating network-bans for {release}" + NC, flush=True)
+    ban_command = [sys.executable, "secrets/ban.py"]
+    if release in {"turing", "ovh"}:
+        ban_command.append(release)
+
+    subprocess.check_call(ban_command)
```
consideRatio, line 159: As there was no inline comment describing why turing/ovh was singled out where release was passed before. I can't figure out if it make sense still or not, but I assume this make sense.

deploy.py
```diff
@@ -174,14 +171,18 @@ def deploy(release, name=None):
...
+    # common config files
+    config_files = sorted(glob.glob(os.path.join("config", "common", "*.yaml")))
+    config_files.append(os.path.join("secrets", "config", "common.yaml"))
+    # release-specific config files
+    for config_dir in ("config", "secrets/config"):
+        config_files.append(os.path.join(config_dir, release + ".yaml"))
+    # add config files to helm command
+    for config_file in config_files:
+        helm.extend(["-f", config_file])
+
```
consideRatio, line 185: I think you ended up reversing the order of release .yaml files and common .yaml files, and the order will make a difference here.

But, I think the latter is given higher priority, and you are adding it in a good order here. So, LGTM but could cause a change.

consideRatio (approved):
LGTM! I didn't review the datacenter.py logic thoroughly though.

deploy.py
minrk: The command is unchanged, only how it is added. Instead of constructing the full command separately or the two branches, the base command is always constructed and the extra argument is only added in the cases where it is required on L156-157. I added a note.

deploy.py
minrk: It was (in priority order, or reverse load order): secrets.config.$release > secrets.config.common > config.$release, now it is secrets.config.$release > config.$release > secrets.config.common > config.common. So the only difference is the order of "common secret config" and per-release public config.

so now all per-release config is ahead of all common config, rather than all secret config being ahead of all publicly readable config. It think this makes more sense, though in practice it does not make any difference as there is no config set in both places, nor is there ever a reason for there to be.

minrk:
Thanks for the reviews! I'll go ahead and merge and keep an eye on it.

A quick log analysis of the last 10k launch requests shows ~160/10000 launches, or 1.6% of launches would be blocked. I added [an announcement](https://discourse.jupyter.org/t/mybinder-org-blocking-launches-originating-from-most-cloud-providers/8094) for folks to report issues.

minrk:
bwomp, CI failed because our own tests can't talk to binder from github actions...Need to think of something there...

Issue #1811, opened 2021-02-02
terminate pods and/or processes based on cumulative cpu usage


minrk:
we have our idle culler, which lets us set a maximum pod lifetime, which we currently set at 6 hours. However,  6 hour pod that's mostly idle has a very different cost to the cluster from a pod that's been pegging a cpu at 100% that whole time.

In looking at minesweeper, I think we can consider looking at max aggregate cpu time as well, to terminate "hardworking" pods earlier than idle ones.

I think we have the ability to handle cpu times both at the process and pod level (I'm still figuring out how to access the pod metric, but it's available [in prometheus](https://grafana.mybinder.org/d/fZWsQmnmz/pod-activity?panelId=18&fullscreen&orgId=1)). The process metric is easy enough, and readily available in minesweeper. The pod metric is trickier, as is perhaps process *groups*.

What do folks think about automatically killing user processes after, e.g. 1 cpu-hour? Or even less? And what about pods? From the above chart, I think we would be shutting down ~30 process per day if we terminated processes that used more than 1 cpu hour.

manics:
1 hour seems fine, even a workshop making heavy use probably won't last more than 3-4 hours, most of which would be coding or listening to a speaker. It might be good to make these limits explicit though, either for people running workshops (https://mybinder.readthedocs.io/en/latest/about/user-guidelines.html), or perhaps on the mybinder front page so it's clear to everyone?

betatim:
Sounds like a smart idea. I have no feeling/intuition for how many CPU hours we'd expect 99% of "legitimate" processes to use. We could start with a large value and then keep turning it down till we get to the point where (on average) we kill "a handful" of processes a day. Then we can inspect what it is we are killing and re-decide?

If we end the "busy" processes in a pod won't it eventually shutdown itself due to inactivity? Makes me think doing the easy thing (cumulative CPU time for processes) gets us a long way along. But learning new things about how to get the metric for the whole Pod is also interesting :D -> process based would already be cool.

minrk:
[...]

PR #1833, opened 2021-02-10
minesweeper: collect total cpu per pod

minrk:
- cpu_total is the sum of cpu times per process
- collect all processes on the node, then associate them with pods via cgroups
- only inspect pods on the current node
- also perform sums across processes within a given pod
- only inspect processes found in pods by default
- attach reasons as strings for suspicious/should_terminate for nicer reprs
- temporary: log cpu totals for pods using more than a few minutes of CPU in order to compare with prometheus

Enables #1811, but does not yet set a threshold


mybinder/files/minesweeper/minesweeper.py
```diff
@@ -119,13 +132,17 @@ def get_procs(userid):
     """Get all container processes running with a given user id"""
     procs = []
     for p in psutil.process_iter(attrs=config["proc_attrs"]):
-        if not p.info["exe"]:
+        # TODO: should we filter to userid?
```
willingc, line 135: This approach makes sense. Moving to filter to userid later if needed.

willingc (approved):
LGTM @minrk 🚢

mybinder/files/minesweeper/minesweeper.py
minrk: For more background: initially, we filtered to userid because containers are run as userid. Then, I realized since we weren't limiting privileges, custom images could call setuid and launch processes as other uids. That means we were specifically not looking at processes that should have be considered suspicious for precisely the reason we weren't looking at them.

But after #1818, with privilege escalation disabled, it *should* be a safe assumption that all processes in containers are run as userid again. But I think there's no harm in looking at all the processes, especially now that we map to pods before inspecting.

minrk:
Thanks @willingc!

PR #1859, opened 2021-03-09
add horizontal pod autoscaler for binder

minrk:
allows autoscaling binder deployment according to cpu load

Disabled by default, but enable on staging, prod (prod needs it, staging for testing).

Prod has been maxing out 1 CPU for both instances much of peak traffic each weekday:

Screen Shot 2021-03-09 at 11 59 08

unit for targetCPU is percent utilization of the given cpu request, so with `resources.request.cpu: 0.5` , a targetCPU of 100 means 100% *of 0.5*, so usage of at least 0.5 CPU on average across binder pods will trigger scale-up.

This may well make sense to promote up to the binderhub chart, but I'd like to test it out here first. The main reason *not* to do it yet is that HPA specs are pretty new and there's lots of variation on how to specify metrics. Doing it here means we don't need to try to meet everyone's goals via configuration.

I also don't know with how we route requests if this is going to help - it's possible that we aren't really spreading out the CPU load and all binder instances will ~always have the same CPU load.


betatim:
> I also don't know with how we route requests if this is going to help - it's possible that we aren't really spreading out the CPU load and all binder instances will ~always have the same CPU load.

What do you mean? Requests to the binder pod should be routed at random because it is done by a `Service`. At least that is what I think happens.

A thought I had was that I don't know why binderhub uses so much CPU, so it could be that it is something that is related to total usage. Like counting active pods or some such. In that case each of the 4 pods would max out their CPU. So here scaling the number of pods won't help. But it could also be that the load scales with the number of (concurrent) requests a binderhub pod is handling in which case having more pods will help drop the load per pod. Do you have an idea?

betatim:
It looks like the HPA scaled up the number of replicas to 4 but looking at the last events it seems like it is not completely happy:
```
  Normal   SuccessfulRescale             20m                horizontal-pod-autoscaler  New size: 4; reason: cpu resource utilization (percentage of request) above target
  Warning  FailedUpdateStatus            20m                horizontal-pod-autoscaler  Operation cannot be fulfilled on horizontalpodautoscalers.autoscaling "binder": the object has been modified; please apply your changes to the latest version and try again
  Warning  FailedGetResourceMetric       18m (x3 over 18m)  horizontal-pod-autoscaler  did not receive metrics for any ready pods
  Warning  FailedComputeMetricsReplicas  18m (x3 over 18m)  horizontal-pod-autoscaler  failed to compute desired number of replicas based on listed metrics for Deployment/prod/binder: failed to get cpu utilization: did not receive metrics for any ready pods
```
The HPA object was 20m old at this moment.

betatim:
Load per pod has gone down from 100% so it looks like my previous comment about load being related to cluster occupancy vs requests is irrelevant now. Looks like the load is related to number of concurrent requests.

minrk:
> What do you mean?
> A thought I had was that I don't know why binderhub uses so much CPU, so it could be that it is something that is related to total usage.

Yeah, that's basically my hunch - that something is not actually a per-request load so increasing replicas won't significantly reduce the load, and reducing some frequency / adding some caching might help more than throwing resources at the problem. It might even increase it!

I think we need to run some `py-spy` on it and hope for meaningful output (I wouldn't bet on it, given our use of threads and async-ness).

It's possible the right combination of caching / rate limits / etc. would help more than throwing pods at the problem.

Historically, it was streaming build logs that cost the most (lots of messages to receive and reserialize). I haven't correlated pod cpu with build requests, though.

minrk:
I ran a quick py-spy, even though load is not high, and it's ~all in kube api deserialization at th moment. I really thought we had already adopted `_preload_content=True` everywhere (mentioned [here](https://github.com/jupyterhub/binderhub/pull/1192)), but we haven't! I'm sure that will make a huge difference because `launch` lists all running pods in order to check quotas - that's O(running pods) independent of binder users.

PR #1860, opened 2021-03-09
Set max CPU time to 1 hour

### LINKED ISSUE #1811: terminate pods and/or processes based on cumulative cpu usage

minrk:
user pods that use more than one cpu-hour will be immediately shutdown

closes #1811


minrk:
Worth mentioning, I think. We should probably be refining how we apply these limits over time.

PR #1862, opened 2021-03-10
disable hpa on prod

minrk:
since we can be pretty confident it won't help right now (https://github.com/jupyterhub/mybinder.org-deploy/pull/1859#issuecomment-795023498)


minrk:
Can re-enable it after deploying https://github.com/jupyterhub/binderhub/pull/1273

PR #1871, opened 2021-03-25
Remove OVH from the rotation

betatim:
Pulling the k8s-network-tools image from dockerhub is running into the rate limit

I thought we had changed our setup to fetch this and the other support images from the local docker registry?


minrk:
Yeah, we did in #1821...

Need to figure out why that didn't happen

minrk:
Ah, `network-tools` is part of the jupyterhub chart, so that's still coming from jupyterhub.

We can pull & push a copy of that one, I guess.

PR #1882, opened 2021-03-30
GESIS servers are back up, bring it back to fed

MridulS:


choldgraf:
we should get one of those friendly jupyter welcome bots to say "welcome back" every time somebody is brought back into the Binder federation 😆

PR #1892, opened 2021-04-19
limit concurrent deployment workflows

minrk:
serializes multiple outstanding deploys if PRs are merged before the last deploy finishes

to avoid multiple helm deploys racing with each other

I just learned about this checking [the changelog](https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/) looking for the upcoming workflow permissions feature, and it seems useful!

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -7,6 +7,11 @@
 
 name: Continuous Deployment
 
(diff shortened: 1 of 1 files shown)


manics (approved):
Nice find, I'm going to be stealing this for my own workflows 😀

Issue #1870, opened 2021-03-23
Turing deployment is broken


manics:
The Turing deployment currently fails, for example https://github.com/jupyterhub/mybinder.org-deploy/runs/2174637361?check_suite_focus=true
```
Starting helm upgrade for turing
Error: UPGRADE FAILED: cannot patch "jupyterhub" with kind Ingress: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post https://hub23-ingress-nginx-controller-admission.hub23.svc:443/networking/v1beta1/ingresses?timeout=10s: context deadline exceeded && cannot patch "binderhub" with kind Ingress: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post https://hub23-ingress-nginx-controller-admission.hub23.svc:443/networking/v1beta1/ingresses?timeout=10s: context deadline exceeded && cannot patch "turing-grafana" with kind Ingress: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post https://hub23-ingress-nginx-controller-admission.hub23.svc:443/networking/v1beta1/ingresses?timeout=10s: dial tcp 10.0.194.230:443: i/o timeout && cannot patch "turing-prometheus-server" with kind Ingress: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post https://hub23-ingress-nginx-controller-admission.hub23.svc:443/networking/v1beta1/ingresses?timeout=10s: context deadline exceeded && cannot patch "redirector" with kind Ingress: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post https://hub23-ingress-nginx-controller-admission.hub23.svc:443/networking/v1beta1/ingresses?timeout=10s: context deadline exceeded && cannot patch "static" with kind Ingress: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post https://hub23-ingress-nginx-controller-admission.hub23.svc:443/networking/v1beta1/ingresses?timeout=10s: context deadline exceeded
Traceback (most recent call last):
  File "./deploy.py", line 346, in <module>
    main()
  File "./deploy.py", line 342, in main
    deploy(args.release, args.name)
  File "./deploy.py", line 189, in deploy
    subprocess.check_call(helm)
[...]

PR #1895, opened 2021-04-21
Scope ingress-nginx to Namespace on Turing

sgibson91:
This PR forces the ingress-nginx sub-chart to only look for ingresses in the deployment namespace. this is to avoid collisions with the Hub23 deployment also running on the cluster, which also has ingress-nginx as a sub-chart, as detailed in https://github.com/jupyterhub/mybinder.org-deploy/issues/1870#issuecomment-820388311 (Same fix has been applied to Hub23)

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -121,7 +121,7 @@ prometheus:
 ingress-nginx:
   controller:
     service:
-      loadBalancerIP: 51.138.65.80
+      loadBalancerIP: 20.76.176.114
     scope:
       enabled: true
 
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -122,6 +122,8 @@ ingress-nginx:
   controller:
     service:
       loadBalancerIP: 51.138.65.80
+    scope:
+      enabled: true
 
 
 static:


manics (approved):
Matches the suggestion from https://github.com/jupyterhub/mybinder.org-deploy/issues/1870#issuecomment-820388311, I guess merge whenever you're on hand to watch the deployment?

Does the DNS need an update too? The following aren't resolving:
- hub.mybinder.turing.ac.uk
- binder.mybinder.turing.ac.uk
- turing.mybinder.org

sgibson91:
Ah yes, probably need to point the load balancer at the A records again, on my to-do list for tomorrow. Also this doesn't revert my previous PR to remove the Turing from the federation, so merging this this won't actually do anything to the live service. Thanks for the review!

consideRatio (approved):
This LGTM!

Separate question: do you think it would be preferable to have only one ingress controller helping out across multiple namespaces?

:+1: It would reduce the required amount of pods in the cluster
:-1:  It could become a more complicated setup to document and understand

sgibson91:
👍🏻 it would be easier for federation members like the Turing to run side-by-side deployments
👍🏻 We already do this with cert-manager

I don't have a strong opinion though, so long as it works 🙂

minrk:
For the shared cluster-wide ingress controller, what do you think is the best way to do it?

1. deploy the ingress controller itself outside the chart, like we do with cert-manager.
2. deploy the ingress controller inside the chart and say other deployments on the same cluster shouldn't deploy an ingress controller (own it, but no other changes - this should be equivalent to 1. from the perspective of *other* charts)
3. deploy it inside the chart, but make it easy to disable the whole thing for cluster maintainers to deploy their own cluster-wide ingress controller (it still must be nginx, I think, given some of our configuration)
4. deploy it inside the chart but recommend (and facilitate, if needed) in-namespace scoping for multi-tenant clusters (what this PR does)
5. Don't have a cluster-wide ingress controller by default, and set `scope.enabled = true` by default, so our ingress controllers don't bleed into other namespaces by default (move this PR's change to the shared values.yaml)

My inclination, assuming this has worled, is to do the namespaced ingress controller, either opt-in (4) or opt-out (5), rather than sharing ingress controllers. What do other folks think?

sgibson91:
DNS zones reset, so the domains should resolve in 5 mins-ish @manics 

I definitely agree that (4) or (5) are the way forward, and would probably go for the opt-in (4) method as perhaps not every federation member would want to do it this way?

minrk:
Yeah, that makes sense. In which case, we are done since this PR was already the only thing to do for (4).

4 and 5 ought to be identical for federation members dedicated to only this deployment, since all our ingresses are necessarily in the same namespace. If I've understood, the difference should only occur when there is an ingress added to another namespace (e.g. hub23), and the question of what to do by default is perhaps which would be least surprising / disruptive:

1. ingress works without adding another ingress controller, or
2. ingress needs its own ingress controller, and the one from this chart won't mess with it

minrk:
BTW, Just did a build & launch at https://turing.mybinder.org and everything worked great. Ready to let it back in, or is there more to test?

sgibson91:
Should be fine to re-add - I also haven't redeployed Hub23 onto the cluster yet, but I'm going to pin ingress-nginx to the same version as on mybinder.org which should hopefully avoid the original bug, and I've also turned off the auto-upgrade bot as well (because it's broken and I need to fix it!)

sgibson91:
Re-adding the Turing in #1896

PR #1909, opened 2021-05-17
Fix outdated configuration reference (nginx -> traefik -> not relevant and removed)

consideRatio:
I was trying out https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2200 which includes a schema validation that can also detect references to values that aren't recognized. Just using `helm template .` made me spot this reference to `nginx` in the JupyterHub helm chart configuration. This probably refers to `traefik` which is now running in the autohttps pod rather than `nginx` which was running there before (perhaps 0.9.X or earlier).

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -254,6 +254,9 @@ binderhub:
       chp:
         # FIXME: remove when hub chart dependency is bumped
         # to include https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2081
+        #
+        # This is when we update to a binderhub that reference z2jh 1.0.0 instead of z2jh 0.11.1.
+        #
         image:
           tag: 4.3.1
         networkPolicy:
@@ -274,7 +277,7 @@ binderhub:
           # - --default-target=http://$(PROXY_PATCHES_SERVICE_HOST):$(PROXY_PATCHES_SERVICE_PORT)
           - --error-target=http://$(PROXY_PATCHES_SERVICE_HOST):$(PROXY_PATCHES_SERVICE_PORT)/hub/error
           - --log-level=error
-      nginx:
+      traefik:
         resources:
           requests:
             memory: 512Mi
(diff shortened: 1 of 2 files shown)


mybinder/values.yaml
```diff
@@ -274,7 +277,7 @@ binderhub:
           # - --default-target=http://$(PROXY_PATCHES_SERVICE_HOST):$(PROXY_PATCHES_SERVICE_PORT)
           - --error-target=http://$(PROXY_PATCHES_SERVICE_HOST):$(PROXY_PATCHES_SERVICE_PORT)/hub/error
           - --log-level=error
-      nginx:
+      traefik:
```
choldgraf, line 280: Hmmm wouldn't something be broken if this were the wrong key name though?

consideRatio: The autohttps pod (running traefik) would not get the explicit resource requests, which could break something but doesn't have to break something.

The pod would still run with the default requests (guarantees/limits in CPU/memory). What the default requests would be depends on the k8s cluster configuration though.

It could be that this makes things more reliable on one but not all of our k8s clusters in the federation, or that it didn't matter, or it could matter in certain situations based on what other pods compete for resources on the node where the pod is scheduled.

choldgraf: > It could be that this makes things more reliable on one but not all of our k8s clusters in the federation, or that it didn't matter, or it could matter in certain situations based on what other pods compete for resources on the node where the pod is scheduled

Well that just about covers all possibilities then 😅

This seems like a reasonable fix to me! I may not be the most well qualified to determine this though

betatim: What does this section of the config control? AFAIK mybinder.org does not use the autohttps pod, so if this config is only used by the autohttps pod then maybe we should just remove it because it is unused

consideRatio: Ah, it is only used for the autohttps pod. Removed it in afee564

choldgraf (approved):

PR #1912, opened 2021-05-28
Add @callummole to list of secret holders

sgibson91:
I've passed the git-crypt secret to @callummole so he can edit Turing secrets if required

PR #1914, opened 2021-05-28
Pin Sphinx and other docs deps

manics:
Sphinx 4.0.2 has broken the docs build: https://github.com/jupyterhub/mybinder.org-deploy/pull/1912#issuecomment-850399699

To avoid problems like this in future this PR pins all docs dependencies and configures dependabot to bump versions for docs only.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: pip
+    directory: /docs
+    schedule:
+      interval: weekly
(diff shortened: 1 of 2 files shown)


betatim:
I thought that dependabot doesn't need configuring, but that is based on it seemingly "at random" turning up on repos, not on actual knowledge :-/

manics:
@betatim I've opened https://github.com/jupyterhub/mybinder.org-deploy/issues/1919 to discuss dependabot config.

PR #1924, opened 2021-06-02
bump prometheus memory limit to 23G

minrk:
WAL startup needs more than 20G, node limit is 24

we are getting close to needing bigger core nodes just for prometheus

cc @betatim

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -119,7 +119,8 @@ prometheus:
         memory: 20Gi
       limits:
         cpu: "2"
-        memory: 20Gi
+        # 20Gi isn't enough to rebuild WAL on startup
+        memory: 23Gi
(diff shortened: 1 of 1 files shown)


minrk:
We probably want to investigate our metrics and/or prometheus config to see if there's anything we can do to reduce prometheus resource usage, especially during WAL replay.

betatim:
What do you think drives memory usage? We didn't use to have this problem, so my guess is somehow it scales with number of "data points".

We could check for how many days we keep data, maybe that is a way to reduce memory. At least in the past I think we didn't pay (much) attention to how long our history is, so it might be set to some value that is larger than it needs to be?

If memory scales with "data points per unit time" it gets harder because we are getting more traffic per unit time now :-/

PR #1925, opened 2021-06-02
Increase prometheus memory limit and remove node selector

betatim:
Let's see if we can use a user node pool instance and more RAM to get
prometheus to start again. This will make debugging easier. Long term we
need a solution where pprometheus uses less memory.

Follow up PR to #1924

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -111,16 +111,16 @@ grafana:
 
 prometheus:
   server:
-    nodeSelector: *coreNodeSelector
-    livenessProbeInitialDelay: 600
+    #nodeSelector: *coreNodeSelector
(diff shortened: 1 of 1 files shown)


manics:
This issue looks relevant: https://github.com/prometheus/prometheus/issues/6934
One option is to delete the WAL logs. The [devs have also requested additional info](https://github.com/prometheus/prometheus/issues/6934#issuecomment-828112915) to help them investigate the issue.

betatim:
Min also suggested deleting the WAL so it sounds like we have some consensus on that. Would be interesting to work out how much data we'd lose, but I think that is mostly academic as getting it started again without needing 40GB of RAM is worth a lot of lost data (to me).

PR #1953, opened 2021-07-02
Re-deploy build tokens

minrk:
Re-bumps binderhub to get build tokens

This time only deploying to staging while we figure out the origin check issue

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -132,6 +132,17 @@ class ProxyHandler(RequestHandler):
         target_url = self.host + uri
 
         headers = self.request.headers.copy()
+        # don't override Host, which kubernetes uses for routing,
+        # otherwise this will be an infinite loop proxying to ourself
+        # Host will be set to the target Host by the request below.
+        current_host = headers.pop("Host", None)
+        if current_host:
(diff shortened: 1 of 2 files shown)


images/federation-redirect/app.py
```diff
@@ -132,7 +132,14 @@ async def get(self):
...
 
         headers = self.request.headers.copy()
-        headers["Host"] = self.host[8:]
+        # don't override Host, which kubernetes uses for routing,
+        # otherwise this will be an infinite loop proxying to ourself
+        headers.pop("Host", None)
+        if "origin" not in self.request.headers and "host" in self.request.headers:
+            # preserve original Host in Origin so this looks like a cross-origin request
+            # even though it's proxied
+            headers["Origin"] = f"https://{self.request.headers['host']}"
```
yuvipanda, line 141: Since "Host" won't be set, let's set X-Forwarded-Host? Just good practice I think?

minrk: Host will be set by the request below, since it's a default header. Setting it here is *overriding* Host. I've updated with more comments and setting x-forwarded-host

minrk:
Confirmed working on staging, deploying to prod

PR #1957, opened 2021-07-02
Remove note about not being able to deploy Chart.yaml changes only to staging

minrk:
I did exactly this in #1953 so it seems to work okay. Is there anything I missed, or is this just an out-of-date note?

### DIFF
--- a/docs/source/deployment/how.md
+++ b/docs/source/deployment/how.md
@@ -177,6 +177,12 @@ master.
 
 ### Deploying to *only* `staging`
 
+```eval_rst
+.. note::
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):
LGTM! If you've tested and it was fine, then great.

~~Could we also make it clearer that the `test-staging` label doesn't work for PRs from forks? OR is that a separate PR?~~ Scratch that. #1959 will allow us to do that.

minrk:
Added the note about not working from forks

sgibson91:
Looks great, thanks!

PR #1959, opened 2021-07-02
Use a staging environment to test PRs from forks

sgibson91:
This PR adds a staging environment to our Continuous Deployment workflow. This will allow us to test PRs that come from forks on the staging cluster. When a PR is opened the `jupyterhub/mybinder-org-operators` team will be notified to approve the run in the environment.

Draft for now until I've added the secrets to the env :grin:

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -36,8 +36,13 @@ jobs:
   # In this dedicated job to deploy our staging environment we build and push
   # images that the jobs to deploy to the production environments depend on.
   staging-deploy:
+    # Run this job in the staging environment
+    # This means that PRs from forks can be tested on staging before being merged
+    environment: staging
     # Only run the job if the 'test-staging' label is present OR if the event
     # is a push to master
+    # NOTE: Could deprecate the ''test-staging'' label in favour of the environment
+    # above
     if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -36,8 +36,13 @@ jobs:
   # In this dedicated job to deploy our staging environment we build and push
   # images that the jobs to deploy to the production environments depend on.
   staging-deploy:
+    # Run this job in the staging environment
+    # This means that PRs from forks can be tested on staging before being merged
+    environment: staging
     # Only run the job if the 'test-staging' label is present OR if the event
     # is a push to master
+    # NOTE: Could deprecate the ''test-staging'' label in favour of the environment
```
minrk, line 44: I think we still want a mechanism to explicitly trigger deploy to staging, not on every push to every PR.

sgibson91: Got it, removed that comment

sgibson91:
Ok, I think I have all the secrets in the environment now, shall we test?

minrk:
Yeah!

minrk:
Merged because I assume a PR cannot grant itself access to an environment (that seems like it would be bad)

PR #1961, opened 2021-07-02
Set imagePullSecret.create, not .enabled

yuvipanda:
From looking at the logic in z2jh, singleuser pods
don't get this automatically set without this explicitly
being set

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/26ece27d5aed76dd47828368f0569b5f3b02bc77/jupyterhub/files/hub/jupyterhub_config.py#L188

Follow-up to #1956


config/ovh.yaml
```diff
@@ -45,7 +45,6 @@ binderhub:
 
   jupyterhub:
     imagePullSecret:
-      enabled: true
```
consideRatio, line 48: @yuvipanda this was problematic!

When ovh.yaml is passed, i think this configuration being merged will erase the z2jh default values because it is not a dict but null value you are passing now.

yuvipanda: Is this the difference between `imagePullSecret: {}` and `imagePullSecret:` ?

consideRatio: Yepp!

`{}` would merge into the z2jh default values object without modifying anything.
`` would reset the z2jh default values object to `null`

If you for example have a config like...

```yaml
# val.yaml
hub:
proxy:
  secretToken: asdfasdf
```

I think this would break everything because the default values of hub would be reset. Yepp confirmed!

```shell
# executed from z2jh folder

tools/generate-json-schema.py

helm template ./jupyterhub --values val.yaml

Error: values don't meet the specifications of the schema(s) in the following chart(s):
jupyterhub:
- (root): hub is required
```

PR #1962, opened 2021-07-02
Use new name for hub secret

yuvipanda:
Part of z2jh 1.0


consideRatio:
This is fine for mybinder.org-deploy's Helm chart, but one could make it a bit more complicated yet robust by doing the following change instead.

```yaml
          - name: CONFIGPROXY_AUTH_TOKEN
            valueFrom:
              secretKeyRef:
                name: {{ include "jupyterhub.hub.fullname" . }}
                key: proxy.token
```

Some related documentation about this can be found here:
https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html

If this was the binderhub helm chart I would suggest we do it using the globally accessible defined helm template.

PR #1937, opened 2021-06-22
explicity set dockerconfig to point towards registry-1.docker.io

MridulS:
For federation members using Docker Hub as a registry, we saw that build requests always results in a rebuild even when an image with the same git ref existed in the registry. This was accompanied by the following error message in the hub logs

```
[E 210622 07:43:08 web:1789] Uncaught exception GET /binder/health (xx.xx.xx.xx)
HTTPServerRequest(protocol='https', host='notebooks.gesis.org', method='GET', uri='/binder/health', version='HTTP/1.1', remote_ip='35.192.107.35')
Traceback (most recent call last):
File "/usr/local/lib/python3.8/site-packages/tornado/web.py", line 1704, in _execute
result = await result
File "/usr/local/lib/python3.8/site-packages/binderhub/health.py", line 213, in get
overall, checks = await self.check_all()
File "/usr/local/lib/python3.8/site-packages/binderhub/health.py", line 200, in check_all
check.update(result)
TypeError: 'NoneType' object is not iterable
```

This was caused by https://github.com/docker/hub-feedback/issues/2113 and this PR is a temporary hotfix to resolve this issue on the Turing cluster.

Config added:

```yaml
config:
  DockerRegistry:
    url: "https://registry-1.docker.io"
    auth_config_url: "https://index.docker.io/v1"
    token_url: "https://auth.docker.io/token?service=registry.docker.io"
```

PR #1989, opened 2021-07-15
Revert docker hub registry URL changes for Turing, it's fixed in upstream Binderhub

MridulS:
This PR reverts https://github.com/jupyterhub/mybinder.org-deploy/pull/1937

The following config is **removed** from Turing cluster:
```
config:
  DockerRegistry:
    url: "https://registry-1.docker.io"
    auth_config_url: "https://index.docker.io/v1"
    token_url: "https://auth.docker.io/token?service=registry.docker.io"
```

Binderhub has the updated registry URL https://github.com/jupyterhub/binderhub/pull/1334


yuvipanda:
@MridulS do you know how we can test this? I'm a bit out of the loop

MridulS:
I'm not sure if there is a best way to test this before deploy but depending on how YOLO you are feeling, feel free to go ahead and merge this in and I can keep an eye out on the Turing cluster. To check if everything is okay, something like https://turing.mybinder.org/v2/gh/binder-examples/requirements/HEAD should open up directly by using the built image and not to try build it again.

I just reverted the changes at GESIS and everything looks okay here (we use docker hub too).

yuvipanda:
Hmm, so https://github.com/jupyterhub/mybinder.org-deploy/pull/1984 was the last binderhub bump - and I don't think it includes your changes. So maybe henchbot didn't pick up that merge?

MridulS:
ahh yes, it hasn't yet.

The change would still work because Docker folks anyway brought up the old registry URL too, but they did mention that `https://registry-1.docker.io` is the recommended one.

consideRatio:
BinderHub is patched in https://github.com/jupyterhub/mybinder.org-deploy/pull/1991, I'll go for a merge here.

PR #1999, opened 2021-07-29
Cleanup secrets that are now stashed in k8s Secrets

consideRatio:
This PR removes proxy.secretToken, hub.cookieSecret, and hub.services.binder.apiToken configuration from our secrets in git, as they are now stashed in k8s Secret resources and read from there as of #1998 successfully deployed that bumped to z2jh 1.1+

I removed some boilerplate declaring admin: true, but that is the default anyhow in binderhub.

https://github.com/jupyterhub/binderhub/blob/ab12fc621c182e6283078c8fc7edb1743279280e/helm-chart/binderhub/values.yaml#L127-L130


config/test-secrets.yaml
```diff
@@ -11,10 +11,7 @@ binderhub:
   jupyterhub:
     hub:
       services:
-        binder:
-          apiToken: "abc123"
-    proxy:
-      secretToken: "abc123"
+        binder: {}
```
consideRatio, line 14:
```suggestion
```

config/test-secrets.yaml
```diff
@@ -11,10 +11,7 @@ binderhub:
   jupyterhub:
     hub:
       services:
```
consideRatio, line 13:
```suggestion
```

minrk (approved):

PR #2002, opened 2021-08-18
Use test-this-pr action for Pull Requests made from forks

sgibson91:
### The Problem

Our usual workflow is "fork the repo -> develop -> open PR". The only downside to this is that repository secrets are not made available to PRs from forks by default. However, sometimes we want to be able to test a PR in our staging environment before merging and deploying for real.

### What does this PR do?

We add the `test-this-pr.yml` workflow file which will run the [`test-this-pr-action`](https://github.com/sgibson91/test-this-pr-action) (written by me) when triggered by a comment containing `/test-this-pr` left by someone with appropriate rights to the repo (OWNER, COLLABORATOR, MEMBER).

This action basically takes the `pull/{PR_NUM}/merge` ref and creates a new branch in **this** repo called `test-this-pr/{PR_NUM}`. The `staging-deploy` job in `cd.yaml` has now been configured to be triggered by **both** pushes to `master` and `test-this-pr/**` branches, so it will start running and now have access to secrets! The action will also leave a comment on the original PR with a link to the GitHub Actions page of the branch to track progress.

A `report-status` job has also been added to the `cd.yml` file which will **only run on pushes to `test-this-pr/**` branches**. This job will poll the GitHub REST API to check when `staging-deploy` has completed. Upon receiving confirmation that the job has finished, it will post a `success` or `failure` comment to the original PR and then delete the `test-this-pr` branch that was created by the `test-this-pr-action`. This is so that the single source of truth is **always** the original PR, not any automatically generated branches.

Multiple `/test-this-pr` comments can be left on the same PR.

### Some quirks

We need a GitHub PAT with `public_repo` scope to pass to `test-this-pr-action`. While giving the `GITHUB_TOKEN` [write permissions on contents and pull requests](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idpermissions) would actually satisfy our requirements, GitHub has configured Actions such that [workflows cannot be triggered by events that were authenticated by `GITHUB_TOKEN`](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#using-the-github_token-in-a-workflow). In short, `staging-deploy` won't run _even though_ a push to a `test-this-pr/**` branch has been made if we use `GITHUB_TOKEN`.

As a result of using a PAT, the comment made by the `test-this-pr-action` will look like it came from whoever provided the PAT. I don't think this is a major issue and I also don't mind providing the PAT, but just thought I'd highlight this quirk.

### Testing

You can see all the work I did to test this workflow in this repo: https://github.com/binderhub-test-org/pr-test

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -43,7 +43,7 @@ jobs:
       (github.event.label.name == 'test-staging') ||
       ((github.event_name == 'push') &&
       (github.ref == 'refs/heads/master')) ||
-      ((github.even_name == 'push') &&
+      ((github.event_name == 'push') &&
       contains(github.ref, 'test-this-pr'))
     runs-on: ubuntu-20.04
 
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -41,10 +41,8 @@ jobs:
     # is a push to master
     if: |
       (github.event.label.name == 'test-staging') ||
-      ((github.event_name == 'push') &&
-      (github.ref == 'refs/heads/master')) ||
-      ((github.event_name == 'push') &&
-      contains(github.ref, 'test-this-pr'))
+      ((github.event_name == 'push') && (github.ref == 'refs/heads/master')) ||
+      ((github.event_name == 'push') && contains(github.ref, 'test-this-pr'))
     runs-on: ubuntu-20.04
 
     strategy:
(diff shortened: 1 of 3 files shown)


minrk:
Looks great! Does it make sense to use an `environment` to supply the PAT with elevated permissions to only these test-pr branches? I guess that would mean making test-this-pr its own workflow instead of a job within the cd workflow?

sgibson91:
> making test-this-pr its own workflow instead of a job within the cd workflow?

Test-this-pr is already in it's own workflow file `test-this-pr.yml`, the extra jobs in cd.yml only use the GitHub token.

minrk:
ah, ok. I misunderstood which token was used where.

Yay! So happy to have this.

So the only step remaining is to issue the `secrets.ACCESS_TOKEN`, right? Does putting it in an `environment` still make sense so it's not available to workflows by default?

sgibson91:
> So the only step remaining is to issue the `secrets.ACCESS_TOKEN`, right? Does putting it in an `environment` still make sense so it's not available to other privileged workflows by default?

Correct, and yes I think so :) Especially since it's linked to someone's account. On reflection, I think this is why `/ok-to-test` requires setting up a GitHub App for OAuth, to circumvent this.

sgibson91:
Also, feel free to rename `ACCESS_TOKEN` to something more descriptive!

sgibson91:
I set up an environment called `test-this-pr-env` and added a PAT from my account called `TEST_THIS_PR_ACCESS_TOKEN` - It only has the `public_repo` scope. I then updated the `test-this-pr.yml` file to reference both the environment and the new secret name.

I think this is ready to merge and test!

minrk:
Yay!

PR #2009, opened 2021-09-06
JupyterLab 3.0: bump image prefix across the federation

minrk:
The order these things should be done in (they can be simultaneous, but some should not be before others):

1. update jupyterlab default UI in repo2docker (merged https://github.com/jupyterhub/repo2docker/pull/1035)
2. invalidate image cache by bumping image prefix (this pr). Cache should be invalidated with/after the r2d bump so we can assume new builds have lab 3.1.
3. bump binderhub filepath handling to use jupyterlab https://github.com/jupyterhub/binderhub/pull/1288 . This change is where we start to *apply* the assumption that cached builds have jupyterlab 3, and provide consistent UI for both filepath and default URLs.

This PR currently does all 3 (includes #2011 and #2014)

Purging the image cache means *lots* of new builds, and we will probably want to sweep out our old cache (related: #2012)

PR #2012, opened 2021-09-06
delete-old-image: support cutoff date

minrk:
supports cutoff date via `--delete-before`, deleting images older than a certain age, and caching for more reasonable performance across multiple tries

also begins support for cleaning out other registries. OVH almost works, but is extremely slow. Docker Hub (turing) needs special auth

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,40 @@
+# scripts for managing mybinder.org
+
+
+## delete-old-images.py
+
+This script talks to the docker registry API and tries to delete old images.
+The script has the following conditions to check:
+
+- images that don't match the build prefix (e.g. changing the BinderHub.image_prefix in configuration)
+- limiting the number of builds stored for a given repo (`--max-builds`)
+- deleting all images older than a certain date (`--delete-before`)
+
+Use:
+
+
+```bash
+python3 scripts/delete-old-images.py prod --dry-run
+```
+
+This script is not run automatically.
+
+## prune_harbor.py
+
+Harbor registries have far more powerful garbage collection and image retention policies than the basic docker registry implementation,
+or Google Container Registry.
+The only missing piece is removing _repositories_ that have no artifacts in them,
+which seems to affect the performance of those policies and other aspects of harbor.
+
+This script only deletes repositories with no images in them.
+
+Right now, this script is run for our harbor-using federation members
+(currently OVH),
+via the [prune-harbor workflow](../.github/workflows/prune-harbor.yaml).
+Credentials are found in `secrets/{name}-harbor.yaml`.
+
+Use:
+
+```bash
+python3 scripts/prune_harbor.py ovh2
+```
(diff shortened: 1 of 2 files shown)


minrk:
thinking of #2009, running a dry-run on prod, which has 48k repositories, running

```bash
./scripts/delete-old-images.py prod --delete-before=2021-02-01 --dry-run
```

would delete 68,722 images, totalling about 73TB.

betatim:
The JupyterLab updated happened, did you use this script to clean up the registry for that? -> Should we merge this as is? (Can always come back to improve it later)

minrk:
I never did run this, and part of reducing costs related to https://github.com/jupyterhub/team-compass/issues/463. I'm going to have a go at just deleting the very old images today (before 2021-02-01) which is still a huge fraction of our current storage costs ($1600/mo) since operational costs on GCP may soon become tight!

minrk:
I tried running this for OVH to deal with #2514, but Harbor doesn't actually support the part of the docker registry API that's needed to do this.

Fortunately, Harbor has its _own_ API, which is even easier to use and has more useful metadata for culling images (basically what Harbor's internal GC is meant to be doing, but isn't for some reason). So I ran a much simpler script:

```python
import os

import requests
from dateutil.parser import parse as parse_date

harbor_url = "https://2lmrrh8f.gra7.container-registry.ovh.net/api/v2.0"
project_name = "mybinder-builds"
date_cutoff = parse_date("2023-02-01T00:00Z")

username = "mybinder-admin"
#  export HARBOR_PASSWORD=$(terraform output -raw registry_admin_password)
password = os.environ["HARBOR_PASSWORD"]

# s = requests.Session()
# s.auth = (username, password)
while True:
    r = requests.get(
        harbor_url + f"/projects/{project_name}/repositories",
        params=dict(sort="update_time", page_size="100"),
        auth=(username, password),
    )

    repos = r.json()
    r.raise_for_status()
    for repo in repos:
        project_name, repo_name = repo["name"].split("/", 1)
        print(repo)
        if parse_date(repo["update_time"]) > date_cutoff:
            break
        r = requests.delete(
            harbor_url + f"/projects/{project_name}/repositories/{repo_name}",
            auth=(username, password),
        )
        r.raise_for_status()
```

It took a while, but it accomplished mostly the same result, and with far fewer API calls (deleted whole repos not updated in the last 6 weeks, rather than _artifacts_).

consideRatio:
Should we go for a merge of this @minrk? I figure it doesn't require much review if its a manually used utility script. So right now, maybe we could do:

- [ ] Copy paste simpler script in https://github.com/jupyterhub/mybinder.org-deploy/pull/2012#issuecomment-1471864870 to a new file
- [ ] Add scripts/README.md and describe that these two scripts can be used complementary or similar

consideRatio (approved):
Looks like a very clear improvement without risk of breaking changes as these are manually run files.

Let's go for a merge!

PR #2017, opened 2021-09-09
use test-this-pr access token to report status

minrk:
since the default token doesn't have permission

@sgibson91 is this right? Does the access token have the permissions it needs to do these steps (delete branch & post comment)?

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -277,6 +277,7 @@ jobs:
     #
     # This job will only run on push events made to test-this-pr/** branches
     if: github.event_name == 'push' && contains(github.ref, 'test-this-pr')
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):
Yes, this should work and use my token instead.

I think the reason why this worked for me in my test repository but doesn't here is because the [GitHub token has the restricted permissions](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) set in the repo (possibly org-wide), whereas my repo probably has the permissive set.

Screenshot 2021-09-09 at 15 51 41

I'm not saying we should change this setting as there is good reason to have a restricted token! Just explaining where the issue arose from :)

minrk:
OK, let's try it!

PR #2032, opened 2021-09-21
Autoban hundreds of mining pools

minrk:
Retrieves a public list of mining pools, bans them all by both ip and DNS

main structural change: match `secrets/config/common/*.yaml` pattern already used in `config/common` instead of single `secrets/config/common.yaml` file.

List is fetched and updated in `secrets/ban.py`


choldgraf:
Is this something that will update with deploys as a central list is updated elsewhere?

minrk:
> Is this something that will update with deploys as a central list is updated elsewhere?

Yes, it's an externally maintained list, explicitly for circumventing DNS-based blocks of mining like we do, but we use the same information to make sure they are all blocked by both DNS and ip. The list is fetched on each deploy.

PR #2043, opened 2021-10-05
use 503 error code for requests to stopped servers

minrk:
I think JupyterLab retrying on 404 is driving most of the traffic to the Hub. https://github.com/jupyterlab/jupyterlab/issues/11207

Using another error code that doesn't already mean something for the REST API seems to behave better, based on https://github.com/jupyterhub/jupyterhub/pull/3636#issuecomment-934174628

503 might be even more appropriate, since that's what JupyterHub 1.x uses and JupyterLab expects for a stopped server, though that may [change in 2.0][].

[change in 2.0]: https://github.com/jupyterhub/jupyterhub/pull/3636

**Edit:** switched to 503, which behaves slightly better than 424 with current JupyterLab.

### DIFF
--- a/mybinder/templates/proxy-patches/configmap.yaml
+++ b/mybinder/templates/proxy-patches/configmap.yaml
@@ -8,15 +8,15 @@ data:
     server {
       listen 80;
 
-      # Return 404 for anything that comes here
+      # Return 424 for anything that comes here
       # serve an informative error about stopped pods
-      error_page 404 /404.html;
-      location /404.html {
+      error_page 424 /424.html;
+      location /424.html {
         root /srv/proxy-patches;
         internal;
       }
       location / {
-        return 404;
+        return 424;
       }
     }
 
@@ -26,7 +26,7 @@ apiVersion: v1
 metadata:
   name: proxy-html-files
 data:
-  404.html: >
+  424.html: >
     <html>
     <head>
     <title>Binder not found</title>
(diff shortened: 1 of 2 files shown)


sgibson91:
This Pull Request is now being tested :tada: See the test progress in [GitHub Actions](https://github.com/jupyterhub/mybinder.org-deploy/actions?query=branch%3Atest-this-pr%2F2043).

sgibson91:
**Job status:** success
Branch 'test-this-pr/2043' has been deleted

minrk:
/test-this-pr

424 is still retried, but not at the same rate as 404. Try again with 503.

sgibson91:
This Pull Request is now being tested :tada: See the test progress in [GitHub Actions](https://github.com/jupyterhub/mybinder.org-deploy/actions?query=branch%3Atest-this-pr%2F2043).

sgibson91:
**Job status:** success
Branch 'test-this-pr/2043' has been deleted

minrk:
Tried with 424 and 503, and there isn't a big difference, but both are a lot better than the 404 we have now.

404 gives this error:

Screen Shot 2021-10-05 at 10 04 51

and a rapid flood of failing requests for /api/contents.

Both 424 and 503 get continuous retries as well because [this logic in the hub extension][hub-extension] is not triggered (we should have an extension that triggers the same logic and mybinder.org-appropriate custom lost-connection message).

[hub-extension]: https://github.com/jupyterlab/jupyterlab/blob/e671140fe7ed03b53de09e935cf8ac399e939cae/packages/hub-extension/src/index.ts#L156-L159


503 will periodically show this accurate message:

Screen Shot 2021-10-05 at 13 56 33

whereas 424 will show no messages at all until the user tries to take specific actions that fail, such as opening a notebook (with 424 instead of 503):

Screen Shot 2021-10-05 at 13 57 17

minrk:
Given JupyterLab's expectations, I think we should stick with 503 here and move to 424 when repo2docker bumps jupyterlab to include https://github.com/jupyterlab/jupyterlab/pull/11205

Issue #2039, opened 2021-10-01
University of Michigan workshop request


cteplovs:
### 1. Who you are
Chris Teplovs, Lecturer & Research Investigator, University of Michigan
### 2. A few details about the event/course/context
I'm running a workshop to introduce business students to Python:
"This session is a hands on experience for students to learn the fundamentals of Back End development through Python coding. During this bootcamp the School of Information Professor Chris Teplovs will lead students through a session to teach Python Basics."

### 3. If money is changing hands (attendees pay a fee, speaker is getting paid, free, etc)
It's free!

### 4. A link to the repository you are using
https://github.com/cteplovs/rossplustech-python
(it's not done yet)

### 5. What type of event it is:
- [ ] talk
- [X] workshop
- [ ] lecture course
- [ ] training session

### 6. How many people you expect to attend the event
60-100, but there might be remote participants that push that number to about 120.

### 7. The exact times and dates of the event(s) (with timezone information)
Wednesday, October 8th, 5:30pm to 8:00pm, Eastern Time Zone

### 8. How do you plan to give mybinder.org a shout out at the event
I'm planning to tell participants that the functionality that allows them to participate is thanks to mybinder.org!

minrk:
@cteplovs feel free to open a PR like [this one](https://github.com/jupyterhub/mybinder.org-deploy/pull/1911/files) bumping the quota for your repo to 200.

rgaiacs:
Closing because the workshop already happen.

PR #2047, opened 2021-10-06
update values.yaml for Ross workshop

cteplovs:
Per instructions, here's my PR for today's workshop:
(@cteplovs feel free to open a PR like this one bumping the quota for your repo to 200.)


consideRatio:
This was a PR associated with #2039 and we didn't get it merged before the actual workshop :scream:

What do we do now? How did it go @cteplovs?

PR #2092, opened 2021-12-21
bump turing pod quota to 200

callummole:
To pick up some of the short-term GKE capacity loss (see [gke.mybinder.org](https://github.com/jupyterhub/team-compass/issues/463) issue) we bump Turing up to 200. Based on current costs the Turing credits will last for just over a month, but we can apply for more in the new year.

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -7,7 +7,7 @@ letsencrypt:
 binderhub:
   config:
     BinderHub:
-      pod_quota: 80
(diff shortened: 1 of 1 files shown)


minrk:
The load shouldn't change too suddenly. But we can scale it down if need be. I'm okay either way. 👍 to deploy.

minrk (approved):

betatim:
Merged because YOLO

PR #2096, opened 2021-12-22
apply pod quotas to prime federation members

minrk:
now that pod quotas are enforced, we can have a limit on the 'prime' members, allowing cost mitigations

federation-redirect still ignores prime quotas, but launches will be rejected if they would exceed the limit

It might be good to have https://github.com/jupyterhub/binderhub/pull/1442 before we start doing this

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -8,6 +8,7 @@ coreNodeSelector: &coreNodeSelector
 binderhub:
   config:
     BinderHub:
+      pod_quota: 400
       build_node_selector: *userNodeSelector
       hub_url: https://hub.gke2.mybinder.org
       badge_base_url: https://mybinder.org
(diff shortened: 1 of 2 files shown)


manics:
I've just merged https://github.com/jupyterhub/mybinder.org-deploy/pull/2097 which includes https://github.com/jupyterhub/binderhub/pull/1442

betatim:
I edited https://grafana.mybinder.org/d/fZWsQmnmz/pod-activity?orgId=1&refresh=1m&viewPanel=7 to add the two new metrics but I don't quite understand why there are so many overlapping lines for the same repo. Is it because we run two binderhubs on the cluster? If someone has a smart idea, let me know. I am a bit stumped.

betatim:
Another interesting metric to look at is https://grafana.mybinder.org/d/fZWsQmnmz/pod-activity?orgId=1&refresh=1m&viewPanel=9 which looks like a lot of launch failures but I think they are explained by the repo quota kicking in.

PR #2102, opened 2022-01-18
terraform: switch to pd-balanced, reduce size to save cost

minrk:
from 1T. Should save ~$1k/month

Ignore the version bumps, which are only to make terraform match what's already been auto-upgraded to, so the new pool has the same version as what's already deployed.

This creates a new user pool to replace the old one, but leaves the existing one with an autoscale max of 1 so it will slowly drain (will need some help with cordoning). Once it's drained, another PR can actually delete the old pool.

I'll do the apply after this is merged. I've synced the versions, but not created the new pool.

xref: [cost calculations](https://github.com/jupyterhub/team-compass/issues/463#issuecomment-998781376)

### DIFF
--- a/terraform/prod/main.tf
+++ b/terraform/prod/main.tf
@@ -12,7 +12,7 @@ provider "google" {
 }
 
 locals {
-  gke_version        = "1.17.14-gke.400"
+  gke_version        = "1.19.14-gke.1900"
   location           = "us-central1" # for regional clusters
   federation_members = ["gke-old", "gesis", "turing", "ovh"]
 }
@@ -61,6 +61,14 @@ resource "google_container_node_pool" "core" {
       disable-legacy-endpoints = "true"
     }
   }
+
+  # do not recreate pools that have been auto-upgraded
+
+  lifecycle {
+    ignore_changes = [
+        version
+    ]
+  }
 }
 
 resource "google_container_node_pool" "user" {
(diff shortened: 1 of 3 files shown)


terraform/prod/main.tf
```diff
@@ -71,6 +79,50 @@ resource "google_container_node_pool" "user" {
   node_locations = ["${local.location}-a"]
   version        = local.gke_version
 
+  autoscaling {
+    min_node_count = 0
+    max_node_count = 1
```
minrk, line 84: This should not trigger immediate scale-down. Autoscale doesn't force these bounds to be satisfied continuously. Instead, it should only prevent scale-up.

yuvipanda:
Would this cost us performance in terms of docker pull throughput? That was why we made them big SSDs to begin with, as performance unfortunately scales linearly with size.

The newer `pd-balanced` type is also something to try to cut cost. That's what I now use on my berkeley clusters.

minrk:
> Would this cost us performance in terms of docker pull throughput?

I have no idea. I don't think we've done any measuring on how much Disk IO limits pulls vs network IO. [gcloud docs](https://cloud.google.com/compute/docs/disks/performance#cpu_count_size) suggest a 512GB SSD should have 384/768 MBps read/write performance. That seems very fast! But it's ~1/2 what we should have now.

I think we should probably switch to pd-balanced and see what happens. Switching to pd-balanced without losing size would save 40%. We could save the same 50% if we dropped to 800, so I did that. I didn't know about pd-balanced! I switched the core pool to that as well.

yuvipanda (approved):

minrk:
OK, I'll give this a go.

PR #2113, opened 2022-01-28
federation-redirect: check versions before health

minrk:
checking versions on the prime host updates the version for other hosts to be checked against

if the prime host is unhealthy (e.g. full), the version check was skipped, which means the prime version will not be updated.

This can result in starving fully up-to-date federation members as long as the prime host is full (which will be more likely to stay that way, as we don't stop routing requests to prime while it is full).

Sequence of events:

1. prime host is full, reference version not updated
2. deploy an upgrade
3. other members get latest version, are considered unhealthy
4. prime member gets all traffic, so is _more_ likely to stay 'full'
5. only when traffic drops such that the prime host passes health check (no longer full) will the version be updated and other members be considered healthy again.

That's why it looks like this:

Screen Shot 2022-01-28 at 08 57 45

where at 18:00 traffic to GKE dropped below its pod quota so it became healthy again, and the version was updated, admitting all other members back into the federation.

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -254,7 +254,10 @@ async def health_check(host, active_hosts):
                 versions = json.loads(response.body)
                 # if this is the prime host store the versions so we can compare to them later
                 if all_hosts[host].get("prime", False):
-                    CONFIG["versions"] = versions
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
LGTM, nice catch!!!

Issue #1940, opened 2021-06-23
Turing Prometheus Server is blocking upgrades


sgibson91:
After pushing the docker fix yesterday (#1937) the Prometheus server on the Turing failed to rollout in time and the server pod went into a `CrashLoopBackOff` state. Turns out, this is what's causing the issue from `kubectl logs turing-prometheus-server-94df968dd-9m27h -c prometheus-server`:

```
level=error ts=2021-06-23T10:31:35.892Z caller=main.go:787 err="opening storage failed: open /data/wal/00000890: no space left on device"
```

At first I tried to bump the size of the PV in #1938, but that didn't effect the PV (because it's persistent, of course!) so reverted it in #1939, which lead to the following error in the [Turing deployment CD](https://github.com/jupyterhub/mybinder.org-deploy/runs/2886179011?check_suite_focus=true#step:16:54):

```
Error: UPGRADE FAILED: cannot patch "turing-prometheus-server" with kind PersistentVolumeClaim: PersistentVolumeClaim "turing-prometheus-server" is invalid: spec.resources.requests.storage: Forbidden: field can not be less than previous value
```

At this point, I stopped poking things before I made anything worse.

sgibson91:
Resizing the PVC/PV was the correct thing to do, but Azure Disk does not support disk resizing when it's in the attached state, therefore we have to stop the prometheus server pod in order to provide the PVC enough time to rescale the PV.

```
kubectl scale deployment turing-prometheus-server --replicas=0 
```

This stops the pod and gives the PVC the chance to recognise the size change. _However, the PV does not resize until the deployment is scaled back up and there is no event originating from the PVC to say that it's ready to resize._ So unfortunately, you have to play a waiting game. We waited ~20 mins before scaling the deployment back up.

```
kubectl scale deployment turing-prometheus-server --replicas=1
```

This triggered the PV resize and it began successfully reporting data.

cc: @callummole

sgibson91:
While the bug is fixed, I'm going to leave this open until it's documented somewhere in the SRE docs

lukehare:
I'm taking over from @callummole as the new Turing Binder cluster admin, so I'm just sharing my handle @lukehare here!

rgaiacs:
Turing is no longer part of the federation. 😢  Closing this issue as out of date.

PR #2120, opened 2022-02-02
increase turing prometheus size, reduce retention to 21d

minrk:
Recently failed to start due to lack of space

reference #1940 when this happened before, and for pv-resize process on AKS.

cc @callummole @sgibson91 since the increase in pv size increases turing costs (not by much, I think).

I think we can do either/or, too: just reduce retention, or just increase size, if folks prefer.

### DIFF
--- a/config/turing.yaml
+++ b/config/turing.yaml
@@ -97,7 +97,7 @@ grafana:
 
 prometheus:
   server:
-    retention: 30d
+    retention: 21d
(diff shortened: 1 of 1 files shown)


sgibson91 (approved):
This looks good for me! Thanks for catching it

minrk:
Resized successfully following steps in #1940

PR #2134, opened 2022-02-15
Disable tracking cookies in matomo

minrk:
To avoid cookie banners in EU, it seems we need to avoid using a state cookie for analytics

also sets matomo's own "respect DoNotTrack" option, which ought to be redundant, but no harm in being extra explicit.


manics (approved):
Relevant docs: https://matomo.org/faq/general/faq_157/

PR #2147, opened 2022-03-09
bump binderhub chart 8da69f2...d19a36a

minrk:
Gets pycurl in the binderhub image, which may improve health checks

https://github.com/jupyterhub/binderhub/compare/8da69f2...d19a36a


minrk:
Many health checks had been failing due to the SimpleAsyncHTTPClient's queue timeout (i.e. request timeout because the client itself is queing requests, not any problem with the destination)

Issue #2156, opened 2022-04-22
Set up donation buttons and supporters sections on mybinder.org


choldgraf:
As a part of https://github.com/jupyterhub/team-compass/issues/430 we decided to make the following pages to mybinder.org:

**On the home page and the loading page**

- Add a "support mybinder.org" call-to-action button. It points to https://numfocus.salsalabs.org/donate-to-binder

**On the home page**

Add a "supporters of mybinder.org" section below the fold. This would be [modeled after the jupyter.org sponsors page](https://jupyter.org/about#sponsors). It would have three sections:

- *mybinder.org federation members* - entities that deploy, fund, and staff their own BinderHub in the federation.
- *mybinder.org partner institutions* - entities that employ individuals that operate any hubs in the mybinder.org federation (if not full federation members).
- *mybinder.org supporters* - entities that provide significant financial resources or credits for any hubs in the mybinder.org federation ("significant" is defined as anything greater than $10,000 annually).

It would also link to a "support mybinder.org" page in the Binder docs that had more information about how Binder's donations work.

choldgraf:
I had a quick thought about using docs.mybinder.org to list this instead of mybinder.org directly, would love any feedback about that in this issue:

- https://github.com/jupyterhub/binder/issues/257

PR #2160, opened 2022-04-24
Adding a donate button to the banner

manics:
Attempting to add a donate button (https://github.com/jupyterhub/mybinder.org-deploy/issues/2156)

This is based on the launch button but white on yellow isn't very visible (which is perhaps also a problem for the launch button!)

image

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -126,7 +126,14 @@ binderhub:
       cors_allow_origin: "*"
 
       banner_message: |
-        <div style="text-align: center;">Thanks to <a href="https://cloud.google.com/">Google Cloud</a>, <a href="https://www.ovh.com/">OVH</a>, <a href="https://notebooks.gesis.org">GESIS Notebooks</a> and the <a href="https://turing.ac.uk">Turing Institute</a> for supporting us! 🎉</div>
+        <div style="display:flex">
+          <div style="flex:1;text-align:center;">
+            Thanks to <a href="https://cloud.google.com/">Google Cloud</a>, <a href="https://www.ovh.com/">OVH</a>, <a href="https://notebooks.gesis.org">GESIS Notebooks</a> and the <a href="https://turing.ac.uk">Turing Institute</a> for supporting us! 🎉
+          </div>
+          <a class="btn-submit" style="width:fit-content;height:fit-content;padding:10px;font-weight:bold;margin-top:-10px;" href="https://numfocus.salsalabs.org/donate-to-binder" target="_blank">
+            Support mybinder.org!
+          </a>
+        </div>
       about_message: |
         <p>mybinder.org is public infrastructure operated by the <a href="https://jupyterhub-team-compass.readthedocs.io/en/latest/team.html#binder-team">Binder Project team</a>.<br /><br />
         The Binder Project is a member of <a href="https://jupyter.org">Project Jupyter</a>, which is a fiscally
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -133,7 +133,7 @@ binderhub:
           <a class="btn" style="width:fit-content;height:fit-content;padding:10px;background-color:#e66581;color:white;font-weight:bold;margin-top:-16px;"
           onmouseover="this.style.backgroundColor='#d15b75'" onmouseout="this.style.backgroundColor='#e66581'"
           href="https://numfocus.salsalabs.org/donate-to-binder" target="_blank">
-            🤍 Donate to mybinder.org! 🤍
+            🤍 Donate to mybinder.org!
           </a>
         </div>
       about_message: |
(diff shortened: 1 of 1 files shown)


betatim:
What do you think about making the call to action "donate .." or something else that makes it clear this isn't about "supporting" in the many ways that is possible with open-source projects but only about money? My thinking was that in the context of an open-source project my first thought isn't "financial donation" when I see a "support this project" link. Instead I expect it to contain information about PRs, bugs, forum, how to contribute, etc and maybe info about donations. This made me think shouldd we make the label of the button super obviously about "helping with cash", so that those who are looking for a button like that know to click it? The main goal would be to make it easier to find the button if you are looking for it.

mybinder/values.yaml
```diff
@@ -126,7 +126,14 @@ binderhub:
       cors_allow_origin: "*"
 
       banner_message: |
-        <div style="text-align: center;">Thanks to <a href="https://cloud.google.com/">Google Cloud</a>, <a href="https://www.ovh.com/">OVH</a>, <a href="https://notebooks.gesis.org">GESIS Notebooks</a> and the <a href="https://turing.ac.uk">Turing Institute</a> for supporting us! 🎉</div>
+        <div style="display:flex">
+          <div style="flex:1;text-align:center;">
+            Thanks to <a href="https://cloud.google.com/">Google Cloud</a>, <a href="https://www.ovh.com/">OVH</a>, <a href="https://notebooks.gesis.org">GESIS Notebooks</a> and the <a href="https://turing.ac.uk">Turing Institute</a> for supporting us! 🎉
+          </div>
+          <a class="btn-submit" style="width:fit-content;height:fit-content;padding:10px;font-weight:bold;margin-top:-10px;" href="https://numfocus.salsalabs.org/donate-to-binder" target="_blank">
```
minrk, line 133: `btn btn-primary` has better contrast, if you want to try that

minrk, line 133: Maybe the link should be to @choldgraf's new [support My Binder page](https://mybinder.readthedocs.io/en/latest/about/support.html)?

choldgraf:
**Question**: Should we list "supporters of mybinder.org" on this home page, or use [our support page](https://mybinder.readthedocs.io/en/latest/about/support.html) for this?

manics:
I tried @choldgraf's suggestion of using the other mybinder colours
`#e66581` I quite like this one!
image

`#579aca` For comparison
image

With a slight darkening onmouseover:
#e66581 #d15b75
#579aca #4f8cb7

Not sure either about whether to go straight to the donation page or to go to the about pages...
However is there any chance we could update https://numfocus.salsalabs.org/donate-to-binder/index.html with a link back to https://mybinder.readthedocs.io/en/latest/about/support.html, and also replace the low-res PNG logo with the SVG from https://mybinder.org/static/logo.svg ?

choldgraf:
Nice! Thanks for these updates - a few quick thoughts:

- I like the red, I think it's more noticeable.
- I think two hearts might be a bit much - maybe just the heart to the left?
- One idea: we could try a button background of `#579aca` and a heart color of `#e66581`
- I like the idea of linking to the donation page directly, and that page having some of our own text that explains what donations are used for, and links back to support.
- I've asked Lisa @ NumFocus how we can update that text + the logo, will get back w/ a response

manics:
Here's a few colour combinations:
image

choldgraf:
I agree with your assessment (also, thanks so much for putting them all together).

I think in the name of simplicity, my vote would be for pink with white heart

choldgraf:
I'll also update the binder docs with a button of the same style once we finalize something, so we can have some consistency

minrk:
👍 to pink with white

manics:
This should be ready now! Unless we want to try and implement this by overriding a template (https://github.com/jupyterhub/binderhub/pull/1472). It would be much cleaner since we could hopefully use CSS classes, including the `:hover` selector instead of javascript `onmouse*` handlers.

choldgraf (approved):
This looks good to me - we can iterate later to improve things as well, IMO (e.g., when we add a below-the-fold list of supporters?)

If there's a "better" implementation and you'd like to do that instead of this one, I'm fine with that too now that we know the look-and-feel of the button that we'd like, but don't want to block progress just because there's a slightly better way to do it :-)

manics:
I don't know exactly how to convert this to a template in the Helm chart.... if no-one else can quickly convert it then probably best to leave it for the future?

choldgraf:
Hmmm - I didn't realize that the button was flush with the top of the screen...I feel like it would be better if it were centered with the other banner content. I'm not sure how to do this with BinderHub CSS, but I think these are the rules we'd need:

```css
#banner-container {
    padding: 8px;  /* To prevent the container from being much bigger */
}

#banner-container > div {
    align-items: center;
}

#banner-container .btn {
    margin-top: unset;  /* to cause centering via the flexbox rather than manually specifying.
}
```

manics:
@choldgraf That was intentional, the default header css has `padding: 16px`
https://github.com/jupyterhub/binderhub/blob/0b49d3aa9e8e59cc1d8904a505cb0eb98b8510ab/binderhub/static/index.css#L56
so adding the image without reducing the top padding leads to a header bar which tkes up too much vertical space:
image

We can't override the parent container CSS using a style in `banner_message` since it's rendered inside the `banner-container` div, there are some hacky ways we could do it but I think we're better off investing time in overriding the whole template using https://github.com/jupyterhub/binderhub/pull/1472 (assuming we can!)

manics:
Actually another option could be a new `c.BinderHub.css_overrides` configurable that gets inserted into the template `<head>` after https://github.com/jupyterhub/binderhub/blob/0b49d3aa9e8e59cc1d8904a505cb0eb98b8510ab/binderhub/templates/page.html#L17

PR #2161, opened 2022-04-27
Reinstate Turing cluster

callummole:
Ref: #495

The cluster is up and running again. Reinstating the Turing cluster to it's pre-paused levels (see #2150).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -177,11 +177,11 @@ jobs:
             hub_url: https://hub.gke.mybinder.org
             chartpress_args: ""
             helm_version: ""
-           - federation_member: turing
-             binder_url: https://turing.mybinder.org
-             hub_url: https://hub.mybinder.turing.ac.uk
-             chartpress_args: ""
-             helm_version: ""
+          - federation_member: turing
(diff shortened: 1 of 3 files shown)


minrk:
Woo!

minrk:
@callummole almost! Looks like while turing was down, we've updated to require k8s 1.19, and turing is still on 1.18. How easy is it for you to run a cluster upgrade?

minrk:
If that's not feasible, I can try to handle the difference with Capability detection

callummole:
@minrk `az aks get-upgrades` is showing available upgrades as `1.21.7` and `1.21.9`. I guess it is showing me only the most recent possible? I can try and force an upgrade to `1.19.X`?

minrk:
1.21 is fine, too! #2157 means we support 1.21

minrk:
I had the capability detection lying around in my git history for #2157, so #2162 means there's no time pressure now.

callummole:
@minrk I'm happy to upgrade it right now? It may be unavailable for a wee bit during the upgrade.

callummole:
Successfully upgraded to 1.21.9

PR #2168, opened 2022-05-06
Center the donate button

choldgraf:
This is a minimal CSS change to (I think) make the UI of the donate button a bit cleaner. It:

- Makes the container div `align-items: center`
- Makes the vertical margin `-8px` on top and bottom, so that it is centered within the div.

This makes the button stay centered without changing the banner height:

Old:

image

New: 

image


manics (approved):
Nicely done!

PR #2171, opened 2022-05-18
Deploy cryptnono

yuvipanda:
cryptnono (https://github.com/yuvipanda/cryptnono/) is a
helm chart that uses [bpftrace](https://github.com/iovisor/bpftrace)
to detect monero mining and kill all processes that mine monero
instantly. It is based off [this blog
post](https://blog.px.dev/detect-monero-miners/)
and installs a daemonset.

This is in addition to our minesweeper program, and focused
specifically on monero.

Ref https://github.com/jupyterhub/team-compass/issues/514

PR #2172, opened 2022-05-20
make cryptono optional

minrk:
disable it on turing, ovh (where it doesn't work yet, see https://github.com/jupyterhub/mybinder.org-deploy/pull/2171#issuecomment-1132838971)

### DIFF
--- a/config/ovh.yaml
+++ b/config/ovh.yaml
@@ -88,6 +88,8 @@ binderhub:
     imageGCThresholdLow: 30e9
     imageGCThresholdType: "absolute"
 
+cryptono:
+  enabled: false
 
 grafana:
   ingress:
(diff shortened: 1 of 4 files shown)


config/ovh.yaml
```diff
@@ -88,6 +88,8 @@ binderhub:
     imageGCThresholdLow: 30e9
     imageGCThresholdType: "absolute"
 
+cryptono:
```
sgibson91, line 91: I think this should be `cryptnono`? Seems to be a repeated typo

mybinder/Chart.yaml
```diff
@@ -56,3 +56,4 @@ dependencies:
   - name: cryptnono
     version: 0.0.1-n012.h3e298eb
     repository: https://yuvipanda.github.io/cryptnono/
+    condition: cryptono.enabled
```
sgibson91, line 59: Ignore previous comment. It is referencing this so all is good!

Issue #2183, opened 2022-05-27
Update cert-manager - almost two years outdated


consideRatio:
Maybe we have worked around/avoided the problems in https://github.com/jupyterhub/mybinder.org-deploy/issues/2051 already caused by using an old version of cert-manager. But overall, I think we should try get cert-manager updated anyhow as we are at 0.15.2, released .on Jul 1, 2020 while 1.8.0 is out now, released Apr 5, 2022.

https://github.com/jupyterhub/mybinder.org-deploy/blob/0cf568bd468c5dd5dbe5c29dfd394057cdd849e0/.github/workflows/cd.yml#L39-L46

Since cert-manager is installed without helm instead of as a dependency of the mybinder chart, this may actually possibly be as easy as updating the version and migrating configuration if needed. I'm not sure. Any helm chart with CRDs like cert-manager is a bit tricky to manage with `helm` as `helm` won't upgrade CRDs part of a Helm chart.

Looking at https://cert-manager.io/docs/release-notes/ and inspecting how we do things, we should be good to make a few upgrade steps at least, to 1.1 based on what I've read so far.

https://github.com/jupyterhub/mybinder.org-deploy/blob/3804b1046b6524e3ec70e6fcd6f6857405ee92ef/deploy.py#L248-L253

---

This issue relates to ingress-nginx upgrades, because cert-manager needs to be able to manipulate and handle Ingress resources. I think the right call is to merge #2214 before doing anything to cert-manager, but since cert-manager isn't critical for function short term, I suggest we finish all upgrade of ingress-nginx if possible without intermixing a cert-manager upgrade in between.

minrk:
closed by #2242

PR #2184, opened 2022-05-27
cd, cert-manager: use --repo during upgrade instead of helm repo add/update

consideRatio:
To use `helm repo add` and `helm repo update` is a more involved practice than just getting the helm chart repository when needed by using the `--repo` flag in `helm install|upgrade` commands.

I'm working towards resolving #2183 but tries to update misc parts along the way.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -253,27 +253,17 @@ def setup_certmanager():
     )
 
     print(BOLD + GREEN + f"Installing cert-manager {version}" + NC, flush=True)
-    subprocess.check_call(
-        ["helm", "repo", "add", "jetstack", "https://charts.jetstack.io"]
-    )
-
-    subprocess.check_call(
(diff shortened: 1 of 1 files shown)


deploy.py
```diff
@@ -253,27 +250,17 @@ def setup_certmanager():
...
     helm_upgrade = [
         "helm",
         "upgrade",
         "--install",
         "--create-namespace",
-        "--namespace",
+        "--namespace=cert-manager",
+        "--repo=https://charts.jetstack.io",
         "cert-manager",
         "cert-manager",
```
consideRatio, line 261: The first "cert-manager" is the release name (helm chart installation name), and the second "cert-manager" is the chart name to be found in the helm chart repo.

sgibson91 (approved):

PR #2224, opened 2022-05-28
Fully transition to Ingress with apiVersion networking.k8s.io/v1

consideRatio:
It seems the ingress-nginx controller we use can handle networking.k8s.io/v1/Ingress resources already, but I found that one OVH Ingress resource was not yet migrated to v1 but stuck at v1beta1.

This PR removes conditional logic no longer needed as we are a k8s v1.20+ systematically already.

---

This makes me think that we don't need to upgrade ingress-nginx in #2214, but can go straight for #2203. But, it won't hurt doing it step by step.


consideRatio:
We still haven't transitioned fully, because for example prometheus and cert-manager still interacts with old versions.

PR #2238, opened 2022-05-30
Update issue template unicode icon

choldgraf:
I noticed that the image template wasn't properly rendering the unicode when spelled out, so this just pastes in the emoji directly

### DIFF
--- a/.github/ISSUE_TEMPLATE/request_resources.yml
+++ b/.github/ISSUE_TEMPLATE/request_resources.yml
@@ -1,4 +1,4 @@
-name: \U0001F4C8 Request to bump resources for repo
+name: 📈 Request to bump resources for repo
 description: Answer some questions to help us make a decision about your request and set things up faster.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/request_resources.yml
```diff
@@ -99,4 +99,4 @@ body:
       label: Extra information
       description: Anything else you'd like to tell us?
     validations:
-      required: false
\ No newline at end of file
+      required: false
```
choldgraf, line 102:
```suggestion
      required: false

```

.github/ISSUE_TEMPLATE/request_resources.yml
```diff
@@ -99,4 +99,5 @@ body:
...
+
```
choldgraf, line 103:
```suggestion
      required: false
```

PR #2255, opened 2022-06-03
Add incident report

choldgraf:
This adds an incident report for an outage that we had yesterday, due to a GKE resource quotas bug.


docs/source/incident-reports/index.rst
```diff
@@ -25,27 +25,8 @@ Incident history
...
-   2018-02-12-launch-fail
-   2018-01-18-reddit-hug
-   2018-01-18-ssl-outdated
-   2018-01-17-ubuntu-upgrade
-   2018-01-04-failed-staging-deploy
-   2017-11-30-oom-proxy
-   2017-10-17-cluster-full
-   2017-09-29-504
-   2017-09-27-hub-403
+   ./2*.md
```
choldgraf, line 31:
```suggestion
   ./2*
```

PR #2262, opened 2022-06-09
let OVH back in

### LINKED ISSUE #2223: OVH TLS certificate expired 28th May
Visiting https://hub-binder.mybinder.ovh/hub/api gives you an expired certificate setup for `*.mybinder.ovh`. I expect that could cause some deployment health check issues.

@mael-le-gal is it you who I should ping about this?

minrk:
SSL issues resolved

closes #2223


minrk:
I guess it makes sense to try letsencrypt _before_ being added back to the federation: #2264

PR #2264, opened 2022-06-09
Use letsencrypt on OVH

### LINKED ISSUE #2223: OVH TLS certificate expired 28th May

minrk:
closes #2223 

makes redirector optional, since we don't need it on anything but prod


mybinder/templates/redirector/configmap.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.redirector.enabled }}
```
consideRatio, line 1:
```suggestion
{{- if .Values.redirector.enabled -}}
```

mybinder/templates/redirector/ingress.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.redirector.enabled }}
```
consideRatio, line 1:
```suggestion
{{- if .Values.redirector.enabled -}}
```

mybinder/templates/redirector/pdb.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.redirector.enabled }}
```
consideRatio, line 1:
```suggestion
{{- if .Values.redirector.enabled -}}
```

mybinder/templates/redirector/service.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.redirector.enabled }}
```
consideRatio, line 1:
```suggestion
{{- if .Values.redirector.enabled -}}
```

consideRatio:
Rules of thumb for whitespace chomping:
- always left
- never right, except when before content has been rendered in a yaml file or named template definition, then also always right

mybinder/templates/redirector/deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.redirector.enabled }}
```
consideRatio, line 1:
```suggestion
{{- if .Values.redirector.enabled -}}
```

consideRatio:
I think there is a dedicated ovh ingress resource standalone from helm chart that is kubectl apply-ed by the deployer script. Should it be deleted? Also, an entry in the deployer script for handling that file.

minrk:
@consideRatio indeed! Removed that as well, adding `ovh.mybinder.org` to the existing binder ingress

consideRatio (commented):
I think this looks good, not 100% confident in any way but +1 foe merge and try

minrk:
Since we got the new cert, I'm going to give this a go and if it doesn't seem to work, I'll just revert and stick with the new OVH-provided cert.

consideRatio:
Oh I hope it works :D

minrk:
It didn't work at first, but that's because our letsencrypt chart upgrade had some leftover resources preventing things from working. Some manual cleanup of cert-manager resources (deleting old certificaterequests) fixed it, and mybinder.ovh is happily secured by letsencrypt

PR #2265, opened 2022-06-09
rollout coredns

### LINKED ISSUE #2249: cert-manager-webhook intermittently reachable from turing's k8s api-server cause issues
NOTE: This issue was originally about killing all coredns pods, as that was believed to be causing an issue for the k8s api-server to reach the cert-manager-webhook.

---

@minrk while there are two coredns replicas and an rollingupdate is made by coredns, this logic bypasses it.

```python
            print("Reloading dns pods")
            kube.delete_collection_namespaced_pod(
                "kube-system",
                label_selector="k8s-app=kube-dns",
            )
```

We could either do...

```
kubectl rollout restart deploy coredns
```

Or we could make a label change on the pod spec of the coredns deployment to have a key=value with a new hash to make the deployment controller rollout a new version of the pods etc.

I think `kubectl rollout restart` may be to actually submit a patch request. See https://github.com/kubernetes-client/python/blob/251a1b3f86bc5e9aba17c5b764a0368167e6f73e/examples/dynamic-client/deployment_rolling_restart.py#L82-L95.

minrk:
When updating coredns, patch the deployment to trigger a rollout, rather than deleting pods

closes #2249

Since this is some not sensitive code in an encrypted file, the actual code is here:

```diff
@@ -178,9 +179,22 @@ def update_stub_dns(cluster=None):
         )
         if before.data != patched.data:
             print("Reloading dns pods")
-            kube.delete_collection_namespaced_pod(
-                "kube-system",
-                label_selector="k8s-app=kube-dns",
+            stub_hash = hashlib.sha1(stub_corefile.encode()).hexdigest()
+            app_client = kubernetes.client.AppsV1Api()
+            app_client.patch_namespaced_deployment(
+                name="coredns",
+                namespace="kube-system",
+                body={
+                    "spec": {
+                        "template": {
+                            "metadata": {
+                                "labels": {
+                                    "mybinder-stub-hash": stub_hash,
+                                }
+                            }
+                        }
+                    }
+                },
             )
     else:
         # kube-dns
```


consideRatio:
- Seems like the use case for an annotation rather than a label.
- maybe clarify the name a bit more? mybinder-stub doesnt help me get what this is, or what set it. To know it was the ban.py script, mybinder.org-deploy associated logic, or that its a hash og a configmap's content are all relevant to involve in the name i think but perhaps a bit much. Clarifying its related to a mounted configmaps content seems most relevant.

consideRatio:
This looks technically sound but i suggested some non-critical changes for code comprehension.

Labels vs annotations is mainly because a label indicates its used as a selector etc i figure, but annotations clarifies its only about data etc.

consideRatio:
Yepp anything changing the pod template of a deployment makes the pods rollout.

The name has meaning for someone inspecting the coredns resource and finding the annotation and becomes curious about it as well ;) i did that before i realized that it was restarted by the ban script and was confused why it was.

consideRatio (approved):

PR #2304, opened 2022-07-25
Redeploy Turing cluster to uksouth

callummole:
Actions:

- A new Turing cluster, named `turing-prod`, is deployed to azure region `uksouth` (see https://github.com/jupyterhub/team-compass/issues/529). It has the same specs as the previous one barring increase node capacity and an update k8s version (1.23.8).
- Binderhub deployed successfully (see https://binder.mybinder.turing.ac.uk/)
- Reinstated CI deployment. 
- Reinstated Turing in federation.


callummole:
do not merge yet - redeploying with updated k8s version

callummole:
Ok redeployed with 1.23.8, ready to merge at your discretion.

manics:
I've hit `Failed to connect to event stream` in every new build lasting more than a few minutes, though refreshing works. Is there a timeout somewhere that causes the event-stream to be cut off?

sgibson91:
@manics I was under the impression that `Failed to connect to event stream` is a browser-side issue (since as you say, refreshing works)?

sgibson91 (approved):
Suggested comment but LGTM!

manics:
@sgibson91 Not sure.... it's often a browser issue, but when it happens frequently with one server but not another it suggests there may be some other factor.

sgibson91:
Well I don't believe the config for Turing prod has changed at all. It was only torn down and redeployed in a new data centre, so maybe it's something to do with the availability of that data centre? `uksouth`

callummole:
Yes, I haven't changed any timeout values. 

@manics I'm not sure what to look for here, do you have any suggestions for me to investigate?

config/turing.yaml
```diff
@@ -133,7 +129,7 @@ prometheus:
 ingress-nginx:
   controller:
     service:
-      loadBalancerIP: 20.76.176.114
+      loadBalancerIP:
```
sgibson91, line 132: We should either update this field with the new IP address of the ingress-nginx controller, or remove the key completely

callummole: updated in [419ccd4](https://github.com/callummole/mybinder.org-deploy/commit/419ccd4b347292f987e45f551409cfe880dddea5)

PR #2313, opened 2022-08-02
Minor deploy refactoring; add `turing` release name to cd.

callummole:
Changed `deploy.py` so that `cluster = args.cluster or args.release` if no cluster name was given. This means you can pass just one argument to `deploy.py`, which is what happens for the CD staging deployment (`deploy.py staging`). 

Also added a turing `release_name` to the CD deployment. During redeployment of the Turing cluster (#2304 ) I didn't want to delete the old turing cluster before redeployment so I renamed the cluster to `turing-prod`, but the release name defaulted to `turing` at that time. 

Finally, I expanded the release options to include `turing-staging` and `turing-prod`, so that in the future we can keep the convention of `release == cluster` without renaming the cluster.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -201,6 +201,7 @@ jobs:
             hub_url: https://hub.mybinder.turing.ac.uk
             chartpress_args: ""
             helm_version: ""
+            release_name: turing
 
           - federation_member: ovh
             binder_url: https://ovh.mybinder.org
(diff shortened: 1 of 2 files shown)


minrk (approved):

manics:
This failed to deploy on Turing:
`HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"configmaps \"kube-dns\" not found","reason":"NotFound","details":{"name":"kube-dns","kind":"configmaps"},"code":404}`

sgibson91:
https://github.com/jupyterhub/mybinder.org-deploy/pull/2315 should fix that!

PR #2316, opened 2022-08-03
make cluster_name and release_name for turing explicit

callummole:
As per #2315 comments (https://github.com/jupyterhub/mybinder.org-deploy/pull/2315#issuecomment-1203930197), the cd was looking for the config file `turing-prod.yml`.  

Changed `cd.yml` so that the federation member remains `turing`, but the cluster_name and release_name are explicit, to allow for easy updating in the future.

PR #2323, opened 2022-08-08
move ban call to update_networkbans()

callummole:
As discussed in https://github.com/jupyterhub/mybinder.org-deploy/pull/2316, there was conflation of `cluster` and `release` within `deploy()`. `Ban.py` expects a `cluster` name, but was given a `release` name inside of `deploy()`.  

I have added the `update_networkbans()` function to separate the logic.

[
](https://github.com/callummole/mybinder.org-deploy/blob/b91ed40311404313a633a93b97c370cb268ef955/deploy.py#L116-L129)

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -113,29 +113,35 @@ def setup_auth_gcloud(release, cluster=None):
         ]
     )
 
+def update_networkbans(cluster):
+    """
+    Run secrets/ban.py to update network bans
+    """
+    
+    print(BOLD + GREEN + f"Updating network-bans for {cluster}" + NC, flush=True)
+
+    # some members have special logic in ban.py,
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

callummole:
Hmmm. It seems that at somepoint during the CD a 504 happens on `turing.mybinder.org`. 

It works after a local deploy (e.g. turing.mybinder.org is up right now after rerunning the deploy locally) then consistently becomes 504 if you rerun the CD. 

I don't have time to investigate right now.

callummole:
OK, it seems like sometimes CD does work. 

@minrk I hope it's just a timeout issue, and not the intermittency we were struggling with before (now there is no hub23 and also improved nodes, those were our two main suspicions).

PR #2339, opened 2022-09-06
remove invalid matrix.experimental key

minrk:
from #2338 

absence is already falsy, so removing it should be the same as it being present and false


.github/workflows/cd.yml
```diff
@@ -69,7 +69,6 @@ jobs:
       # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run
       continue-on-error: ${{ matrix.experimental }}
```
minrk, line 70: If it's needed, we can add `|| false` here

PR #2369, opened 2022-10-14
BinderHub 0.2.0-0.dev.git.2782.h3ccfc44 (most recent version using JupyterHub<2)

manics:
This is the most recent BinderHub chart using JupyterHub 1.* Subsequent versions use Z2JH 2.0.0 JupyterHub 3.0.0

---

https://github.com/jupyterhub/binderhub/compare/72e1852...3ccfc44


consideRatio (approved):

consideRatio:
Thanks! It would be good to test this by itself to avoid going for too many changes at once.

Issue #2407, opened 2022-11-10
new OVH cluster


minrk:
OVH is migrating us to a new OVH project, as a part of renewed funding via numfocus.

We'll have an actual cloud project, so we have full control of the cluster and anything else we need to deploy, instead of having a cluster provided to us, so we'll be able to do things like upgrade and recycle nodes. The new setup should also make the cluster less unique (I think we can use load balancers, letsencrypt, etc. unlike the current cluster), since OVH now has a managed kubernetes service. The project is owned by numfocus, as the direct recipient of the funding.

Things to do:

- [ ] get folks access to the project (I think folks have to create an OVHCloud account and send your user id (`rm111111-ovh`) to @SylvainCorlay as the numfocus representative)
- [ ] (figure out if we can get project admin permissions to add users, so @SylvainCorlay doesn't have to be the only one who can do it)
- [x] deploy container registry, cluster with terraform, like we do for GKE
- [x] add new cluster to federation as `ovh2`
- [ ] when things are going smoothly, remove `ovh` from federation so OVH can shut it down
- [ ] (later) after `ovh` is shutdown, migrate `ovh` domain(s) to new cluster

I'll have time to work on this starting next week when I finish with most of my teaching for the semester.

cc @mael-le-gal 

---

- EDIT by Erik: Closes #2176
- EDIT by Erik: Closes #2252

SylvainCorlay:
Regarding the admin access to the NumFOCUS institutional account, this was created with a dedicated email that could have the credentials made available in a password manager. For most operations (besides adding and removing maintainers), being an admin on the *project* should be the only requirement.

minrk:
I've started testing out deployment with terraform in #2414, and so far it seems to be going fine. I've hit some hiccups with failed deployments without any error messages (specifically, deploying the registry only got 'state=ERROR' with no messages), but retrying a couple of times and it went okay.

manics:
I'm happy to help/look, could I have access please?

minrk:
@manics You have to [create an account](https://www.ovh.com/auth/?onsuccess=https%3A//www.ovh.com/manager&ovhSubsidiary=IE), then send the user id (`rm111111-ovh`) to Sylvain, and he can add you to the project.

consideRatio:
@minrk I've deployed prometheus 16.0.0 now, but not on ovh2 or similar.

[...]

PR #2414, opened 2022-11-14
new OVH cluster

minrk:
implementing #2407

So far:

- [x] cluster is deployed with terraform
- [x] registry is deployed with terraform, along with accounts for credentials
- [x] binderhub is deployed to the new cluster (currently at https://ovh-test.mybinder.org)
- [x] added new cluster to CI scripts as `ovh2`

So far it's simpler and less unique than the previous OVH cluster, since we are using standard credentials, letsencrypt, a load-balancer service, etc. Nothing is manually created and no nodes need special treatment.

Stil to do:

- [x] figure out the right node flavors (after quotas are worked out)
- [x] select data center
- [x] put the cluster on a private network?
- [x] assign actual domain name, once we are confident we've deployed the cluster for the last time.
- [x] put docker credentials in github repo secrets

Questions for @mael-le-gal and OVH folks, especially since I'm new to the OVH tools:

- data centers - do you have any suggestions for region in which to run the cluster and/or registry? I've picked GRA/GRA9 for now.
- tokens for OVH - I created credentials following the steps [in the ovh docs](https://docs.ovh.com/sg/en/kubernetes/creating-a-cluster-through-terraform/) to use in terraform, but it's quite unclear what permissions those credentials actually have. It seems like the token has permission to act _as me_, which is not what we want. Presumably, I should create credentials that only have access to this particular project, but I can't find any way to manage the token credentials once they've been created. Where do these tokens/access get managed? Is there an example of creating a 'service account' with appropriately limited access to only the project it is meant to control?
- are there any examples of using OVH object storage for terraform state? We use `backend = gcs` for google, and I've done the same with the OVH deploy, but it would probably be nice to store it in the OVH object store. I'm guessing the s3 backend can be configured with the right options, I just don't know what they are.
- the quotas on the account are currently very low. It's asking me for billing info, but I shouldn't need any billing info on my account to manage the _project_. Does this mean only @SylvainCorlay can do this? If so, @SylvainCorlay can you order the 20VM quota? We may need 50, but let's start with 20 for now. We can't really deploy a toy instance with the base limit, since it doesn't have enough RAM for a single cluster node.
- The version of Harbor in the managed service is quite old (2.0.1 from 2020, current is 2.6), which causes some problems with the harbor terraform provider, which doesn't support harbor older than 2.2. What is the chance of harbor getting upgraded? I've worked around it since I think this will take time, but it would be nice to have a reasonably up-to-date registry

Thanks!


mael-le-gal:
Hello @minrk 

Here are some answers for yours questions :

### About the region :
Your previous cluster was hosted in GRA region (which stands for Graveline) so I suggest to select the same.

### About the ovh token : 
They are all related to our API. 
The description of the API can be found here  : https://api.ovh.com/console/ 
In your case the routes that you need will mainly be under the `/cloud/project/{serviceName}/*` section. If you replace **serviceName** by your project id and you create your token with that route configured, then your token will be scoped only on your project.
If later you want to manage your created tokens you can do it by calling the routes starting with [/me/api/credential](https://api.ovh.com/console/#/me/api/credential~GET). This current system is not intuitive, sadly I agree on that ...

### About the s3 provider :
Maybe [this can help you](https://docs.ovh.com/gb/en/public-cloud/use_high_perf_object_storage_terraform_backend_state/)

Before to do that I guess you will have to  [created your s3 user](https://docs.ovh.com/fr/storage/object-storage/s3/identity-and-access-management/) and [created your bucket][https://docs.ovh.com/fr/storage/object-storage/s3/getting-started-with-object-storage/]

mael-le-gal:
### About the registry
I got an answer from the registry team :  
They are aware that the registry version is an old one.
The Harbor version 2.4.1 should be available soon (December)
They are working on improving the speed of their delivering process and next year they should provide new version but faster (2.5 / 2.6)

minrk:
Thanks! I was able to migrate state to s3 on OVH.

THe cluster is now deployed on a private network, which seems right.

The node flavors don't quite line up how we're used to. We typically use GCP's 'highmem' nodes which have a ~7:1 ram:cpu ratio, whereas OVH offers ~4:1 or ~12:1, but not in between. We have used 4 cpu nodes for the core pool, and 8 cpu nodes for the user pool.

I suggest we start with node pools:

- b2-15 (4 cpu, 15GB) for core, and
- r2-60 (4 cpu, 60GB) for user nodes

Given the ratios in OVH quotas (~6 GB / core), we could also go with b2-60 (16 cpu, 60GB) for user nodes, and have ~twice the cpu headroom per user that we do on GKE.

I also don't know if there's anything special we want to do with disks for image building. on GKE, we mount SSDs for this, but I think perhaps the boot disk SSDs are fine? At least okay for now.

So if we're confident in the cluster setup so far, I can update the domain and merge to try to start deploying on CI.

minrk:
@mael-le-gal where can we see logs, e.g. kubernetes cluster event logs? I tried clicking on the 'Logs Data Platform' in the cloud project sidebar, but then it tried to get me to buy it on my personal account. How do we associate logs with an existing cloud project that has the credits, etc.?

mael-le-gal:
> @mael-le-gal where can we see logs, e.g. kubernetes cluster event logs? I tried clicking on the 'Logs Data Platform' in the cloud project sidebar, but then it tried to get me to buy it on my personal account. How do we associate logs with an existing cloud project that has the credits, etc.?

Currently I think you can't, except from the control panel UI directly.

More info [here](https://github.com/ovh/public-cloud-roadmap/issues/51)

The global OVHcloud offer is currently missing an integrated observability solution so that all logs from all services ends in the same place for customers. We are aware of that limitation and that's on the long term roadmap.

minrk:
@mael-le-gal can I associate a cloud public IP with a kubernetes load balancer? We do that in other deployments so that if the public kubernetes Service gets recreated, the IP doesn't get released. As far as I can tell, though, I can only associate public IPs with instances, not load balancers.

minrk:
OK! I think this is ready to go to the next step. The cluster is deployed, and I've deployed it manually with `python deploy.py ovh2`. I believe I've made all the changes necessary for the cluster to get updated from CI here as well, but the only way to know is to merge and try it.

Note that this will _not_ make ovh2 part of the federation. That's a separate step. This will just get the new cluster deployed from CI instead of my laptop, to ensure everything really is working.

mybinder/templates/image-pull-secret.yaml
```diff
@@ -0,0 +1,30 @@
...
+      "password": {{ .password | quote }},
+      {{- if .email }}
+      "email": {{ .email | quote }},
+      {{- end }}
+      "auth": {{ (print .username ":" .password) | b64enc | quote }}
+    }
+  }
+}
+{{- end }}
+{{- end }}
```
consideRatio, line 20: Is this perhaps already something defined in the z2jh helm chart? Binderhub's pods can via BinderHub's helm chart now rely on z2jh helm chart's created image pull secret resource as well. This is https://github.com/jupyterhub/binderhub/pull/1444.

Maybe it could make sense to pass values into z2jh's imagePullSecret configuration directly and reference that in mybinder.org-deploy pods to avoid adding duplicated logic like this?

Another option is to reference the globally available named template function directly as defined in z2jh: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/391e37d0376e8c10e931c4207a4fd335f34f3f67/jupyterhub/templates/_helpers.tpl#L147-L176

You can do things like `{{ include "jupyterhub.dockerconfigjson" . }}`, it is available for a chart depending on some chart where its defined. That is why its important for people to put a chart name first when defining these template functions.

consideRatio (commented):
Looked good to me overall! I just had some ideas related to `imagePullSecret`

mybinder/templates/image-pull-secret.yaml
```diff
@@ -0,0 +1,30 @@
+{{- define "_dockerconfigjson" -}}
+{{ include "_dockerconfigjson.yaml" . | b64enc }}
+{{- end }}
```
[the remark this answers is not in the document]
consideRatio: Similar feedback comment in another PR: https://github.com/jupyterhub/binderhub/pull/1577#discussion_r1025522645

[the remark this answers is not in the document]
minrk: Interesting! This is verbatim from z2jh

mybinder/templates/image-pull-secret.yaml
minrk: It's the same pattern, but this a different secret, applied on different pods, and z2jh can only create _one_ secret, not two. We use both because the mybinder chart images come from a different source with different credentials.

But I think I realized there's a problem with how I do that, and we can get rid of it here. Because when harbor is updated to 2.4, we won't need to use a separate secret for this.

minrk: I removed this whole imagePullSecret bit, as I realize it wasn't needed.

deploy.py
```diff
@@ -245,13 +245,43 @@ def setup_certmanager():
...
+            "kubectl",
+            "set",
+            "resources",
+            "-n",
+            "kube-system",
+            "deployments/coredns",
+            "--limits",
+            "memory=250Mi",
+            "--requests",
+            "memory=200Mi",
```
minrk, line 265: @mael-le-gal coredns was getting OOMKilled after we applied our coredns config to ban a bunch of IPs, so we have to raise this a bit. I'm not sure if that's feedback that would be useful to you.

minrk:
@consideRatio thanks for the review! Going ahead since the pull secret that had some review was entirely unnecessary and is now removed.

PR #2425, opened 2022-11-21
add OVH2 to federation

minrk:
#2414 seems to be working fine.

This adds OVH2 to the cluster with a capacity of only 10, to see how it goes. Can slowly increase capacity over the week.


manics (approved):
Awesome!

PR #2433, opened 2022-11-22
federation redirect: record redirect reason

minrk:
- tracks how often cookie is responsible vs load balancer, which I'd like to know
- limits cookie age to configurable number of hours (default: 4). Previously, once a user landed on an instance (~always GKE eventually), they would be stuck there ~forever.

Background: GKE is still getting far more traffic than the load-balancer input would suggest. Either this means the load-balancer doesn't work as intended, or the cookie is having an outsized effect because it doesn't expire. But right now, we can't tell _why_ a redirect target is getting traffic (cookie or load balancer).

### DIFF
--- a/images/federation-redirect/app.py
+++ b/images/federation-redirect/app.py
@@ -1,4 +1,5 @@
 import asyncio
+import datetime
 import json
 import math
 import os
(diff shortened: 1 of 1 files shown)


betatim:
Sounds like a good idea. Both finding out and diagnosing that eventually all users end up on the GKE instance.

PR #2440, opened 2022-11-23
switch ovh user nodes to r2-120

minrk:
more cpu (8), memory (120) and also more disk (200GB). The previous node ran out of disk and wasn't able to cleanup quickly enough.

May need to investigate attaching an external SSD volume for dind.


minrk:
New pool was added first, then old pool drained, then old pool deleted after draining.

Issue #2449, opened 2022-11-28
New federation member on AWS backed by Pangeo


sgibson91:
As discussed in https://github.com/jupyterhub/team-compass/issues/501, Pangeo want to simplify their Binder infrastructure by removing dask-gateway. This effectively makes it an "ordinary" binder and, therefore, a great candidate as a federation member. Especially since this will be deployed to AWS, which we don't currently cover. Since 2i2c are working with Pangeo (via a grant held at Columbia) to operationalise their cloud infrastructure, I will be taking the lead on the technical aspect of deployment, and this issue will be tracking that.

### Things to do

- [ ] Setup new terraform config for AWS
  - I will move the existing config to a `terraform/gcp` folder, and then create a `terraform/aws` folder for the new config
- [ ] Deploy the cluster to an AWS account
- [ ] Create helm config for a new binder deployment
- [ ] Deploy a new Binder instance to the AWS cluster
- [ ] Add the new cluster to a mybinder.org sub-domain
- [ ] Add new cluster/binder to deploy.py and CI/CD
- [ ] Add the new cluster to the federation

sgibson91:
Thanks @yuvipanda! I think I had a hybrid idea between (1) and (2) which was to use the official aws provider for as much as possible, and then the vpc network. But perhaps that's unnecessary.

manics:
> An IAM user that can be used by CI/CD to authenticate to the cluster for automated deployment

Can you use an IAM role and GitHub OIDC instead? This avoids needing hard-coded secrets, you'll get a temporary token every time.
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
It's been working well for me when deploying infrastructure on AWS.

manics:
Also https://github.com/jupyterhub/binderhub/pull/1055 (Add AWS ECR support (round two)) is probably relevant, I was going to review it but never had time 😞. It includes IAM instructions.

sgibson91:
> Can you use an IAM role and GitHub OIDC instead?

Let's keep this as a stretch goal. I'm learning and so want to focus on the MVP for now. This feels like something that can be switched to later on after we've got the basic infrastructure setup first.

> Also [jupyterhub/binderhub#1055](https://github.com/jupyterhub/binderhub/pull/1055) (Add AWS ECR support (round two)) is probably relevant, I was going to review it but never had time 😞. It includes IAM instructions.

Thank you

[...]

PR #2467, opened 2022-12-15
[WIP] Add terraform config for an AWS cluster

sgibson91:
related to #2449 

Following https://github.com/hashicorp/learn-terraform-provision-eks-cluster and other bits and pieces, this is the terraform config I've come up with so far to deploy a cluster and supporting infrastructure on AWS.

Currently I get the following error when deploying this:

```bash
╷
│ Error: failed creating IAM Role (pangeo-mybinder-k8s-ecr-iam-role): MalformedPolicyDocument: Has prohibited field Resource
│ 	status code: 400, request id: 6e4da11d-7d10-4d2c-84c6-72cf624bca4d
│
│   with aws_iam_role.k8s_ecr_iam_role,
│   on main.tf line 140, in resource "aws_iam_role" "k8s_ecr_iam_role":
│  140: resource "aws_iam_role" "k8s_ecr_iam_role" {
│
╵
```

The policy in question was retrieved from this section of some as-yet-unmerged documentation, so it could be that it's just not correct to be using the Resource attribute here:

- https://github.com/jupyterhub/binderhub/pull/1055/files#diff-12fbfe728e3edf4301ff30644079395e48e3eeb466ac9bc6cedad95c830a3fe6R185-R233


sgibson91:
The two READMEs look like I've edited more than I did. What I did was:

- Move the original `terraform/README.md` to `terraform/gcp/README.md` and changed the title to indicate that it refers to GKE only
- Create a new `terraform/README.md` file explaining the need for different folders/config for different cloud providers

terraform/aws/pangeo/main.tf
```diff
@@ -0,0 +1,229 @@
...
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        "Action": [
+          "ecr:ListImages",
+          "sts:AssumeRole"
+        ],
+        "Effect": "Allow",
+        "Resource": "${aws_ecr_repository.image_repo.arn}*",
```
yuvipanda, line 156: Temporarily, does getting rid of all the "Resource" fields or converting their values to "*" make the config apply?

manics: This is an assume role role policy so it say which entities (e.g. users, or other AWS services) are allowed to assume the role. The privileges for the role are specified separately, either using [`inline_policy`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role#inline_policy), or in a separate [`aws_iam_policy`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy) linked to the role using [`aws_iam_policy_attachment`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy_attachment).

If you want to use an IRSA role which you can link directly to a K8s service account ... it's more complicated as you need an K8s OIDC provider to act as the entity that assumes the role. I've got an example somewhere which I can dig out if you want?

sgibson91: - Setting all "Resource" values to "*" produces the same error
- Got a different error by removing all instances of "Resoource"

```bash
│ Error: failed creating IAM Role (pangeo-mybinder-k8s-ecr-iam-role): MalformedPolicyDocument: Missing required field Principal
│ 	status code: 400, request id: be65c496-2b4c-4dbf-8676-8a2b8e29bf5e
│
│   with aws_iam_role.k8s_ecr_iam_role,
│   on main.tf line 140, in resource "aws_iam_role" "k8s_ecr_iam_role":
│  140: resource "aws_iam_role" "k8s_ecr_iam_role" {
│
```

sgibson91: Sure, I don't know how to move this forward otherwise. Though I won't be doing any more work here until the new year.

sgibson91:
@consideRatio and I debugged this a little today and ran up against a problem with ECR.

ECR as a container registry, compared to GCR/DockerHub/quay.io, requires the creation of a image repository before we push to it. This goes against the binderhub software's assumptions that you can push directly to a new repository name to define it.

This is [documented by AWS](https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html) as highlighted below:

> 

Instead, we intend to use quay.io. Other federation members, such as Turing and GESIS, have used dockerhub in the past, but we have started using quay.io after dockerhub tightened their free tiers a bit. Both services are free and could provide us with a container registry. (We also did some digging and found that the original Pangeo AWS BinderHub was using dockerhub too, not ECR.) We have created the mybinder-org organisation on quay.io and added some of the JupyterHub team members over there to store the images BinderHub produces.

This decision unblocks this PR since we no longer need to create a role that can push to/pull from an ECR. We will instead create a robot account on quay.io and provide the username and password in encrypted config files read by the BinderHub helm chart (in a follow-up PR).

minrk:
OVH has been having quite a bit of trouble with the registry because OVH's private container registry has a very small size limit and seems to have lots of performance problems, so I think perhaps oVH should try quay.io as well, and see how it goes.

sgibson91:
I have just pushed the following commit which set up cluster-autoscaler as an optional dependency of the helm chart (because this is another thing we don't get out of the box with EKS)

- [`a31b959` (#2467)](https://github.com/jupyterhub/mybinder.org-deploy/pull/2467/commits/a31b959201dbaedb6dfe412d2a03827ce4c9bcfc)
- [`5cf922c` (#2467)](https://github.com/jupyterhub/mybinder.org-deploy/pull/2467/commits/5cf922c5888bd8ac3f188fcb480dafcec47917ce)

I would've preferred to do this in a new PR, as this is starting to move out of terraform set-up and into helm chart config. But thought having the PR reflect the most recent status was more important.

sgibson91:
In the interest of getting unblocked, I've marked this as ready for review

terraform/aws/pangeo/main.tf
```diff
@@ -0,0 +1,163 @@
...
+
+  private_subnet_tags = {}
+}
+
+# Setup security groups for core and user node groups
+resource "aws_security_group" "node_group_core" {
+  name_prefix = "node_group_core"
+  vpc_id      = module.vpc.vpc_id
+
+  ingress {
```
yuvipanda, line 51: I think this is no longer necessary, as we can use AWS SSM to ssh into these nodes: https://infrastructure.2i2c.org/en/latest/howto/troubleshoot/ssh.html#aws.

terraform/aws/pangeo/main.tf
```diff
@@ -0,0 +1,163 @@
...
+      "10.0.0.0/8",
+    ]
+  }
+}
+
+resource "aws_security_group" "node_group_user" {
+  name_prefix = "node_group_user"
+  vpc_id      = module.vpc.vpc_id
+
+  ingress {
```
yuvipanda, line 66: I think this is no longer necessary, as we can use AWS SSM to ssh into these nodes: https://infrastructure.2i2c.org/en/latest/howto/troubleshoot/ssh.html#aws.

terraform/aws/pangeo/main.tf
```diff
@@ -0,0 +1,163 @@
...
+
+# Setup EKS cluster with control plane and two managed node groups
+module "eks" {
+  source       = "terraform-aws-modules/eks/aws"
+  version      = "~> 18.31"
+  cluster_name = local.cluster_name
+
+  # k8s v1.24 allows us to scale managed node groups to zero
+  # ref: https://github.com/aws/containers-roadmap/issues/724
+  cluster_version = "1.24"
```
yuvipanda, line 85:
```suggestion
  cluster_version = "1.25"
```

New version is out!

terraform/aws/pangeo/main.tf
```diff
@@ -0,0 +1,163 @@
...
+
+  vpc_id     = module.vpc.vpc_id
+  subnet_ids = module.vpc.public_subnets
+
+  eks_managed_node_group_defaults = {
+    # Disabling and using externally provided security groups
+    create_security_group                 = false
```
yuvipanda, line 92: Can you help me understand why we are managing security groups ourselves instead of having the module manage them for us?

yuvipanda (commented):
Awesome work! Left some comments :)

Does terraform apply run cleanly now?

terraform/aws/pangeo/main.tf
sgibson91: I think to make certain that they were explicitly connected to the VPC I'd created by parsing the ID. There's nowhere else in the terraform config that says "attach this eks cluster to this vpc" so I guess I'm uncertain if that happens automatically and figured explicit is better than implicit.

terraform/aws/pangeo/main.tf
sgibson91: I think to make certain that they were explicitly connected to the VPC I'd created by passing the ID. There's nowhere else in the terraform config for the eks module that says "attach this eks cluster to this vpc" so I guess I'm uncertain if that happens automatically and figured explicit is better than implicit.

terraform/aws/pangeo/main.tf
sgibson91: Done in [`d6cbf1b` (#2467)](https://github.com/jupyterhub/mybinder.org-deploy/pull/2467/commits/d6cbf1bf0e6982032fcfc025da7146a731e15186)

terraform/aws/pangeo/main.tf
sgibson91: Done in [`d6cbf1b` (#2467)](https://github.com/jupyterhub/mybinder.org-deploy/pull/2467/commits/d6cbf1bf0e6982032fcfc025da7146a731e15186)

sgibson91:
> Does terraform apply run cleanly now?

@yuvipanda Yes, but that's because we dropped the problematic config. Instead favouring a quay.io registry instead of an AWS ECR, c.f. https://github.com/jupyterhub/mybinder.org-deploy/pull/2467#issuecomment-1405338819 If there's no ECR to push to, then there's no need to create the role that was causing terraform to not apply cleanly.

yuvipanda:
Ah that's awesome! Using quay seems fine to me.

sgibson91:
**Update:** I just clocked the phrase "credits have depleted" in this Discourse post about taking down the Pangeo AWS deployments https://discourse.pangeo.io/t/aws-pangeo-jupyterhubs-to-shut-down-friday-march-17/3228

So maybe we need to find somewhere else to test this?

manics:
I've suggested some steps in https://github.com/jupyterhub/mybinder.org-deploy/issues/2556#issuecomment-1514458702

sgibson91:
To be clear, this PR has been tested. Yuvi and I both have/had access to Scott's AWS account for the Pangeo Binder deployment and this terraform code applies cleanly over there. The problem is that we can't _continue_ to use that AWS account to develop and test now the credits have depleted.

So _for this PR specifically_, I think I would like it merged or somehow resolved relatively soon so that 1) it doesn't need to continually be rebased, as it touches other terraform config to move them into subfolders, and 2) when we get access to another AWS account, someone can open a new PR and iterate as needed, without having to manage my PR as well. And I say someone because I don't believe I have the capacity to technically lead this.

manics:
@sgibson91 Thanks for clarifying! I couldn't tell from all the comments what the state of this was. I don't feel able to review this without access to a deployment, but I'm happy for @yuvipanda to make the decision, and I can take on the follow-up work when we have AWS access.

yuvipanda:
@manics we can give you access!

manics:
@yuvipanda Thanks, but to avoid delaying this further, if you've already reviewed this I think we should merge it when you're happy with it.

yuvipanda:
Great, done! We can iterate on this :)

Thanks for working on this, @sgibson91!

PR #2469, opened 2022-12-16
bump ovh2 out of rotation

minrk:
seems to be having network problems, and credential problems:

> unauthorized: unauthorized to access repository: mybinder-builds/r2d-g5b5b759binder-2dexamples-2drequirements-55ab5c, action: push: unauthorized to access repository: mybinder-builds/r2d-g5b5b759binder-2dexamples-2drequirements-55ab5c, action: push

Did something change in how repo2docker's credentials are passed?

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -621,7 +621,7 @@ federationRedirect:
       versions: https://ovh.mybinder.org/versions
     ovh2:
       url: https://ovh2.mybinder.org
-      weight: 100
+      weight: 0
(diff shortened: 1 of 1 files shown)


minrk:
> Did something change in how repo2docker's credentials are passed?

Nope, robot puller account expired after 30 days!

Screenshot 2022-12-16 at 20 27 07

minrk:
So I know what to fix, but I think we should still bump it out while it can't build.

PR #2503, opened 2023-02-08
get-prs.py: Ignore merge commits

manics:
Haven't tested this yet

Chartpress won't include the most recent commit in the Helm chart hash because it's a merge commit which doesn't change anything. This means the merge commit from the previous chart update will instead be associated with the current chart update.

### DIFF
--- a/scripts/get-prs.py
+++ b/scripts/get-prs.py
@@ -58,7 +58,9 @@ if len(commits) > args.max_commits:
     ]
 else:
     for c in commits:
-        prs.update(c.get_pulls())
+        if len(c.parents() == 1):
+            # Chartpress ignores merge commits when generating the Helm chart SHA
+            prs.update(c.get_pulls())
     pr_summaries = [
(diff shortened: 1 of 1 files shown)


scripts/get-prs.py
```diff
@@ -58,7 +58,9 @@ def extract_gitref(s):
     ]
 else:
     for c in commits:
-        prs.update(c.get_pulls())
+        if len(c.parents() == 1):
```
minrk, line 61: Looks like it should be:

```suggestion
        if len(c.parents == 1):
```

```
  File "/Users/minrk/dev/jpy/hub/mybinder.org-deploy/scripts/get-prs.py", line 61, in <module>
    if len(c.parents() == 1):
TypeError: 'list' object is not callable
```

manics: It shouldn't matter. The script lists all commits between the two SHAs, and finds the PR associated with each of those commits. There should always be a non-merge commit alongside the merge commit, so the PR should still be included

PR #2517, opened 2023-02-28
Update repo2docker (Python 3.10 by default) and BinderHub (switch to jupyter-server)

minrk:
combines bot updates #2509 and #2515

closes #2509 
closes #2515

### DIFF
--- a/mybinder/Chart.yaml
+++ b/mybinder/Chart.yaml
@@ -8,7 +8,7 @@ dependencies:
   # Source code:    https://github.com/jupyterhub/binderhub/tree/main/helm-chart
   # App changelog:  https://github.com/jupyterhub/binderhub/blob/main/CHANGES.md
   - name: binderhub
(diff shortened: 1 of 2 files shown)


manics:
We also need to modify
https://github.com/jupyterhub/mybinder.org-deploy/blob/2b9d60d5ed6fae93d4778332da720f0989df6fa9/mybinder/values.yaml#L370

manics:
On the other hand this means we should be able to merge https://github.com/jupyterhub/mybinder.org-deploy/pull/2515 as soon as the bot picks up https://github.com/jupyterhub/binderhub/pull/1635 since it has no effect here!

minrk:
I think that was a temporary workaround for testing, which should have been removed here when the BinderHub changes solidified. It's now removed here, so it's not set in two places.

minrk:
https://dataverse.ucla.edu/ appears to be entirely down, causing the one test failure.

manics (approved):

PR #2528, opened 2023-03-15
Remove Turing config and references from repo

### LINKED ISSUE #2524: Remove Turing from the federation
I just received an email from Docker Hub stating that they are sunsetting Free Team organisations, which is the subscription type of the account that the Turing federation member is using. https://web.docker.com/rs/790-SSB-375/images/privatereposfaq.pdf (There is also no option to transform the org into a regular user either, apparently.)

However, we've had discussions about removing the Turing federation member completely. I believe it is currently out of rotation due to some bugs, and the folk that were maintaining it at the Turing Institute have moved onto other things. So I'm taking this as a sign that we should clear out any Turing-related config from the repo.

sgibson91:
fixes #2524 

I think possibly terraform may need to be run and `scripts/grafana-export` to update the grafana dashboards?


deploy.py
```diff
@@ -29,46 +29,6 @@
     "prod": "us-central1",
 }
 
-# Mapping of cluster names (keys) to resource group names (values) for Azure deployments
-AZURE_RGs = {"turing-prod": "binder-prod", "turing-staging": "binder-staging"}
-
-
-def setup_auth_turing(cluster):
```
minrk, line 36: Maybe keep this unused as `setup_auth_azure`? We may have new azure clusters in the future.

minrk (approved):

deploy.py
sgibson91: Fixed in [`1a15755` (#2528)](https://github.com/jupyterhub/mybinder.org-deploy/pull/2528/commits/1a1575517ec18eda91de995377f1fab254fcbc9c)

minrk (approved):

sgibson91:
I applied terraform for prod and staging to revoke the events archiver keys

For the grafana dashboard export:

1) the api key stored in `secrets/grafana-api-key` throws an error when used with `scripts/grafana-export`

```
  File "scripts/grafana-export", line 154, in <module>
    opts.action(opts.host, opts.data_dir, confirm=(not opts.no_confirm))
  File "scripts/grafana-export", line 46, in export_grafana
    r.raise_for_status()
  File "/usr/local/Caskroom/miniconda/base/lib/python3.8/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://grafana.mybinder.org/api/search?limit=1000
```

2) there's is a discrepency: `secrets/grafana-api-key` exports to `GRAFANA_API_KEY` but `scripts/grafana-export` looks for `GRAFANA_PASSWORD`

However, this is only a backup and not critical, so I will just merge this and we can address grafana stuff later

PR #2542, opened 2023-03-30
add script to prune empty repos on harbor (ovh)

minrk:
harbor has image retention rules that lets it prune unused images (yay!), but it doesn't prune the empty _repositories_, left behind, which seems to be a source of problems.

This adds a script and a workflow to run the script weekly to delete any repos with zero artifacts in them.

terraform changes:

- remove unused ovh harbor credentials
- add credentials for repo culler script
- run retention policy on saturday instead of sunday to avoid running retention policy and garbage collection at the same time, since they seem to conflict


scripts/prune_harbor.py
```diff
@@ -0,0 +1,128 @@
...
+        if not prune_progress.total and "X-Total-Count" in r.headers:
+            prune_progress.total = fetch_progress.total = int(
+                r.headers["X-Total-Count"]
+            )
+
+        repos: list[dict] = r.json()
+        if not repos:
+            break
+        fetch_progress.update(len(repos))
+        r.raise_for_status()
```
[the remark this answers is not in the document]
minrk: yes, good catch!

scripts/prune_harbor.py
```diff
@@ -0,0 +1,128 @@
...
+                    harbor_url + f"/projects/{project_name}/repositories/{repo_name}",
+                    auth=(username, password),
+                )
+                r.raise_for_status()
+                prune_progress.update(1)
+                page_deleted += 1
+        # if we deleted less than 90% of the images, request the next page
+        # i.e. stay on the current page if we delete at least 50% of the repos
+        # so we don't skip over them
+        # the alternative is to fetch _all_ repos
```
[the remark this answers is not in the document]
minrk: Yeah, I was trying to think of a way to do it, and paging backwards should work. Good idea!

.github/workflows/prune-harbor.yaml
```diff
@@ -0,0 +1,44 @@
...
+        # --upgrade is important since we carry a cache and may risk grow very
+        # outdated otherwise.
+        run: |
+          pip install --upgrade setuptools pip
+          pip install --upgrade requests ruamel.yaml tqdm
+
+      - name: "Stage 2: Unlock git-crypt secrets"
+        uses: sliteteam/github-action-git-crypt-unlock@8b1fa3ccc81e322c5c45fbab261eee46513fd3f8
+        env:
+          GIT_CRYPT_KEY: ${{ secrets.GIT_CRYPT_KEY }}
```
consideRatio, line 40: Weird to call this "Stage 2:" without a mention of stage 1 or similar.

scripts/prune_harbor.py
```diff
@@ -0,0 +1,129 @@
...
+    harbor_url: str,
+    project_name: str,
+    username: str,
+    password: str,
+) -> None:
+    """Deletes all repositories with no activity older than 30 days
+
+    These are often going to be repos with no content
+    """
+    print("Deleting repositories with no images")
```
consideRatio, line 45: Repositories with no activity seems different than repositories with no images, what's actually done? Looking at the code below I see `if repo["artifact_count"] == 0:` so I figure the docstring is outdated - it should be that repos with no images are deleted.

consideRatio (approved):
Small details in comments, but overall this LGTM and can be merged!

.github/workflows/prune-harbor.yaml
minrk: Yes, copied and pasted. This is why I don't think numbers belong in step names.

scripts/prune_harbor.py
minrk: Good catch! This script originally did a similar thing to #2012, but not anymore now that GC appears to be working. Fixed.

PR #2558, opened 2023-04-19
make binderhub itself optional

minrk:
So we can deploy the redirector, etc. separately in their own cluster, without a federation member at all.


minrk:
seems to work. Need to update CI to not attempt launching on clusters that don't deploy a BinderHub.

consideRatio (approved):
LGTM! I'll leave it to you to self-merge @minrk

Issue #2438, opened 2022-11-22
GCP cloud storage price increase


minrk:
GCP cloud storage prices [increased](https://cloud.google.com/storage/pricing-announce) in October, and we have our first bill. The _increase_ in costs came to about $800 between September and October, ~all from new charges for "Networking Traffic Egress GCP Multi-region within Northern America" ($625) and "Networking Traffic Egress GCP Replication within Northern America" ($162), both of which were zero-rated prior to October 1.

I believe we could save (most of?) these costs by switching to Artifact Registry and using a regional registry in us-central1 (instead of multi-region). Not sure if that's worth it right now, but something to keep an eye on.

minrk:
These costs have increased to $1100 out of $1500 total storage costs (only , amounting to fully 20% of the operating cost of gke.mybinder.org (less than $400 for actual at-rest storage).

<img width="573" alt="Screenshot 2023-04-19 at 10 28 05" src="https://user-images.githubusercontent.com/151929/233016449-8a6b49c1-c39c-478f-b3f1-3f24f87b37f7.png">

It's still unclear to me if these numbers would go all the way to zero or merely be reduced if we used a regional artifact registry, but I _think_ the egress numbers would go to zero, and the storage costs would drop by 20% (from .026/GB to .02/GB), reducing our storage costs from $1500 to ~$300.

PR #2562, opened 2023-04-21
Reduce cost of staging

minrk:
- set lower resource requests for several components (staging is almost always idle, so it doesn't need much). Needed to move some hardcoded resources from templates to values.yaml
- switch to cheap e2-standard-2 nodes (2 core, 8GB RAM) (edit: e2-medium was _too_ cheap, with only 1 allocatable CPU per node, which mostly got eaten up by kube-system.
- smaller boot disks (100GB)
- deploy artifact registry (cheaper - solves #2438) and will make it easier to delete _all_ images in the old Container Registry used by prod)

### DIFF
--- a/config/staging.yaml
+++ b/config/staging.yaml
@@ -6,7 +6,7 @@ binderhub:
       pod_quota: 20
       hub_url: https://hub.gke2.staging.mybinder.org
       badge_base_url: https://staging.mybinder.org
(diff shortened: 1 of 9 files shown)


minrk:
There's a problem with artifact registry credentials. Will need to figure that out.

I think I'll leave it pushing to gcr and land the rest

minrk:
I think I've almost got artifact registry figured out, but I don't want to block other things before figuring that out.

manics (approved):

PR #2569, opened 2023-04-25
document adding/removing from the federation

minrk:
make sure we mention updating mybinder status pages, which haven't been updated after recent changes (https://github.com/jupyterhub/mybinder.org-user-guide/pull/277)

remove redundant federation status table, instead linking to the canonical one


manics:
Also Grafana as pointed out by @rgaiacs https://github.com/jupyterhub/mybinder.org-deploy/issues/2572

minrk:
Added mention of grafana, events archive credentials

PR #2580, opened 2023-04-26
add GESIS back to prod federation

### LINKED ISSUE #2577: invalid build token on gesis
gesis builds, when run via the federation proxy, fail with 'invalid build token'

Test URL, with gesis in the staging federation: https://staging.mybinder.org/v2/gh/xbrlus/xule/jupyter?filepath=sample.ipynb (#2576)

@rgaiacs there should be hints in the logs of the `binder` pods, but only gesis has access to these. 

Questions:

- do you have `BinderHub.build_token_secret` config set?
- do you have more than one `binder` replica?

Useful log statements to look for:

- [Generating random build token secret](https://github.com/jupyterhub/binderhub/blob/cef23ed1594b4509260d4d85fef2c2752f38dee0/binderhub/app.py#L700-L701)
- [Failure to validate build token](https://github.com/jupyterhub/binderhub/blob/cef23ed1594b4509260d4d85fef2c2752f38dee0/binderhub/base.py#L86)
- [Build token from mismatched origin](https://github.com/jupyterhub/binderhub/blob/cef23ed1594b4509260d4d85fef2c2752f38dee0/binderhub/base.py#L92)

Looking at our federation redirector, it looks like the `/v2/` page and the `/build/` page are likely to be served by different federation members, and as a result the `BinderHub.build_token_secret` config value must be the same across the federation. We set it in `secrets/config/common.yaml` in this repo, but GESIS does not share our config. I think it probably needs to be copied. If this is the issue, as I suspect it is, you would see the 'Failure to validate build token' errors, and setting this config will fix it, which can be verified at the above staging URL.

minrk:
start with a lower weight, in case there are issues

closes #2577

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -225,8 +225,8 @@ federationRedirect:
       health: https://gke.mybinder.org/health
       versions: https://gke.mybinder.org/versions
     gesis:
-      url: https://notebooks.gesis.org/binder/
-      weight: 0
+      url: https://notebooks.gesis.org/binder
+      weight: 20
       health: https://notebooks.gesis.org/binder/health
       versions: https://notebooks.gesis.org/binder/versions
     ovh2:


rgaiacs:
Looks goot to me. Thanks!

PR #2588, opened 2023-04-27
Remove gke from federation (later today)

minrk:
Will deploy this in the evening (Thursday), CEST, then spend tomorrow (Friday) actually deleting things


minrk:
Deploying this!

PR #2593, opened 2023-04-27
try to enable proxy protocol for OVH ingress

minrk:
per [ovh documentation](https://help.ovhcloud.com/csm/en-gb-public-cloud-kubernetes-getting-source-ip-behind-loadbalancer) the proxy protocol is not enabled by default for load balancers.

One piece that's missing, and I can't tell how important it is, is the 'egress-ips' annotation they describe for getting the load balancer IP doesn't exist. I'm not sure it's critical, though. We'll see.

I believe this is the source of rate limit errors on OVH, because everybody looks like they are coming from the load balancer (10.0.0.160, 10.0.0.164)

### DIFF
--- a/config/ovh2.yaml
+++ b/config/ovh2.yaml
@@ -117,6 +117,11 @@ ingress-nginx:
       enabled: true
     service:
       loadBalancerIP: 162.19.17.37
+      annotations:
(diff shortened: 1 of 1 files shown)


minrk:
This appears to have worked

Issue #2586, opened 2023-04-26
GESIS not yet publish to events archive


minrk:
Adding some tasks for GESIS visibility (@rgaiacs)

The GESIS deployment is not yet publishing launch events to the archive. This is more tedious:

1. for deployments not on GCP, and
2. for deployments not managed from this repo

Reference links:

- publishing events to the archive is configured [here](https://github.com/jupyterhub/mybinder.org-deploy/blob/339ccb1de8107dc7854cac45f0a5b6e99937a91b/mybinder/values.yaml#L200-L219)
- GKE clusters don't need further configuration, but outside GKE (or outside our GCP project, maybe?) need a service account.
  These accounts are configured [in terraform](https://github.com/jupyterhub/mybinder.org-deploy/blob/339ccb1de8107dc7854cac45f0a5b6e99937a91b/terraform/gcp/prod/main.tf#L17), and can be retrieved via `terraform output events_archiver_keys`.
  For OVH, a secret is added to the chart [here](https://github.com/jupyterhub/mybinder.org-deploy/blob/main/mybinder/templates/events-archiver/secret.yaml#L1) and mounted in the binder pod [here](https://github.com/jupyterhub/mybinder.org-deploy/blob/339ccb1de8107dc7854cac45f0a5b6e99937a91b/config/ovh2.yaml#L25-L34) (in our chart, the secret itself is added to [eventsArchiver.serviceAccountKey](https://github.com/jupyterhub/mybinder.org-deploy/blob/339ccb1de8107dc7854cac45f0a5b6e99937a91b/mybinder/values.yaml#L555-L557) helm config, in secrets/config/ovh2.yaml).

rgaiacs:
cc @minrk 

I did the follow steps

1. Add event log Python code to `extraConfig` (https://github.com/gesiscss/orc2/blob/main/helm/gesis/values.yaml#L115)
2. Create Kubernetes secret volumne (https://github.com/gesiscss/orc2/blob/main/helm/gesis/templates/events-archiver-key.yaml)
3. Mount secret volumne (https://github.com/gesiscss/orc2/blob/main/helm/gesis/values.yaml#L94)

How can I check that things are working?

PR #2594, opened 2023-04-27
log events-archive logger hookup

minrk:
so it's easier to tell that it's hooked up

cf #2586


mybinder/values.yaml
```diff
@@ -215,7 +218,9 @@ binderhub:
           client = google.cloud.logging.Client()
```
minrk, line 218: This line will raise an error if it lacks the correct credentials

PR #2604, opened 2023-04-28
make ovh prime again

minrk:


minrk:
I don't think any of last night's issues have been fixed, but at least some diagnostics have been added.

PR #2610, opened 2023-04-28
stop old prod nodes

minrk:
leaving only the inexpensive core pool


minrk:
This has been deployed. The prod cluster is currently running happily on a single 2-core node.

PR #2611, opened 2023-04-28
serve static files from a static pod

minrk:
instead of assuming there's a binder pod to serve

We are serving 503 on badges, because the static ingress points to a service that doesn't exist.

Seems simpler to serve actual static files.


minrk:
Incredibly surprised that this appears to have worked on the first try.

Issue #2556, opened 2023-04-18
Deploy a BinderHub on AWS and add it to the federation


choldgraf:
In https://github.com/jupyterhub/team-compass/issues/642 we discussed that we would like to deploy a BinderHub on AWS so that we can start to bring in credits resources for the federation via that route. This is an issue to track that effort.

### Technical steps

I believe that @manics and @sgibson91 may have extra context about what is _technically_ possible here and what we need to do. **Can you two flesh out the to-dos here?**

1. Set up the BinderHub on AWS
2. Connect with funds
3. Connect to the federation
4. Ensure others on the team have access to maintain
5. Document this hub for others to learn from

### Funding steps

We need to find credits for AWS that will let us run the infrastructure. There were a few leads mentioned before:

1. ~~@scottyhq may have credits for a few months of service.~~ The credits are now gone https://discourse.pangeo.io/t/aws-pangeo-jupyterhubs-to-shut-down-friday-march-17/3228
6. @cgentemann mentioned that the NASA TOPS program may allow us to apply for credits in the coming months.
7. https://github.com/jupyterhub/mybinder.org-deploy/issues/2629
9. @rabernat may have funds that could support Binder via a grant for Pangeo, but this may require us deploying the AWS cluster in a specific way
   - xref: https://github.com/jupyterhub/mybinder.org-deploy/issues/2449

We'll then need to connect those credits to the deployed infrastructure so that it can start running with them.

sgibson91:
https://github.com/jupyterhub/mybinder.org-deploy/pull/2467 was my attempt. I last visited when setting up the auto-scaler (which is another thing that doesn't come out of the box with EKS clusters). It worked, but it spun up more nodes than the maximum desired nodes I'd configured via terraform and I was confused as to whether that was intentional.

I have access to Scott's AWS account which is where I was testing these things.

minrk:
If it's a blocker, would it help to deploy initially with a fixed size and capacity to get things started, then figure out the autoscaler later? OVH 1 always had a fixed size.

sgibson91:
@minrk Maybe blocker isn't quite the right word. The autoscaler works, just while autoscaling it exceeds the maximum number of nodes I set. And that concerns me because I'm not sure if that's okay, or if it's a misconfiguration. I also don't have the capacity to work it out right now.

sgibson91:
[...]

PR #2652, opened 2023-06-01
EKS in AWS Curvenote account

manics:
This is enough to deploy a basic EKS cluster (e.g. `kubectl get nodes` works). The Terraform is originally copied from an EKS cluster I deployed several months ago, unrelated to jupyter/binder.
Todos include:
- [x] OIDC (IRSA and GitHub OIDC)
- [x] Test a manual BinderHub deployment
- [x] Add docs on the private deployment config (setting up limited IAM roles, creating service linked roles, creating a permissions boundary and require it for `iam:CreateRole`)
- [x] Add docs on how this is setup
- [x] Deploy BinderHub via CI

Related to
- https://github.com/jupyterhub/mybinder.org-deploy/issues/2556

The deployment of BinderHub and other applications will be handled in a separate PR


.github/workflows/aws-dev.yml
```diff
@@ -0,0 +1,40 @@
+# This is a temporary workflow to check whether AWS OIDC EKS access is working
```
manics, line 1: I'll delete workflow this in a follow-up PR where I add the binderhub deployment

.pre-commit-config-terraform.yaml
```diff
@@ -0,0 +1,20 @@
...
+# Common tasks
+#
+# - Run on all files: pre-commit run --all --config .pre-commit-config-terraform.yaml
+#
+# Prerequisites:
+# - terraform
+# - tflint
+
+# Currently only aws/curvenote is checked
+files: "^terraform/aws/curvenote/"
```
manics, line 12: I think we should run as many static checks as possible on our terraform code since we can't really test them, though we can discuss that later. For now I've limited it to the new AWS Curvenote terraform.

.github/workflows/aws-dev.yml
manics: .... though it's intermittently broken at the moment due to a change in certificates
https://github.blog/changelog/2023-06-27-github-actions-update-on-oidc-integration-with-aws/

This is configured in the private AWS config repo since it requires elevated account privileges.

manics:
This is ready for review! This is just the Terraform for deploying the EKS cluster on Curvenote's AWS account. Instructions are in the README.

I've deployed a limited BinderHub to https://3.13.147.101.nip.io/
Since this is just for testing:
- maximum pod lifetime in the idle culler is set to 15 minutes though it doesn't always work due to https://github.com/jupyterhub/mybinder.org-deploy/issues/2686
- I've deployed an ECR registry cleaner which deletes images that haven't been pulled after ~1 day
- Single node

I'll open a separate PR for the Helm chart deployment.

manics:
I've pushed some changes so some (but not all) Terraform changes can be automatically deployed by GitHub workflow in future. For example, I think K8s upgrades and changes to node sizes/numbers can be deployed without manual intervention.

This requires some repo changes:
- `TFPLAN_ARTIFACT_PASSPHRASE` secret
- `aws-curvenote` environment with branch restriction and required reviewers for deployments
 
See the `## Automatic deployment (GitHub workflows)` section of the README file included here for details.

If everyone's happy with this plan I'll make the relevant changes in the mybinder.org-deploy repo settings before his is merged.

.github/workflows/aws-dev.yml
manics: Now fixed forever (hopefully) since AWS added a workaround to avoid the dependence on GitHub's intermediate certificates https://github.com/aws-actions/configure-aws-credentials/issues/357#issuecomment-1626357333

manics:
I've added some screenshots of the deployment workflow to the `terraform/aws/curvenote/README`: https://github.com/manics/mybinder.org-deploy/blob/aws-curvenote/terraform/aws/curvenote/README.md

.github/workflows/terraform-deploy-aws-curvenote.yml
```diff
@@ -0,0 +1,116 @@
+# See terraform/aws/curvenote/README.md
+name: Terraform aws-curvenote
+
+on:
+  push:
+    branches:
+      - main
+      - aws-curvenote2
```
yuvipanda, line 8: I'm guessing this would need to be removed before merge?

.github/workflows/terraform-deploy-aws-curvenote.yml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Download plan
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ env.TFPLAN }}
+          path: ${{ env.WORKDIR }}
+
+      - name: Decrypt plan
+        run: |
+          echo ${{ secrets.TFPLAN_ARTIFACT_PASSPHRASE }} | gpg --batch --yes --passphrase-fd 0 --decrypt --cipher-algo AES256 --output "${TFPLAN}" "${TFPLAN}.gpg"
```
yuvipanda, line 111: Can we use anything other than gpg? [age](https://github.com/FiloSottile/age) perhaps?

terraform/aws/curvenote/README.md
```diff
@@ -0,0 +1,139 @@
...
+terraform init
+terraform apply
+cd ..
+```
+
+If you want to use this to create multiple deployments you **must** modify
+
+`terraform { backend "s3" { ... } }`
+
+in [`provider.tf`](provider.tf) to use a different key and/or bucket.
```
yuvipanda, line 26: I would love for us to just use s3 or gcs as the backend from the very start, and just not use filesystem based lock files at all. I am worried about corruption, and rollbacks being hard if the terraform state is lost when it's applied on GH Actions - the plan succeeds but the apply fails (has happened to me many times).

So the concrete suggestion here is to not use local state files at all, and just start off on s3 or gcs.

terraform/aws/curvenote/bootstrap/backend.tf
```diff
@@ -0,0 +1,71 @@
+# Initial setup of S3 bucket to store tfstate file
+variable "bucket-name" {
+  type = string
+  # python -c 'import random; import string; print("".join(random.choices(string.ascii_lowercase + string.digits,k=12)))'
```
yuvipanda, line 4: I'd suggest `pwgen -1 12` instead?

terraform/aws/curvenote/bootstrap/backend.tf
```diff
@@ -0,0 +1,71 @@
+# Initial setup of S3 bucket to store tfstate file
+variable "bucket-name" {
```
yuvipanda, line 2: Maybe call this `state_bucket_name`?

terraform/aws/curvenote/bootstrap/backend.tf
```diff
@@ -0,0 +1,71 @@
...
+
+resource "aws_s3_bucket" "bucket" {
+  bucket = var.bucket-name
+}
+
+resource "aws_s3_bucket_server_side_encryption_configuration" "bucket_encryption" {
+  bucket = aws_s3_bucket.bucket.id
+  rule {
+    apply_server_side_encryption_by_default {
+      sse_algorithm = "AES256"
```
yuvipanda, line 43: Do we need to specify this? I thought it's on by default now https://aws.amazon.com/blogs/aws/amazon-s3-encrypts-new-objects-by-default/

terraform/aws/curvenote/eks-cluster.tf
```diff
@@ -0,0 +1,111 @@
...
+  enable_irsa                   = var.enable_irsa
+  iam_role_permissions_boundary = local.permissions_boundary_arn
+
+  eks_managed_node_group_defaults = {
+    capacity_type                 = "SPOT"
+    iam_role_permissions_boundary = local.permissions_boundary_arn
+  }
+
+  eks_managed_node_groups = {
+    worker_group-1 = {
```
yuvipanda, line 47: Let's stick to just using `_` or just using `-`? I'd prefer just using `_`

terraform/aws/curvenote/eks-cluster.tf
```diff
@@ -0,0 +1,111 @@
...
+locals {
+  permissions_boundary_arn = (
+    var.permissions_boundary_name != null ?
+    "arn:aws:iam::${data.aws_caller_identity.current.account_id}:policy/${var.permissions_boundary_name}" :
+    null
+  )
+}
+
+# This assumes the EKS service linked role is already created (or the current user has permissions to create it)
+module "eks" {
```
yuvipanda, line 17: I don't want to block this PR on this as it's a significant change, but just a few weeks ago I had to rip out this particular eks module out of terraform I maintain for pangeo-forge - https://github.com/yuvipanda/pangeo-forge-cloud-federation/pull/4. Ongoing maintanence was difficult and quite buggy unfortunately.

I'm ok with us giving it a try here, but given how simple the pangeo-forge env was I'm a little sad this module didn't work out.

terraform/aws/curvenote/k8s-access.tf
```diff
@@ -0,0 +1,122 @@
...
+      {
+        Action = "sts:AssumeRoleWithWebIdentity"
+        Effect = "Allow"
+        Principal = {
+          Federated = data.aws_iam_openid_connect_provider.github_oidc_provider[0].arn
+        }
+        Condition = {
+          StringLike = {
+            "token.actions.githubusercontent.com:sub" = [
+              # GitHub repositories and refs allowed to use this role
```
yuvipanda, line 46: this is pretty cool!

terraform/aws/curvenote/outputs.tf
```diff
@@ -0,0 +1,18 @@
+# output "cluster_name" {
```
yuvipanda, line 1: Should this be uncommented or removed from the PR?

yuvipanda (commented):
Thank you so much for working on this, @manics. I've left some comments inside.

I think two overarching questions for me.

1. I have so far not trusted `terraform plan` to run on CI, anywhere. In general, I always recommend humans run it manually and type the 'yes', after making a PR with proposed changes. I've personally accidentally run terraform apply that has deleted entire clusters, user data, etc and had to ctrl-c them - things that went past review by others of both the code and the plan. Due to the intense destruction possible, I've never trusted it to auto run on CI. I'm not opposed to it here because you've already done all the hard work (the github oidc stuff is pretty cool!), but just wanted to ask if you're open to not having terraform apply be run on CI?
2. I don't think a lot of this is curvenote specific - is a curvenote directory with `.tf` files fully necessary? Can they not be just a `.tfvars` file? I am trying to understand what makes it special vs any other AWS setup we might have in the future. 
3. I left another comment inline about the eks module.

In general, I think it's alright to merge this with minor modifications and then evolve this over time.

Thank you so much for working on this!

.github/workflows/terraform-deploy-aws-curvenote.yml
manics: Yes, I'm using it to test the continuous deployment process on my fork. I'll remove it just before merging (or in a follow-up PR).

.github/workflows/terraform-deploy-aws-curvenote.yml
manics: I'm using gpg because it's pre-installed in the base environment. This is purely for internal GH workflow use- there's no way to privately pass an artefact between separate jobs, and since this is a public repo it needs to be encrypted. It's not intended to be used by people/maintainers since if we need the full details we can run Terraform anyway, it's just a way to ensure the public can't view any sensitive information in the plan. My plan is to create a random GitHub secret `TFPLAN_ARTIFACT_PASSPHRASE` and then throw it away so we'll never know the secret.

terraform/aws/curvenote/README.md
manics: This is part of the setup/bootstrapping stage- you need to create and configure the S3 bucket/prefix and DynamoDB table before you can use it!

terraform/aws/curvenote/bootstrap/backend.tf
manics: Maybe- though `pwgen` isn't installed on my system, whereas Python is almost guaranteed to be installed on mybinder maintainers' systems.

terraform/aws/curvenote/bootstrap/backend.tf
manics: Does that apply to the bucket's default policy, or only to the S3 upload API call?

terraform/aws/curvenote/eks-cluster.tf
manics: It's been working fine for me, I've used it for a few clusters now. What problems did you run into?

manics:
> I have so far not trusted terraform plan to run on CI, anywhere. In general, I always recommend humans run it manually and type the 'yes', after making a PR with proposed changes. I've personally accidentally run terraform apply that has deleted entire clusters, user data, etc and had to ctrl-c them - things that went past review by others of both the code and the plan. Due to the intense destruction possible, I've never trusted it to auto run on CI. I'm not opposed to it here because you've already done all the hard work (the github oidc stuff is pretty cool!), but just wanted to ask if you're open to not having terraform apply be run on CI?

You're right that it's a potentially very destructive action, but the same applies to Helm or any other continuous deployment process. In practice I've seen many more errors, near-misses, and other problems due to relying on manual terraform deployments in a team environment. Things like using the wrong AWS account profile, mixing up dev and prod, deploying the wrong repo, using the wrong branch, having an out-of-date branch, forgetting to push commits, pushing commits but not deploying, closing a laptop in the middle of a long running deployment leading to a corrupt state file, etc. The two step GitHub job with environment requiring manual approval can be setup to notify particular named users of a pending plan. It's not as good as a full hosted application for managing Terraform deployments, but I think it's good enough given the constraints of a public GitHub repo.

> I don't think a lot of this is curvenote specific - is a curvenote directory with .tf files fully necessary? Can they not be just a .tfvars file? I am trying to understand what makes it special vs any other AWS setup we might have in the future.

Good point. I've moved it into a module `binder-eks` instead of relying on tfvars files. The benefit of the module is you can just do `cd <deployment>; terraform apply` for each deployment as with all the other clouds. Using variables would require `terraform apply -var-file=<deployment>.tfvars` which is more error prone.

terraform/aws/curvenote/eks-cluster.tf
yuvipanda: @manics https://github.com/yuvipanda/pangeo-forge-cloud-federation/issues/3 was the problem when I tried to use it in pangeo-forge, and I couldn't debug what was going on after a few hours. Eventually I just had to tear it all down and rebuild it without the module. It was a similar setup with the 2i2c infra (where we are still using eksctl directly)

.github/workflows/terraform-deploy-aws-curvenote.yml
manics, line 8:
```suggestion
```

yuvipanda:
I will schedule some time for wednesday / thursday to give this another thorough looking over! Hopefully merge soon. One comment before that is:

> Good point. I've moved it into a module binder-eks instead of relying on tfvars files. The benefit of the module is you can just do cd <deployment>; terraform apply for each deployment as with all the other clouds. Using variables would require terraform apply -var-file=<deployment>.tfvars which is more error prone.

To me, the `.tfvars` are useful when you need to have many instances. So when we have another AWS binder, we could just add another `.tfvars` file rather than copy paste a whole directory. Could we not deal with the possible errors of not passing `-var-file` by making variables as required instead?

.github/workflows/terraform-deploy-aws-curvenote.yml
yuvipanda: I'd still love to get rid of it. I basically can't trust myself to ever verify if a GPG commandline is actually doing what it is supposed to do. `age` can come from apt here, and we can just use it to symmetrically encrypt / decrypt. then:

1. Make an private age key with `age-keygen -o <some-file>`. The file is just one line, we put it into a GitHub secret
2. When we wanna encrypt, `echo $KEY > <some-file>; age --identity <some-file> --encrypt --output <encrypted-output-file> <unencrypted-input-file>`. We can also pass in input via stdin.
3. And to decrypt, do it in reverse: `echo $KEY > <some-file>; age --identity <some-file> --decrypt --output <unencrypted-output-file> <encrypted-input-file>`. The input can come via stdin too.

IMO this is much simpler to understand and verify, and given there's already a lot of complexity here I'd love to not touch GPG at all.

manics: OK, I'll see if I can get that to work.

manics:
> So when we have another AWS binder, we could just add another .tfvars file rather than copy paste a whole directory.

This is a module, so you just reference it as a path, you don't need to copy anything:
https://github.com/manics/mybinder.org-deploy/blob/8c1cfb1c9080a0c325a90f7cac113faa88aa519f/terraform/aws/curvenote/main.tf#L35

terraform/aws/curvenote/.terraform.lock.hcl
minrk, line 1: probably don't want to commit the lock, right?

minrk: resolved in meeting that this _is_ something we want to do.

terraform/aws/binder-eks/k8s-access.tf
```diff
@@ -0,0 +1,122 @@
...
+        Action = "sts:AssumeRoleWithWebIdentity"
+        Effect = "Allow"
+        Principal = {
+          Federated = data.aws_iam_openid_connect_provider.github_oidc_provider[0].arn
+        }
+        Condition = {
+          StringLike = {
+            "token.actions.githubusercontent.com:sub" = [
+              # GitHub repositories and refs allowed to use this role
+              "repo:jupyterhub/mybinder.org-deploy:ref:refs/heads/main",
```
yuvipanda, line 47: This is pretty cool I must say :)

terraform/aws/binder-eks/k8s-access.tf
```diff
@@ -0,0 +1,122 @@
...
+              # TODO: Remove this, just for development:
+              "repo:manics/mybinder.org-deploy:ref:refs/heads/aws-curvenote",
```
yuvipanda, line 49: Just a reminder to get rid of this before merging.

terraform/aws/binder-eks/k8s-access.tf
```diff
@@ -0,0 +1,122 @@
...
+        }
+        Condition = {
+          StringLike = {
+            "token.actions.githubusercontent.com:sub" = [
+              # GitHub repositories and refs allowed to use this role
+              "repo:jupyterhub/mybinder.org-deploy:ref:refs/heads/main",
+              # Can't have branch and environment in the same condition
+              # https://github.com/aws-actions/configure-aws-credentials/issues/746
+              "repo:jupyterhub/mybinder.org-deploy:environment:aws-curvenote",
+              # TODO: Remove this, just for development:
```
yuvipanda, line 84: Reminder

terraform/aws/binder-eks/vpc.tf
```diff
@@ -0,0 +1,30 @@
+# data "aws_availability_zones" "available" {}
+
+module "vpc" {
+  source  = "terraform-aws-modules/vpc/aws"
+  version = "5.0.0"
+
+  name = var.cluster_name
+  cidr = "10.0.0.0/16"
+  # azs                  = data.aws_availability_zones.available.names[:1]
```
yuvipanda, line 9: Let's get rid of this commented out line, or use it?

terraform/aws/curvenote/bootstrap/backend.tf
yuvipanda: I was trying to figure out if we should use the `secrets` module instead, given that `random` is a PRNG and shouldn't be used for 'security' purposes. However, this is not being used in any cryptographic manner, just as a way to avoid clashes - so is fine.

terraform/aws/curvenote/bootstrap/backend.tf
yuvipanda: Pretty sure it applies to the default policy. Can be verified by taking this out and looking in the AWS console to see if this has changed?

yuvipanda (changes requested):
Some super minor changes requested, but the primary one would be to replace `gpg` with `age`. After that I think this is good to go! 

I'm still somewhat skeptical of the EKS module, but only because I *really* want it to work and have been burnt by that in the past. However, this is generally a fast moving space anyway, and I hope this does work. Let's try it out!

Thank you so much, @manics!

yuvipanda (approved):
Thanks a lot, @manics! I think this is ready to merge! I've marked it as approved, so you should merge whenever you feel convenient.

EXCITED TO SEE THIS WORK!

manics:
@yuvipanda Thanks for working through this! I've pushed a couple more minor changes, if no-one else has comments I'll setup the GitHub secret and environment tomorrow and merge.

manics:
I've created the secret `TFPLAN_ARTIFACT_SECRET_KEY`
and the Environment:
image
So far I've only added myself to be notified when there is a pending terraform change, but I think we should change that to a GitHub team- to be followed up.

PR #2682, opened 2023-06-17
source cert-manager.env before calling deploy.py

manics:
Follow-up to https://github.com/jupyterhub/mybinder.org-deploy/pull/2672


manics:
Merging since it's an obvious bug

PR #2698, opened 2023-07-12
Deploy binderHub and other K8s apps on AWS curvenote

manics:
Follow-on from https://github.com/jupyterhub/mybinder.org-deploy/pull/2652 for AWS Curvenote. Deploys
- Some kube-system charts (load balancer, block storage)
- BinderHub
- ECR registry helper and cleaner charts

This has been manually deployed (`./deploy.py ....`)


manics (commented):
I think this is ready for review!

The main missing things are:
- Adding this to the redirector: I'd like to leave this for a later PR, and focus on getting the continuous deployment running first .
- `GOOGLE_APPLICATION_CREDENTIALS`: Could be done now, or in a follow-up

config/curvenote.yaml
```diff
@@ -0,0 +1,324 @@
...
+
+    10-external-registry-helper: |
+      import json
+      from os import getenv
+      from tornado import httpclient
+      from traitlets import Unicode
+      from binderhub.registry import DockerRegistry
+
+
+      class ExternalRegistryHelper(DockerRegistry):
```
yuvipanda, line 60: I'm guessing this will eventually go upstream into binderhub

deploy.py
```diff
@@ -140,6 +143,27 @@ def setup_auth_gcloud(release, cluster=None, dry_run=False):
...
+
+    eks_kubeconfig = [
+        "aws",
+        "eks",
+        "update-kubeconfig",
+        "--name",
+        AWS_DEPLOYMENTS[cluster],
+    ]
+    stdout = check_output(eks_kubeconfig, dry_run)
+    print(stdout)
```
yuvipanda, line 164: Should this be removed?

config/curvenote.yaml
manics: Hopefully! I think we should try it in production for a bit, and if there are no unforeseen issues add it upstream.

deploy.py
manics: `check_output` will hide stdout, so I think it's useful to print it out as it provides confirmation that auth worked, and it also mirrors the effect of running the command manually.

yuvipanda (approved):
Given the lack of review time, @manics at this point I think you should self merge and iterate so we can add more capactiy. I think overall this looks good to me and we can obviously change after.

THANK YOU FOR WORKING ON THIS!

manics:
@yuvipanda Thanks! I'll merge when I have a decent block of time to deal with any problems.

manics:
@yuvipanda I'm hoping to spend some time on this later today, but let me know if you'd rather I wait until the cryptnono work is finished.

yuvipanda:
@manics nah you can go ahead!

PR #2772, opened 2023-10-12
Revert "Updates binderhub chart to 1.0.0-0.dev.git.3228.h30733fe"

yuvipanda:
Reverts jupyterhub/mybinder.org-deploy#2770

Until we fix https://github.com/jupyterhub/nbgitpuller/issues/329
https://discourse.jupyter.org/t/nb-on-binder-with-parameters-now-resulting-in-404/21860


yuvipanda:
Let's just revert this deploy (which was all frontend) until we fix this.

PR #2776, opened 2023-10-16
Add Erik to list of people with `git-crypt` key

yuvipanda:
Given I just got it from him :)


consideRatio:
Thank you @yuvipanda !

PR #2784, opened 2023-10-23
Tune cryptomining bans

yuvipanda:
Based on observations by @rgaiacs


yuvipanda:
TIL that the easiest / only way to deploy to GESIS is to merge this PR so am gonna do that.

PR #2785, opened 2023-10-23
Ban a fake user cryptomining

yuvipanda:
I've also reported the user to GitHub and then cursed them severely


rgaiacs:
Looks good to me. Don't have merge permission.

PR #2800, opened 2023-11-01
Bump to newer version of cryptnono & re-enable

yuvipanda:
https://github.com/jupyterhub/mybinder.org-deploy/security/advisories/GHSA-j42g-x8qw-jjfh has more information


yuvipanda:
Deployed and tested on staging!

manics (approved):

yuvipanda:
I've tested this on staging, and then killed all pods on kube-system and `staging` to make sure this doesn't kill any more pods.

I've tried this on prod too and works ok. Now trying to roll it out on ovh2 and actually dealing with the various broken nodes.

PR #2804, opened 2023-11-01
Add diff option to deploy.py

manics:
I worked on this in #2698 but decided to make it into a seperate PR so we can discuss whether it's useful.

Adds a `--diff` option that uses `kubectl diff` (builtin) and `helm diff` (requires https://github.com/databus23/helm-diff) which is useful for local development.

It's not ideal since if there's a diff it shows the whole file instead of just the changed region, and CHP/hub will always show a difference because `helm diff` always generates a random token instead of fetching the CHP token.


minrk:
This is great!

If it's useful for reference, I've used [this script](https://gist.github.com/minrk/207d852d8faf613dc840b7238b0278f6) to check out helm-diffs for mybinder.org for some time.

specifically piping helm-diff through `| grep -C 1 -v '^[[:space:]]'` to limit the output to nearby changes.

manics:
I've belatedly realised `helm diff` includes a `--context` argument!

minrk (approved):

PR #2823, opened 2023-11-10
backport event_loop_seconds metric

minrk:
from https://github.com/jupyterhub/jupyterhub/pull/4615

so we can try it out without waiting for a release


yuvipanda (approved):
NICE

minrk:
Working. Still not sure how best to visualize this, but got something:

Screenshot 2023-11-10 at 12 15 24

PR #2834, opened 2023-11-19
Generate events_archiver_keys for AWS curvenote

manics:
Following instructions in https://mybinder-sre.readthedocs.io/en/latest/operation_guide/federation.html#adding-or-removing-a-federation-member


minrk:
Ran terraform to create the key, put the key in secrets/config/curvenote.yaml. For the record, ran specifically:

```
terraform output -json events_archiver_keys | jq -r '.curvenote' | pbcopy
```

stored in `secrets/config/curvenote.yaml` as:

```yaml
eventsArchiver:
  serviceAccountKey: |
    {
      "type": "service_account",
```

Related: it looks like there's no github client credentials set up for the curvenote deployment. That will likely hit rate limits very soon after getting significant traffic. We have not historically shared github client credentials across federation members, but I'm not sure we really need to keep them separate, if you want to copy the one for `prod` (especially since `prod` doesn't actually have a binderhub instance anymore!), that seems fine.

manics:
Thanks for the reminder about GitHub client credentials! It's best practice to rotate credentials every now and then so I'll create a new set.

PR #2836, opened 2023-11-21
track terraform lockfiles

minrk:
Suggested by @manics


manics (approved):
Thanks! `terraform plan` shows no changes for staging and prod

PR #2838, opened 2023-11-23
Curvenote: Add GitHub tokens, and publish events

manics:
Events token was added in https://github.com/jupyterhub/mybinder.org-deploy/pull/2834


manics:
Events are being published, for example see https://archive.analytics.mybinder.org/events-2023-11-23.jsonl

PR #2846, opened 2023-11-30
Curvenote: add to federation!

manics:
Add to mybinder.org, weight 1

### DIFF
--- a/config/curvenote.yaml
+++ b/config/curvenote.yaml
@@ -17,7 +17,7 @@ binderhub:
       memory_request: "2G"
 
     LaunchQuota:
-      total_quota: 10
+      total_quota: 40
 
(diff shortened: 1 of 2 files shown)


minrk:
Hooray! Seems to be working, but it doesn't seem to get ~any traffic (I found a couple in the redirect logs). Maybe bump weight to 10?

A weight of 1 means it will only get assigned on average one _repo_ out of 201 (not 0.5% of builds, could be far less or far more than that), and we don't build all that many unique repos.

Our redirect logic favoring a specific repo->member doesn't give us a good way to distribute load smoothly, because a small number of repos usually represent a very large fraction of load at any given time, meaning that our load is very rarely evenly distributed.

PR #2849, opened 2023-12-01
curvenote: weight=10

manics:
As pointed out by @minrk a weight of 1 is too small for real testing.


manics:
Definitely made a difference :smile: 
image

PR #2940, opened 2024-03-04
Curvenote: increase nodesize to r6a.4xlarge, upgrade K8s to 1.29

manics:
@stevejpurves has enough AWS credits to double capacity until the end of the year.

Doubling the node size from r6a.2xlarge (8 cpu, 64 memory) to r6a.4xlarge (16 cpu, 128 memory) is more efficient than doubling the number of nodes because the maximum number of EKS pods on r6a.2xlarge is 58, whereas with r6a.4xlarge it's 234. Currently nearly half the total pod capacity is taken up by system, BinderHub, and other non-user pods.

This also upgrades K8s to 1.28 (though we could also go to 1.29 https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#available-versions - might try that since it means we won't have to upgrade again in 2024).


manics:
Done!

PR #3076, opened 2024-08-26
Update donate button

manics:
https://numfocus.org/donate-to-jupyter

See https://github.com/jupyterhub/team-compass/issues/742


rgaiacs:
Looks good to me. Merging!

PR #3165, opened 2025-01-13
temporary: redirect all instead of proxying mybinder.org to GESIS

minrk:
should avoid CORS issues with mismatched GESIS page templates


rgaiacs:
Looks good to me.

@minrk you can merge for us to test.

PR #3169, opened 2025-01-17
Add a 2i2c federation member on Hetzner

yuvipanda:
https://2i2c.mybinder.org/ is up now!

https://github.com/2i2c-org/2i2c-org.github.io/pull/356/files#diff-7244b57e647732dd6a8f006bdf63943e1dcb813fa1a085073522ccf40e2cdfc6 has more context - that's also an announcement blog post. It came together quickly.

This is a single node k3s cluster running on Hetzner. It's not as large as we'd like it to be - which is CCX63 on https://www.hetzner.com/cloud. That's 48 vCPUs and 192GB of RAM. And with k3s, we can override the number of pods on a node. Given the current guarantee of 450M, we can put approximately 400 pods on this one node! That runs out to less than $1 / month per user capacity which is pretty good.

Still need to figure out:

1. Access for everyone else on the team
2. Resize the server to be big, and set up k3s again there from scratch + document (I simply followed the quickstart with traefik disabled)
3. Test prometheus and grafana
4. Add 2i2c to list of supporters

Am excited to try this out and see how it goes.

Thanks to @choldgraf, @colliand, @jmunroe and others at 2i2c for supporting me through this.


choldgraf:
It was so fast to spin this up!

yuvipanda:
Am going to try to put the registry in cluster as well, let's see.

minrk:
This is awesome! I won't have time to monitor a deployment rollout until Monday, but this looks great.

mybinder/templates/registry/deployment.yaml
```diff
@@ -0,0 +1,56 @@
...
+          name: registry-config
+      - name: registry-secret
+        secret:
+          secretName: registry-secret
+      - name: registry-storage
+        persistentVolumeClaim:
+          claimName: registry
+      containers:
+      - name: registry
+        image: registry:2.8.3
```
manics, line 42: In a future PR we should add this to the watch-dependencies workflow
https://github.com/jupyterhub/mybinder.org-deploy/blob/b9439b5b15263ff703ea46c07bfd6c4584d1422c/.github/workflows/watch-dependencies.yaml#L33-L39

mybinder/values.yaml
```diff
@@ -10,6 +10,20 @@ cryptnono:
       containerdHostPath: /run/containerd/containerd.sock
       dockerHostPath: /run/dind/docker.sock
 
+registry:
+  enabled: false
+  config:
+    version: 0.1
```
manics, line 16: What is `version` used for?

mybinder/values.yaml
```diff
@@ -10,6 +10,20 @@ cryptnono:
...
+    auth:
+      htpasswd:
+        path: /etc/distribution/auth.htpasswd
+  service:
+    type: ClusterIP
+  storage:
+    filesystem:
+      storageClassName: ""
+      size: 10Gi
```
manics, line 25:
```suggestion
      # Size is currently ignored- using local path provisioner
      size: 10Gi
```
I presume you're just using K3s' default local storage?

manics (approved):
This is great! Do you want to merge this now?

Is your plan to figure out the automated deployment in a future PR, after you work out how to setup external K8s API access?

> This is awesome! I won't have time to monitor a deployment rollout until Monday, but this looks great.

I think it's fine to merge now and revert if necessary- doesn't add a Hetzner GitHub deployment workflow, it only modifies the redirector to direct builds and make Hetzner the prime host.

manics:
Can you check the websocket timeout?
https://2i2c.mybinder.org/v2/gh/jupyterhub/jupyter-remote-desktop-proxy/HEAD?urlpath=desktop
The desktop reproducibly goes blank after ~70 seconds

yuvipanda:
@manics I already sorted out external access (see https://github.com/jupyterhub/mybinder.org-deploy/pull/3169/files#diff-7bd97b74c45fda1d974a0f752288c28f7f38eaf85005a1953adb7dfc6e34df9e)! no github deploy yet

mybinder/values.yaml
yuvipanda: From https://distribution.github.io/distribution/about/configuration/

yuvipanda:
I've a tattoo appt all day, I'll get back to this once that's done!

yuvipanda:
@manics I wonder if that's nginx timeout that needs tuning

yuvipanda:
@manics hmm, I can't seem to reproduce the timeout! There's no external load balancer here, just our nginx. Is it still happening to you?

yuvipanda:
The current node is really small - only a `CCX23`. I'm waiting for a quota increase to make it bigger. I think it's worth waiting for that to happen, as CCX23 is only 16GB of RAM

rgaiacs:
This looks good to me. Thanks 2i2c! <3

yuvipanda:
I've added an encrypted ssh key for ssh access for other team members!

yuvipanda:
I've sent email invites to the hetzner project to @minrk and @manics. @rgaiacs if you share an email address with me i can send one to you too!

yuvipanda:
poking to see if i can switch the registry to object storage already, while we wait for the quota increase. This would mean the instance is purely zero state to a much greater extent.

yuvipanda:
Alright, now we use the hetzner object storage as the storage backend for the registry! and thus we run 2 replicas of the registry as well :)

I'm also leaving in comment the small bit of config change that's required to continue using the filesystem backend as well. The goal here is to make it *as easy as possible* for people to join the federation. And with this, that's simply down to '1 VM'

yuvipanda:
I started adding k3s docs!

manics:
Are you missing a commit? `secrets/hetzner-2i2c.yml` is identical to `secrets/ovh2-kubeconfig.yml`

manics:
I've got SSH access, copying the k3s.yaml kubeconfig file and editing the server IP gives me K8S access!

yuvipanda:
@manics ah yes - i was missing a commit. Added now.

yuvipanda:
I validated that we can change the number of pods on a node by following https://stackoverflow.com/a/65899273. This node currently is set to max 250 pods although it can't support that many (is smol). Shall be added to the documentation on how to setup k3s.

secrets/hetzner-2i2c.yml
minrk, line 1: From the changes in deploy.py, I think this wants to be `secrets/hentzner-2i2c-kubeconfig.yml`

rgaiacs:
@yuvipanda you can send the invitation to `Raniere.CostadaSilva@gesis.org`.

yuvipanda:
The quota increase was approved! I've taken the existing node offlien and bringing up a new node. will be done tonight

yuvipanda:
@rgaiacs done!

yuvipanda:
Created new server from scratch, rebuilt it, and it's all good to go! I'm going to sleep though, so either I'm happy for someone else to merge it (ssh keys updated so you can debug if necessary), and if not I"ll try to find time.

minrk:
Awesome! Dealing with a plumber now, but I'll give it go when I'm free in an hour or two unless someone else is ready first.

minrk:
/test-this-pr

just to make sure the registry validates and doesn't deploy to staging

minrk:
Giving this a try!

PR #3182, opened 2025-01-22
Fixes for 2i2c hetzner member

yuvipanda:
- Put object storage in same DC as the server

  When I recreated the server, I had to do so in nuremberg
  rather than falkenstein because their nuremberg server was full.
  But the object storage for the images remained in the old location.

  I've created a new object storage bucket and moved that!

- Bring registry replicas back to 1

  It doesn't do much since the pulls go directly from hetzner I believe,
  and keeping it at 1 makes debugging easier for right now

- Increase the amount of resources dind has

  As this is a new member, there's gonna be a lot of builds to
  start with. Let's support that with higher resource limits.

- Increase binderhub replicas to 2

- Add suspicious but maybe useful nginx limit increase

- Increase 2i2c quota even more

### DIFF
--- a/config/hetzner-2i2c.yaml
+++ b/config/hetzner-2i2c.yaml
@@ -2,7 +2,7 @@ projectName: hetzner-2i2c
 
 registry:
   enabled: true
-  replicas: 4
+  replicas: 1
   config:
     storage:
       # Uncomment this and comment out the s3 config to use filesystem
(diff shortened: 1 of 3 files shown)


yuvipanda:
I'm also manually increasing `--max-concurrent-uploads` on the dind daemonset after each deployment as that was the source of all the build backups we were seeing earlier.

Issue #3188, opened 2025-01-23
Update test to support new frontend UI


rgaiacs:
BinderHub frontend UI migrated to vanilla HTML to React based app in https://github.com/jupyterhub/binderhub/pull/1856. Because of this, the test https://github.com/jupyterhub/mybinder.org-deploy/blob/6e2a0e2b4557ecffe0dd5521859a907fa21ca5c7/tests/test_federation_redirect.py#L13-L16 requires re-design.

The "simplest" solution is to replace assert to check for

```
<div id="root"></div>
```

from https://github.com/jupyterhub/binderhub/blob/65f807c12c61aed2179293d1bd0f72d117e92793/binderhub/templates/page.html#L24. In this simplest solution will don't need to run the JavaScript. We only check that the React based app has a HTML anchor.

Alternatively, the assert can be replaced with a solution powered by an automation library for browser testing like

- [Playwright](https://playwright.dev/)
- [Selenium](https://www.selenium.dev/)
- [Puppeteer](https://pptr.dev/)
- [Cypress](https://www.cypress.io/)

rgaiacs:
In https://github.com/jupyterhub/mybinder.org-deploy/pull/3189#issuecomment-2612156130, @minrk suggested to use [Playwright](https://playwright.dev/).

The continuous deployment run the test at https://github.com/jupyterhub/mybinder.org-deploy/blob/c46ce00a8e589ed7fa24ff6a7d97a72efa41a134/.github/workflows/cd.yml#L182-L188

A couple of more steps documented at https://playwright.dev/python/docs/ci#github-actions will need to be added before the tests start.

PR #3189, opened 2025-01-23
Re-write test_proxy_page() the simplest way possible

rgaiacs:
Related to https://github.com/jupyterhub/mybinder.org-deploy/issues/3188

A more robust way is to parse the HTML and check for the node instead of use the raw text representation.


manics:
@yuvipanda This may unblock https://github.com/jupyterhub/mybinder.org-deploy/pull/3156#issuecomment-2574753436
Does it make sense to you?

minrk:
I think a playwright test to make sure the page actually loads is probably a good idea, but this helps unblock us

PR #3211, opened 2025-02-08
Increase and centralize build memory limits

yuvipanda:
Follow-up to https://github.com/jupyterhub/mybinder.org-deploy/pull/3204, which failed because it removed the limit, which caused it to be lower than the request. All the reasons given in that PR still apply.

Here I bump up the limit enough until the super popular https://github.com/jupyterlab/jupyterlab-demo is able to build. It's not a functionally super complex environment.yml, and even with simplifications in https://github.com/jupyterlab/jupyterlab-demo/pull/147, it fails to build without the memory increase limit. As mentioned in the other PR, https://github.com/jupyterhub/repo2docker/pull/1402 will get rid of this kind of limit anyway, so it's ok to raise this.

### DIFF
--- a/config/hetzner-2i2c-bare.yaml
+++ b/config/hetzner-2i2c-bare.yaml
@@ -35,12 +35,7 @@ binderhub:
       image_prefix: registry.2i2c-bare.mybinder.org/i-
       # image_prefix: quay.io/mybinder-hetzner-2i2c/image-
       # build_docker_host: /var/run/dind/docker.sock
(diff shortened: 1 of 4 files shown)


rgaiacs:
@yuvipanda Regarding GESIS, the configuration is still in https://github.com/gesiscss/orc2/blob/59a67b86fa4325455041332596532c6afcbcfd8e/helm/gesis-config/production.yaml. I'm trying to move it to this repository.

PR #3213, opened 2025-02-08
Centralize and keep 3 months of metrics

yuvipanda:
I checked hetzner-2i2c, and it had about 10G of usage for 30d of metrics. Let's bump to 90. Eventually I wanna keep at least a year or two


yuvipanda:
I'm doing this because I think we can easily build a image registry cleaner off prometheus metrics.

PR #3178, opened 2025-01-21
[WIP] Streamline deployment of GESIS acceptance server

rgaiacs:
This is related with https://github.com/jupyterhub/mybinder.org-deploy/issues/2797 and https://github.com/jupyterhub/mybinder.org-deploy/issues/3087.

SSH access to GESIS Kubernetes cluster requires access to GESIS VPN. This makes the use of GitHub actions challenge.

Different than https://github.com/jupyterhub/mybinder.org-deploy/pull/3090, this PR assumes that the GESIS Kubernetes is configured. GESIS Kubernetes runs on bare-metal and is configured following [MetalLB + Ingress NGINX Controller](https://kubernetes.github.io/ingress-nginx/deploy/baremetal/#a-pure-software-solution-metallb).

GESIS operates a on-premise GitLab. This PR addes a GitLab CI/CD configuration to deploy the mybinder.org Helm chart to the GESIS Kubernetes using [GitLab agent for Kubernetes](https://docs.gitlab.com/ee/user/clusters/agent/install/).

After https://github.com/jupyterhub/mybinder.org-deploy/pull/3159, any the commit on `main` is pushed to GESIS GitLab and will trigger a new deployment of this mybinder.org Helm chart.

At the moment, the Helm chart deploy fails with

```
Error: Unable to continue with install: IngressClass "nginx" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "interactive-environment": current value is "ingress-nginx"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "interactive-environment": current value is "ingress-nginx"
```

My understanding of the error is that GESIS Kubernetes already has a Ingress Class `nginx` that was configured for MetalLB + Ingress NGINX Controller but the Binder Hub and Jupyter Hub upstream Helm chart tries to create the same Ingress Class `nginx` and this operation is blocked. Is my understanding correct? @manics @yuvipanda @minrk

PR #3217, opened 2025-02-12
Set template config correctly

yuvipanda:
Turns out just setting it on NotebookApp doesn't set it on ServerApp now, and so all of our templates were actually just no-ops for a while.

Discovered while testing https://github.com/jupyterhub/mybinder.org-deploy/pull/3178

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -42,7 +42,7 @@ tags: {}
 
 etcJupyter:
   jupyter_notebook_config.json:
-    NotebookApp:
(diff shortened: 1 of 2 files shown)


mybinder/values.yaml
```diff
@@ -42,7 +42,7 @@ tags: {}
 
 etcJupyter:
   jupyter_notebook_config.json:
```
minrk, line 44: Also need the right filename, I think:

```suggestion
  jupyter_server_config.json:
```

PR #3226, opened 2025-02-16
Run a cronJob every 30m to cleanup docker buildkit cache

yuvipanda:
https://github.com/docker/buildx/issues/1065#issuecomment-1098357043 is the best documentation on this command, and it should fully be able to replace what our imageCleaner currently does. And it's necessary, as I think imageCleaner doesn't actually clean up buildkit cache anymore.

This should be upstreamed into the helm chart, and imageCleaner deprecated / removed. However, I want to merge https://github.com/2i2c-org/binderhub-service/ into binderhub and turn that into the chart first, so we don't have to keep repeating work there. So in the meantime, this cronjob is here so we don't have mybinder.org outages (as happened in https://jupyter.zulipchat.com/#narrow/channel/469744-jupyterhub/topic/mybinder.2Eorg.20outage)

### DIFF
--- a/mybinder/templates/buildkit-pruner.yaml
+++ b/mybinder/templates/buildkit-pruner.yaml
@@ -1,0 +1,31 @@
+{{ if .Values.buildkitPruner.enabled }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: hello
+spec:
+  schedule: {{ .Values.buildkitPruner.schedule | quote }}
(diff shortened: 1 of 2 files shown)


yuvipanda:
testing on the bare instance to see if that works

yuvipanda:
We still need the imageCleaner, until https://github.com/jupyterhub/repo2docker/issues/1414 lands. Then it can be disabled.

PR #3230, opened 2025-02-19
k3s: Document tuning kubernetes image cache

yuvipanda:
When we aren't using disk on the host for anything else, kubernetes can simply use as much of the disk as it wants for caching images. The default for this is upto 85%. However, since we *do* use the disk for our dind, we want to tune this to take up only half that size.

I've already applied this to the 2i2c-hetzner node

Follow-up to https://github.com/jupyterhub/mybinder.org-deploy/pull/3229


yuvipanda:
image

You can immediately see the effect of this

PR #3249, opened 2025-03-08
Disable image cleaner

yuvipanda:
As part of https://github.com/jupyterhub/binderhub/pull/1941, I want to not port imageCleaner to the new chart. Since the image store is not really used to push anymore (since https://github.com/jupyterhub/repo2docker/pull/1421) and we will be able to disregard it completely in the future, we can simply just prune the entire image store than do any complex logic here.


manics:
Cronjobs only run on one node at a time- are we sure they'll be randomly scheduled across all build nodes at a frequent enough interval?

PR #3262, opened 2025-03-24
ci: bump to use ubuntu 24.04 based runners

consideRatio:
Ubuntu 20.04 runners on GitHub Actions now experience brownouts, so let's move away from using them.

image


minrk:
Thanks for keeping up with these!

Issue #3264, opened 2025-03-28
New GESIS Server at Hetzner Online GmbH


rgaiacs:
A decision at GESIS was made to replace the server running https://notebooks.gesis.org/ with a server at [Hetzner Online GmbH](https://www.hetzner.com) following the [successful project](https://2i2c.org/blog/2025/binder-singlenode/) that @yuvipanda and 2i2c did. All credits to @arnim.

The next steps are

- [x] @arnim to get credentials from GESIS IT
- [x] @rgaiacs to clone https://github.com/jupyterhub/mybinder.org-deploy/blob/b0d4d8ef56c99d887d74b900527e6c0de68da17f/config/hetzner-2i2c.yaml to GESIS
- [ ] @rgaiacs to bring the GESIS Corporate Design to gesis.mybinder.org
- [x] new domains to be created
  - [ ] `gesis.mybinder.org`
  - [ ] `hub.gesis.mybinder.org`
  - [ ] `grafana.gesis.mybinder.org`
  - [ ] `prometheus.gesis.mybinder.org`
  - [ ] `static.gesis.mybinder.org`
  - [ ] `binder.gesis.org` or similar yet to be decided by GESIS with @arnim 
- [x] @rgaiacs to do the deployment

arnim:
server at Hetzner <- CCX63 ;)

Maybe some extra questions to consider: What location should we use? I think @yuvipanda mentioned NBG1, right?
We should also place the image builder on a separate volume.

Overall, I think it would be good to set things up as similarly as possible to the other node. The only thing we’d need in addition is a GESIS CSS-themed launch form (like binder.gesis.org or something similar).

rgaiacs:
@arnim and @rgaiacs are meeting with GESIS IT to get credentials for Hetzner on the 3rd April at 14:00 CEST.

We will send invites to the project to @yuvipanda @minrk and @manics.

rgaiacs:
I'm reopening this because the continuous deployment is failing as described in https://github.com/jupyterhub/mybinder.org-deploy/issues/3266. The continuous deployment fail is not related to the new GESIS server.

arnim:
Having such a mail from Hetzner could indeed help us with our IT.

yuvipanda:
@rgaiacs @arnim I've forwarded it to you both.

rgaiacs:
We are almost done with the new GESIS server.

The ingress controller is not working because of

```
[...]

PR #3265, opened 2025-03-31
Configure new GESIS Server at Hetzner Online GmbH

rgaiacs:
Closes https://github.com/jupyterhub/mybinder.org-deploy/issues/3264

- [x] Get credentials from GESIS IT
- [ ] Add new credentials to this repository
- [x] Clone https://github.com/jupyterhub/mybinder.org-deploy/blob/b0d4d8ef56c99d887d74b900527e6c0de68da17f/config/hetzner-2i2c.yaml to GESIS f566bd80d62035722932e6fa5f05eacc6309fda4
- [ ] Merge https://github.com/jupyterhub/binderhub/pull/1923
- [ ] Merge https://github.com/jupyterhub/binderhub/pull/1953
- [ ] Merge https://github.com/jupyterhub/binderhub/pull/1954
- [ ] Bring the GESIS Corporate Design to gesis.mybinder.org dec96acbc0ed7cdb16bc8f3f451c7ef068d37c3c
- [ ] new domains to be created
  - [ ] `gesis.mybinder.org`
  - [ ] `hub.gesis.mybinder.org`
  - [ ] `grafana.gesis.mybinder.org`
  - [ ] `prometheus.gesis.mybinder.org`
  - [ ] `static.gesis.mybinder.org`
  - [ ] `binder.gesis.org` or similar
- [x] Install K3s on the new server
- [ ] Do the deployment


config/hetzner-gesis.yaml
```diff
@@ -0,0 +1,126 @@
+projectName: hetzner-gesis
+
+# We share the registry with 2i2c
+registry:
+  enabled: false
```
yuvipanda, line 5: This still needs to be enabled. What we'll share is the backing object store that the registry talks to, so the *storage* that the registry uses will be the same. But we still need a registry to serve them from here. So this config should match exactly what's in the hetzner-2i2c one (including secrets) so it can use the same storage backend but run its own server.

rgaiacs:
@rgaiacs and @arnim have access to the new GESIS Server at Hetzner Online GmbH. [K3s](https://k3s.io/) is configured in the new GESIS Server at Hetzner Online GmbH.

rgaiacs:
@yuvipanda I want to invite you to the project but GESIS IT did not even added me or @arnim to the project. The IP of the server is `78.46.233.119`.

yuvipanda:
Ah! Is that something they'd be willing to do? I think it's important for us to have access there, for two reasons:

1. Hetzner sends out network abuse reports for us to deal with, and we must deal with them. I'm guessing GESIS IT doesn't wanna do that, nor can they realistically
2. Need access to the console for setting up additional disks (For DIND) as well as firewall rules

rgaiacs:
> I think it's important for us to have access there, for two reasons

I agree with you. @arnim and I explained the reasons. But the person that assisted us today to get the server running could not make the decision regarding giving us access to the dashboard.

rgaiacs:
> Need access to the console for setting up additional disks (For DIND) as well as firewall rules

We mirror the configuration of the server provided by 2i2c that you configured.

yuvipanda:
I've added DNS records for gesis.mybinder.org and *.gesis.mybinder.org to 78.46.233.119. Hopefully that unblocks you, @rgaiacs.

Hetzner's network abuse policy is that we must respond and take action often within 48 or 72h, or they will simply stop all network traffic to the server. So, I think we must have project access before we can take this into rotation.

rgaiacs:
This PR will only configure the server. In another PR we will add the server to the federation. I'm merging this as it only changes the files in the GESIS namespace.

PR #3267, opened 2025-04-03
Add GESIS Server at Hetzner Online GmbH to federation

rgaiacs:
Related to https://github.com/jupyterhub/mybinder.org-deploy/issues/3264

In https://github.com/jupyterhub/mybinder.org-deploy/pull/3265, @yuvipanda said

> Hetzner's network abuse policy is that we must respond and take action often within 48 or 72h, or they will simply stop all network traffic to the server. So, I think we must have project access before we can take this into rotation.

cc @arnim

### DIFF
--- a/config/prod.yaml
+++ b/config/prod.yaml
@@ -239,6 +239,12 @@ federationRedirect:
       weight: 0
       health: https://2i2c-bare.mybinder.org/health
       versions: https://2i2c-bare.mybinder.org/versions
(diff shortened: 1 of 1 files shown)


rgaiacs:
I will merge this. It replaces the old GESIS server with the new GESIS server. It uses a weight of 5 to avoid the container registry to be overwhelm. I will later increase the weight.

Issue #3168, opened 2025-01-17
Fail to install dependencies with Python 3.13 and 3.12


rgaiacs:
### Bug description

One of the dependency, `ruamel-yaml-clib`, is not compatible with Python 3.13 and 3.12.

### How to reproduce

```bash
micromamba create -n py3.13 python=3.13
```

```bash
micromamba activate py3.13
```

```bash
python -m pip install -r requirements.txt
```

#### Expected behaviour

Install dependencies with success.

#### Actual behaviour

```
        111 | PyAPI_FUNC(int) _PyLong_AsByteArray(PyLongObject* v,
            |                 ^~~~~~~~~~~~~~~~~~~
      _ruamel_yaml.c: In function ‘__Pyx_PyIndex_AsSsize_t’:
      _ruamel_yaml.c:29276:45: error: ‘PyLongObject’ {aka ‘struct _longobject’} has no member named ‘ob_digit’
      29276 |     const digit* digits = ((PyLongObject*)b)->ob_digit;
            |                                             ^~
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for ruamel-yaml-clib
Failed to build ruamel-yaml-clib
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (ruamel-yaml-clib)
```

rgaiacs:
- [ ] The problem here need to be reported to upstream developers https://sourceforge.net/p/ruamel-yaml-clib.
- [ ] The GitHub Actions are using Python 3.10 defined in https://github.com/jupyterhub/mybinder.org-deploy/blob/3428e5497761f9b4d89456af95d6b8160eddb369/.python-version This should be included in the documentation.

PR #3269, opened 2025-04-04
Add virtual environemnt for development managed by Pixi

rgaiacs:
This adds a [Pixi](https://pixi.sh/) project configuration file to enable a virtual environment managed by Pixi.

Developers should be able to run

```bash
pixi run helm-update
pixi run deploy ...
```

and not have to worry about dependencies like https://github.com/jupyterhub/mybinder.org-deploy/issues/3168.

The `requirements.txt` was imported to the Pixi configuration using

```bash
pixi add --pypi $(cat requirements.txt | grep -v '#')
```

This can be used again to avoid disruption with the current workflow using `pip-compile`.

Pixi's task runner is built using [deno_task_shell](https://deno.land/manual@v1.35.0/tools/task_runner#built-in-commands). This imposes some limitations:

- the `for`-loop is not defined,
- the `source` command is not defined,
- `xargs` was re-implemented and does not supports many options.

A workaround the limitations was used.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -6,7 +6,15 @@ platforms = ["linux-64"]
 version = "0.1.0"
 
 [tasks]
-deploy = { cmd = ["source", "cert-manager.env", "&&", "python", "deploy.py"] }
+helm-mybinder = { cmd = ["helm", "dependency", "update", "mybinder"] }
+
+helm-mybinder-kube-system = { cmd = ["helm", "dependency", "update", "mybinder-kube-system"] }
+
+helm-mybinder-tigera-operator = { cmd = ["helm", "dependency", "update", "mybinder-tigera-operator"] }
+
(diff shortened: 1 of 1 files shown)


yuvipanda (changes requested):
I try really hard to stay away from VC funded software dependencies in community projects I help with. All the harm the VC world in general has done comes to me every time I run into new VC funded software, and I can't help but have an emotional reaction :( It's demotivating and I notice I do end up avoiding such interactions if possible, which then leads to lower desire to engage. I understand that may not be rational, but unfortunately it's currently true for me.

As a favor to me, can I ask we keep pixi out of this repo? You could keep this in a separate repo perhaps?

Thank you for considering it

rgaiacs:
> As a favor to me, can I ask we keep pixi out of this repo?

No problem.

PR #3280, opened 2025-04-23
Update URL of GESIS in banner

rgaiacs:
This is a small change in the HTML of the banner.

@arnim and @rgaiacs are sunsetting the GESIS Notebooks brand.


arnim:
Thank you @rgaiacs  😊

PR #3323, opened 2025-06-16
add dependabot config for each image requirements

minrk:
groups: one PR per image


minrk:
we're getting sporadic security PRs, but not general update PRs for the image requirements.

PR #3361, opened 2025-07-11
fix `gcloud container clusters ...` command

manics:
`--project` has to include a suffix

Was the numeric suffix intentionally omitted for security, or was it accidental?


yuvipanda:
Project names aren't considered secure, so I figure it's accidental?

PR #3421, opened 2025-09-05
add plausible

minrk:
closes https://github.com/jupyterhub/team-compass/issues/803

Notably: this doesn't check DNT, unlike Matomo. You can see some discussion about why here: https://github.com/plausible/analytics/discussions/646

We could respect DNT ourselves, like we do for matomo, before loading the plausible scripts.


choldgraf (approved):
This looks good to me - I'll wait for you to merge, and then we can confirm whether this works by checking https://plausible.io/mybinder.org/verification

Candycrush001 (commented):
Please set up the automatic docs review and code complete

Issue #3387, opened 2025-08-25
Launch get stuck in building phase after write to registry


rgaiacs:
This was first reported to me by @shyamgupta196.

When launching https://gesis.mybinder.org/v2/gh/YSKartal/reference_disam/1ba96c28c0c748f95bfbdc33d95f2553c2fc2640, the launching stops after push the container to the registry, i.e.

```
#29 [19/20] COPY /python3-login /usr/local/bin/python3-login
#29 DONE 0.2s

#30 [20/20] COPY /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
#30 DONE 0.2s

#31 exporting to image
#31 exporting layers
#31 exporting layers 1.2s done
#31 writing image sha256:f1e0b6c3e99914b518957492e687a8f88aafeced0b0eefaf3f7b189afac49b0b done
#31 naming to registry.gesis.mybinder.org/i-yskartal-2dreference-5fdisam-e52280:1ba96c28c0c748f95bfbdc33d95f2553c2fc2640 done
#31 DONE 1.2s
```

My understanding is that after the new image is push, the container should stop.

I testes with a fork of https://github.com/binder-examples/conda, i.e. https://github.com/rgaiacs/conda, and the same happens, i.e. the launch get stuck after push the container image to the registry.

```
#28 [16/18] COPY --chown=1000:1000 src/ /home/jovyan/
#28 DONE 0.1s

#29 [17/18] COPY /python3-login /usr/local/bin/python3-login
#29 DONE 0.0s

#30 [18/18] COPY /repo2docker-entrypoint /usr/local/bin/repo2docker-entrypoint
#30 DONE 0.1s

#31 exporting to image
#31 exporting layers
#31 exporting layers 10.9s done
#31 writing image sha256:f17cb1828e42b94520b4f5d8aa817ca3d1949efbd39cef807464d8ceabd72e6a done
#31 naming to registry.gesis.mybinder.org/i-rgaiacs-2dconda-e71d0c:f00a783146e9c6a2ed9726f01fc09fbfbad2f89e done
#31 DONE 10.9s
```

rgaiacs:
[...]

PR #3429, opened 2025-09-10
Downgrade Docker dind from 28.3.3-dind to 27.5.1-dind

rgaiacs:
Following up with [conversation on Zulip](https://jupyter.zulipchat.com/#narrow/channel/469744-jupyterhub/topic/mybinder.2Eorg.20is.20down/near/538019619).

Investigating https://github.com/jupyterhub/mybinder.org-deploy/issues/3387

Reverts https://github.com/jupyterhub/binderhub/pull/1937 after report by [Kaustubh_Patil](https://discourse.jupyter.org/u/kaustubh_patil) and [JuliusEnBW](https://github.com/JuliusEnBW) that the problem appear again.

I will merge this.

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -352,6 +352,10 @@ binderhub:
   imageBuilderType: dind
   dind:
     daemonset:
+      image:
+        # Investigating https://github.com/jupyterhub/mybinder.org-deploy/issues/3387#issuecomment-3239287733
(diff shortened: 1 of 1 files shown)


minrk:
Thanks for isolating it! I wonder if we can figure out a test case to reproduce it without a public deployment. Knowing what it is takes some of the stress off, thank you!

Should we change the default in binderhub as well until we figure this out, so other deployments don't face this, too?

PR #3476, opened 2025-10-20
Move Staging to run on OVH with k3s

yuvipanda:
- Changes existing OVH terraform code to instead be able to support OVH registry only, as we use k3s for kubernetes
- Setup a new kubernetes k3s cluster, following our existing k3s docs, for running the staging cluster
- Deploy to this new staging cluster
- Have the staging cluster's launch events come to the same analytics pipeline as our prod cluster for now, instead of setting up two separate ones as we have so far. This adds a tiny amount of noise to our pipeline (potentially - i see it's a different log name) but is worth it
- Turn off the event publisher related code for now. This means we'll only have this in prod.
- Turn off matomo, as we use plausible now in prod instances.
- Setup k3s on a manually created VM on OVHCloud. I added an ssh key with access to the `ubuntu` user (which has full sudo) to this PR as a secret. 
- DNS is currently set to https://staging.mybinder.2i2c.cloud/ because i no longer have access to mybinder.org DNS. It's on cloudflare but it seems to require 2fa now and i don't have access. Maybe @minrk does? 

## TODO

- [x] Fix the DNS to point to staging.mybinder.org instead of the 2i2c.cloud domains
- [x] Test that this works in our CI
- [ ] Rip out the existing staging cluster and everything with it, say 'thank you' and gently bury it with respect

See https://github.com/jupyterhub/mybinder.org-deploy/issues/3477 (keep it open until we've cleaned up the old staging)


manics:
The mybinder.org domain is managed by Jupyter: https://github.com/jupyterhub/team-compass/issues/424#issuecomment-1772712025

yuvipanda:
i no longer have access to that cloudflare account. It seems to require 2fa and i don’t have 2fa for it :(

manics:
The deployment failed because `events-archiver-secrets` isn't available for staging. It's not configured on our current staging, so I'll revert the `extraVolumes` config added in https://github.com/jupyterhub/mybinder.org-deploy/pull/3476/commits/4d2594f0ce7924d389ca1fa26f0e73ff3b15b024

Issue #3437, opened 2025-09-25
Move away from that GKE cluster


rgaiacs:
On 22 September, we were having some problems with the federation redirect hosted on GKE.

On 25 September, Yuvi [mentioned](https://jupyter.zulipchat.com/#narrow/channel/469744-jupyterhub/topic/mybinder.2Eorg.20is.20completely.20down/near/541348711) on Zulip that

> we should move away from that gke cluster somehow
>
> it's really just running fededration-redirect and the analytics stuff and nothing else

manics:
> and apply the GESIS website skin

Would the work you and @yuvipanda did on decoupling the BinderHub client let us run two BinderHub UIs connecting to the same BinderHub backend, so we could have one with the standard theme and one with the Gesis theme?

manics:
I've pulled out our Google Cloud costs for August 2025.

**Total: $490.28** (includes $99.56 of "savings")

## Split by Service

Service description | Cost ($) | Other savings ($) | Subtotal ($)
-- | -- | -- | --
Compute Engine | 215.85 | 0 | 215.85
Cloud SQL | 131.91 | 0 | 131.91
Kubernetes Engine | 148.79 | -74.39 | 74.4
Cloud Storage | 43.88 | 0 | 43.88
Networking | 48.58 | -25.16 | 23.42
Artifact Registry | 0.57 | 0 | 0.57
Cloud Logging | 0.26 | 0 | 0.26

## Split by cluster name

goog-k8s-cluster-name | Cost ($) | Other savings ($) | Subtotal ($)
-- | -- | -- | --
goog-k8s-cluster-name:prod | 259.39 | 0 | 259.39
  | 227.27 | -25.16 | 202.11
goog-k8s-cluster-name:staging | 103.18 | -74.39 | 28.78

227.27 (middle row) is for resources that aren't labelled with a k8s cluster name. This can be broken down by service into:

Service description | Cost ($) | Other savings ($) | Subtotal ($)
-- | -- | -- | --
All Cloud SQL cost | 131.91 | 0 | 131.91
All Cloud Storage cost | 43.88 | 0 | 43.88
Part of the Networking cost | 43.76 | 25.16 |	18.60

manics:
## Cloud SQL $131.91
This the MySQL database for Matomo (staging and prod). If we're using plausible we could get rid of it.

SKU description | Usage amount | Usage unit | Cost ($) | Other savings ($) | Subtotal ($)
[...]

PR #3513, opened 2025-11-23
Terraform: Delete Matomo on GCP

manics:
Follow-up to https://github.com/jupyterhub/mybinder.org-deploy/pull/3505
https://github.com/jupyterhub/mybinder.org-deploy/issues/3437#issuecomment-3512902667

This may require some manual deletions if the DBs are protected from deletion.


minrk (approved):

manics:
Done!

PR #3537, opened 2025-12-08
Delete GCP staging Terraform infrastructure

manics:
This needs to be implemented by running `terraform destroy` in `terraform/gcp/staging/`


yuvipanda:
YESSS GOOD RIDDANCE! Do you have all the creds needed to run this? If so you should self merge

manics:
Going ... going ... gone!

PR #3551, opened 2025-12-12
add BIDS-funded member on OVH

### LINKED ISSUE #3546: Add a mybinder.org federation member on OVH paid for by BIDS
More information in [#jupyterhub > BIDS joining the mybinder.org federation?](https://jupyter.zulipchat.com/#narrow/channel/469744-jupyterhub/topic/BIDS.20joining.20the.20mybinder.2Eorg.20federation.3F/with/562824080).

We want to add a new federation member sponsored by BIDS (via 2i2c), with cloud costs upto $400 a month.

This issue tracks that.

## Open questions

- [ ] Should we use OVH's managed registry or use object storage managed by our own registry?
- [ ] How do we handle splitting registries, now that we have registries on both OVH and Hetzner? 

Both these have one potential option, which helps us everywhere - move to [harbor](https://github.com/goharbor/harbor-helm) by ourselves backed by object storage. This helps us set up retention policies on hetzner as well, and [replication](https://goharbor.io/docs/1.10/administration/configuring-replication/) so we can replicate images between hetzner and OVH (and any future provider we are on).

Since putting this federation node in rotation would split the registry and slow down launches, we should figure out what choice we want to make here explicitly before that.

minrk:
closes #3546 

Already deployed at https://bids.mybinder.org but still working on it, might come and go.

Characteristics:

- single k3s VM on OVH
- OVH s3-backed registry using Harbor, deployed in the mybinder chart
- at https://bids.mybinder.org
- Tofu deploys all OVH resources
- still just a couple of manual k3s setup steps https://mybinder-sre.readthedocs.io/en/latest/deployment/k3s.html
- Harbor configured with tofu once deployed via helm

Moved ovh staging to ovh-staging (#3550) to make room for 'real' ovh deployment. ovh harbor config is duplicated, but harbor is deployed via helm instead of managed.

Advantage of managed: we don't manage it! And can be deployed with terraform. With limited VM resources, putting harbor on the VM may eat up semi-precious resources
Advantage of self-deployed: we control it, quite a bit cheaper, size not limited to (pretty small!) 6TB

Disadvantage of self-deployed: I don't know how much we lose if we get a fresh VM, even though registry data is stored on s3. I _think_ we might lose it all unless we migrate the postgres database from a backup.

We _could_ have just managed external postgres for durabillity. That would reduce some resource cost on the VM, but the cheapest is ~$55/month.

Still to do:

- [x] maybe redeploy with private networking
- [x] resource planning. With the current budget, I think I have room for a b3-64 node and 5TB of image cache. I haven't put any resource limits on harbor or anything, but need to plan that out. I will bring some revised budget proposals, which might involve creating a fresh VM, which in turn might mean changing IP. Allocating a floating IP would solve that, but that also requires setting up a whole private network, which maybe I should do.
- [x] add to federation, CD

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -217,7 +217,7 @@ binderhub:
             🤍 Donate to mybinder.org!
         </a>
         <div style="text-align:center;">
-        Thanks to <a href="https://www.gesis.org">GESIS</a> and <a href="https://2i2c.org">2i2c</a> for supporting us! 🎉
+        Thanks to <a href="https://www.gesis.org">GESIS</a>, <a href="https://2i2c.org">2i2c</a>, and <a href="https://bids.berkeley.edu">BIDS</a> for supporting us! 🎉
         </div>
         <div style="text-align:center;">
         mybinder.org has updated the base image to Ubuntu 22.04! See the <a href="https://repo2docker.readthedocs.io/en/latest/howto/breaking_changes.html">upgrade guide</a> for details.


config/bids-ovh-kubelet.yaml
```diff
@@ -0,0 +1,8 @@
+# scp me to bids.mybinder.org:/etc/kubelet.yaml
```
manics: This matches the existing docs for all k3s nodes, so could move this to `k3s/kubelet.yaml`?

terraform/ovh/scripts/lookup_flavor_image.py
```diff
@@ -0,0 +1,30 @@
+"""
+Get UUID for OVH flavor and image
+
+OVH provider needs UUID,
+but UI only presents name.
+Need to use the API to produce a mapping.
+"""
```
manics:
```suggestion

# TODO: make the region a parameter
"""
```
Adding this comment will hopefully make it easier to know what to change in future.

config/bids-ovh-kubelet.yaml
minrk: It _almost_ matches - the number of pods is different depending on the available resources of the node. The hetzner nodes are bigger, others could be smaller.

config/bids-ovh.yaml
```diff
@@ -0,0 +1,152 @@
...
+      type: s3
+      s3:
+        # https://github.com/goharbor/harbor-helm/blob/v1.18.1/values.yaml#L225
+        region: us-east-va
+        bucket: mybinder-registry-bids-ovh
+        regionendpoint: https://s3.us-east-va.io.cloud.ovh.us
+        rootdirectory: harbor
+  trivy:
+    enabled: false
+
```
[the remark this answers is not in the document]
minrk: Good call. I moved all the common config to mybinder/values.yaml with `enabled: false`

PR #3588, opened 2026-01-07
Centralise Cryptnono configuration

manics:
All federation members are running the same Kubernetes setup (single node K3s) so we can move all the cryptnono config into a single place (this also fixes the incorrect `containerdHostPath` on staging).

### DIFF
--- a/mybinder/values.yaml
+++ b/mybinder/values.yaml
@@ -1,8 +1,6 @@
 cryptnono:
   enabled: true
   detectors:
-    monero:
-      enabled: false
     # execwhacker and flowkiller need access to containerd, and docker on build hosts,
     # to lookup the container.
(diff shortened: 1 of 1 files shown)


mybinder/values.yaml
```diff
@@ -1,6 +1,8 @@
 cryptnono:
   enabled: true
   detectors:
+    monero:
+      enabled: false
```
manics: Can anyone remember why monero is disabled on all our nodes?

rgaiacs: I don't remember.

minrk: I asked Yuvi and he thought it was related to the ubuntu version maybe having some trouble, and wasn't sure if it should still apply, so we should try removing this and see if whatever it was is fixed, since it seems to be in staging.

PR #3640, opened 2026-01-28
k3s: forward pod logs to syslog

minrk:
This is the simplest solution to logs retention I could find.

Kubernetes pod logs are written to `/var/logs/pods/...`, but are deleted when the pod is deleted, meaning we can't do things like review logs of pods or containers killed by cryptnono.

What this does is add a tiny rsyslog daemon that watches `/var/logs/pods/**` files for logs and then forwards them to the host's existing `/run/systemd/journal/syslog` socket, so they end up in `/var/log/syslog` on the host

Allows

    tail -f /var/log/syslog | grep -o 'podlogs.*'

to show logs from all pods, including _stopped_ pods.

we get to inherit syslog's rotation, retention (and cleanup), which is by default:

- rotate every 7 days
- keep 4 rotations (total 28 days retention)

alternative: deploy a fancier logging agent like loki.

Advantage of Loki (or similar): we get log viewer/query in Grafana
Disadvantage: seems like a lot to deploy just to get a few days of log retention. But if it's easy, maybe that's better

Things to watch out for: disk usage in /var/log. Retaining pod logs will _dramatically_ increase the size of syslog. Not sure by how much.

### DIFF
--- a/mybinder/templates/pod-logs/daemonset.yaml
+++ b/mybinder/templates/pod-logs/daemonset.yaml
@@ -17,7 +17,7 @@ spec:
         release: {{ .Release.Name }}
         heritage: {{ .Release.Service }}
       annotations:
-        checksum/configmap: {{ include (print $.Template.BasePath "/rsyslog/configmap.yaml") . | sha256sum }}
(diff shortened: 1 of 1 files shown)


manics:
/test-this-pr

(A subsequent merge to main will have undone the last test deploy)

PR #3783, opened 2026-04-29
harbor: increase quotas to 15TB

minrk:
baseline requirements increased with full circular mirroring

@rgaiacs is this okay? It is also an option to reduce retention.

I think we are _likely_ to pass 10TB. I don't think by _much_, but if we keep a 10T ceiling, we would need to watch carefully.

We are currently at around 1.4T new builds per week and half that deleted due to retention. I expect it to stabilize around 10-12 with current patterns


rgaiacs (approved):
Looks good to me.

PR #3794, opened 2026-05-07
limit build max age to 1 hour

minrk:
approximately all builds that take more than one hour fail due to being stuck on something, wasting hours of resources


manics:
https://github.com/jupyterhub/binderhub/blob/9695125979fd3e22e91ff1b42ba8799178369829/binderhub/app.py#L682-L699

says we should use `c.{cleaner_name}.max_age`

minrk:
Good catch, thanks

PR #3811, opened 2026-05-19
expand ban sources

minrk:
increases number of banned sources


minrk:
need to watch this deploy, since a bigger ban list can affect configmap limits and memory requirements for DNS

PR #3812, opened 2026-05-19
add cluster-entrypoint service

minrk:
preparation for shifting away from ingress-nginx

first step of #3811 

adapted from https://github.com/2i2c-org/infrastructure/pull/7728 (adds annotations required for stable IP on gcloud)

I don't actually know how this is going to work on k3s, we haven't had multiple LoadBalancer services on k3s nodes yet, and I think you can't have multiple load balancers on one port on one host. I think we'll have to switch over in one go, which is more disruptive than I would want it to be.

### DIFF
--- a/config/bids-ovh.yaml
+++ b/config/bids-ovh.yaml
@@ -1,5 +1,8 @@
 projectName: bids-ovh
 
+clusterEntryPoint:
+  enabled: false
+
 harbor:
   enabled: true
   expose:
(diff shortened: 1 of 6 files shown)


minrk:
k3s can only have one LoadBalancer service, so we have to shift nginx.service to ClusterIP at the same time as enabling clusterEntrypoint. So this PR only enables it on:

1. staging (k3s) and
2. prod (separate LoadBalancer, new public IP)

after that works, I can

1. enable the clusterEntrypoint service on members one at a time
2. update mybinder.org DNS to point to the public IP of clusterEntrypoint on prod for the zero downtime upgrade

Issue #3639, opened 2026-01-27
Upcoming deprecation of ingress-nginx


minrk:
ingress-nginx is going to [end support in March](https://www.kubernetes.dev/blog/2025/11/12/ingress-nginx-retirement/).

2i2c discussion on the same topic: https://github.com/2i2c-org/infrastructure/issues/7106

I see 3 main options:

- switch to nginx ingress maintained by nginx itself: https://docs.nginx.com/nginx-ingress-controller/install/helm/open-source/
- traefik ingress controller (builtin to k3s by default, so we could just not disable it) https://doc.traefik.io/traefik/migrate/nginx-to-traefik/
- Full migration to Gateway API via envoy-gateway https://gateway.envoyproxy.io

Summary from my experiments:

- Gateway migration seems to be "the future" but it doesn't really seem like the future is now. In a lot of ways, the Gateway API is worse for our use case than Ingress, because we don't have any of the problems it tries to solve. Also, BinderHub doesn't support Gateway API yet (it should and isn't hard), but everything else we deploy does.
- F5 nginx-ingress chart seems least disruptive, since it's the same thing, with config expressed a little differently
- traefik would probably be simplest, since it's builtin to k3s, but hardest to customize if we do it that way (since it's builtin to k3s, not part of our CD pipeline), but it's vulnerable to being run by a single VC-backed company, which have a habit of pulling the rug out from under users.

I have deployed a JupyterHub migration to traefik and another to envoy-gateway. Both went fine, traefik was easy while envoy-gateway took some figuring out and a lot more manual and redundant configuration. But what I actually think we should do _here_ is switch to F5 nginx, which also happens to be the only one I _haven't_ done.

minrk:
Based on 2i2c's experience, I think we should:

1. follow their example deploying a cluster-entrypoint Service so we don't have DNS/IP disruptions (https://github.com/2i2c-org/infrastructure/pull/7728)
2. make sure DNS points to the new Service
3. deploy traefik (not nginx-ingress, which has caused problems like https://github.com/2i2c-org/infrastructure/issues/8063), which seems to have the smoothest, simplest [migration](https://doc.traefik.io/traefik/migrate/nginx-to-traefik/)
4. point cluster-entrypoint to traefik
5. retire nginx-ingress
6. keep looking out for when to transition to envoy-gateway, which I think shouldn't be too big, but I don't want to do it all at once

consideRatio:
[...]

PR #3819, opened 2026-05-20
add traefik ingress controller

minrk:
enable kubernetesIngressNGINX to take over nginx ingresses without modification

not enabled by default yet, only on 

steps:

1. deploy traefik (gets no traffic)
2. point clusterEntrypoint at traefik
3. shutdown ingress-nginx
4. migrate ingresses to `kubernetesIngress` away from `kubernetesIngressNGINX` (no pressure, I think, plus there's very little)

rough plan: test traefik on staging; once happy, enable it on one federation member and see how it goes for a while, then switch the rest over

part of #3639

### DIFF
--- a/.github/workflows/watch-dependencies.yaml
+++ b/.github/workflows/watch-dependencies.yaml
@@ -152,6 +152,13 @@ jobs:
             chart_dep_devel_flag: ""
             github_repo: ""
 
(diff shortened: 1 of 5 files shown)


minrk:
not sure why test-this-pr is getting a 404 on this one, but merging should only affect staging

PR #3820, opened 2026-05-20
fix watch-dependencies with oci changes

minrk:
`helm show chart --repo=oci://ghcr.io/traefik/helm traefik` doesn't work, we have to use `helm show chart oci://ghcr.io/traefik/helm/traefik`, but `helm show chart https://other.example/charts/name` doesn't work, so we have to know if we are checking an oci registry or not.

This should allow watch-dependencies PRs to be tested (not including the PR step)


minrk:
Can't test PRs due to our use of environment secrets

If we separate the jobs so that the create a PR is its own job with the inputs of the other stages, we can have testing. Until then, we can only test in main.

PR #3823, opened 2026-05-20
test cert-manager with traefik ingress

minrk:
staging is now serving through traefik and it's working fine. This is a test of one of the main hiccups 2i2c faced, which is F5 nginx-ingress did not work properly with cert-manager where ingress-nginx was working

part of #3639


minrk:
cert-manager successfully issued certificates via traefik ingress, no problem

PR #3840, opened 2026-05-21
opt prometheus, static ingress into traefik ingressClass

minrk:
something low-impact to test, moving from kubernetesIngressNGINX to kubernetesIngress

part of #3639


minrk:
while this changed the ingressClassName to traefik, it did _not_ cause the kubernetesIngress provider to be used, it's still using `kubernetesIngressNGINX`

PR #3843, opened 2026-05-21
everything on regular traefik ingress except the key ones: hub, binderhub, harbor

minrk:
adds to traefik provider:

- federation-redirect (mybinder.org)
- gcs-proxy (archive.analytics)
- redirector (docs.mybinder.org)
- matomo (not used, should probably be removed)

part of #3639

### DIFF
--- a/mybinder/templates/federation-redirect/ingress.yaml
+++ b/mybinder/templates/federation-redirect/ingress.yaml
@@ -8,7 +8,7 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
   annotations:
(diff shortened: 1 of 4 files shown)


minrk:
will move the remainder over tomorrow morning, to check on things like large harbor requests, notebook saves, websocket sessions, etc.

minrk:
to be clear: all of this traffic is already served by traefik, it is just changing over from the `kubernetesingressnginx` interim provider to traefik's default`kubernetesingress` provider. Which changes how some annotations are interpreted and perhaps some defaults as well.

PR #3850, opened 2026-05-22
finish moving to traefik kubernetesIngress provider from kuberntesIngresNGINX

minrk:
this finishes the move from traefik supporting nginx config annotations to traefik-native config

will test on staging first, specifically:

- request timeout, that uploads can take more than a minute
- cert-manager certificates by adding a domain to static.staging

part of #3639


minrk:
cert-manager works

Confirmed the 600s read timeout is working as well, and that a 2G layer push that takes 90s succeeds.

PR #3851, opened 2026-05-24
Add more diagnostic info to terminate_pod log

manics:
The minesweeper logs contain a lot of context for each pod termination. Include some of that in the `terminate_pod` log which is forwarded to Zulip.

Is there a nice way to trigger minesweeper for testing?


minrk:
Not sure about testing, but minesweeper has features similar to cryptnono's execwhacker that we don't really use anymore since execwhacker is so much more effective. You could add a test string there and try it.

minrk (approved):

minrk:
This is deployed on staging and minesweeper logged a nice json payload, but it didn't get to zulip

minrk:
need to update the regex in `k8s-log-alerter-zulip.rules.Minesweeper killed` to match the new JSON message.

minrk:
(or the message to match the existing pattern)

manics:
Fixed!

PR #3943, opened 2026-09-18
internal culler: cull idle terminals as well as kernels

minrk:
otherwise, opening a terminal effectively keeps the server alive


minrk:
actually, opening a terminal _doesn't_ keep a server alive, but apparently it's meant to: https://github.com/jupyter-server/jupyter_server_terminals/issues/112

manics (approved):
Might as well future proof mybinder by adding it now

